PR #13, opened 2022-07-01
Add DCO to CONTRIBUTING.md

susodapop:
Introduces Databricks standard DCO to the CONTRIBUTING guide.

Also adds a GitHub Action step which checks each commit in the pull request for a DCO sign-off and leaves a comment if this check fails. 

One challenging aspect of this approach is that our current DCO check step doesn't have any `outputs` hooks that let us access its list of which commits failed the DCO check. I can write a custom action step that forks the existing behaviour and sets those outputs as a follow-up if it becomes necessary.

For now, any member of this org can see exactly which commits failed the DCO check by examining the PR Check results. These aren't visible to outside contributors, though, which is not as convenient.

#### Example comment:
CleanShot 2022-07-29 at 16 03 38@2x


yunbodeng-db (approved):
Nice!

yunbodeng-db (approved):

yunbodeng-db:
We need to do this to all other connectors as well?

susodapop:
Yes I've created tracking tickets internally to implement this for NodeJS and Go.

PR #15, opened 2022-07-12
OAuth implementation

moderakh:
This PR:
* adds the foundation for OAuth against Databricks account on AWS with BYOIDP.
* it copies one internal module that Steve Weis @sweisdb wrote for Databricks CLI (oauth.py). Once ecosystem-dev team (Serge, Pieter) build a python sdk core we will move this code to their repo as a dependency. 
* the PR provides authenticators with visitor pattern format for stamping auth-token which later is intended to be moved to the repo owned by Serge @nfx and and Pieter @pietern 

```
connection = sql.connect(
   server_hostname='***.cloud.databricks.com',
   http_path='sql/protocolv1/o/***',
   auth_type="databricks-oauth"
)

for x in range(1, 100):
  cursor = connection.cursor()
  cursor.execute('SELECT * FROM RANGE(10)')
  result = cursor.fetchall()
  for row in result:
    print(row)
  cursor.close()

connection.close()
```


src/databricks/sql/auth/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Databricks CLI
```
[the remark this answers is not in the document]
yunbodeng-db: Ditto to others.

src/databricks/sql/auth/auth.py
```diff
@@ -0,0 +1,103 @@
...
+                         token=access_token,
+                         auth_type=auth_type,
+                         use_cert_as_auth=use_cert_as_auth,
+                         tls_client_cert_file=tls_client_cert_file)
+
+
+def get_authenticator(cfg: ClientContext):
+    if cfg.auth_type == AuthType.DATABRICKS_OAUTH.value:
+        return OAuthAuthenticator(cfg.hostname, cfg.oauth_client_id, cfg.oauth_scopes)
+    elif cfg.token is not None:
```
yunbodeng-db, line 81: My concern with adding other authenticators in here too is it adds test burdens (which I cannot see tests being added in the same PR). I would leave the untested (out of scope) code out of the PRs too so we are sure that submitted code has been verified and has good test coverage.

pyproject.toml
```diff
@@ -13,6 +13,9 @@ python = "^3.7.1"
 thrift = "^0.13.0"
 pyarrow = "^5.0.0"
 pandas = "^1.3.0"
+click=">=7.0"
```
pietern, line 16: I recommend against depending on click in a non-CLI library.

src/databricks/sql/auth/__init__.py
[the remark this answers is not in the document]
pietern: These are copied from https://github.com/databricks/databricks-cli. This is not an issue.

src/databricks/sql/auth/auth.py
```diff
@@ -0,0 +1,103 @@
...
+class SqlConnectorClientContext(ClientContext):
+    def __init__(self,
+                 hostname: str,
+                 username: str = None,
+                 password: str = None,
+                 access_token: str = None,
+                 auth_type: str = None,
+                 use_cert_as_auth: str = None,
+                 tls_client_cert_file: str = None):
+        super().__init__(oauth_scopes="sql offline_access",
```
pietern, line 66: This should be a constant.

src/databricks/sql/auth/auth.py
```diff
@@ -0,0 +1,103 @@
...
+                         # to be changed once registered on the service side
+                         oauth_client_id="databricks-cli",
```
pietern, line 68: This should be a constant as well.

src/databricks/sql/auth/auth.py
pietern: There are no tests at all at the moment. @moderakh Are you planning to include a couple in this PR?

If so, I recommend keeping these simpler authenticators around.

src/databricks/sql/auth/auth.py
```diff
@@ -0,0 +1,103 @@
...
+        self.password = password
+        self.token = token
+        self.auth_type = auth_type
+        self.oauth_scopes = oauth_scopes
+        self.oauth_client_id = oauth_client_id
+        self.use_cert_as_auth = use_cert_as_auth
+        self.tls_client_cert_file = tls_client_cert_file
+
+
+class SqlConnectorClientContext(ClientContext):
```
pietern, line 57: What is the value of the subclass here? It doesn't seem to do anything.

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,78 @@
...
+        request_headers['Authorization'] = self.__authorization_header_value
+
+
+class OAuthAuthenticator(Authenticator):
+    # TODO: moderakh the refresh_token is only kept in memory. not saved on disk
+    # hence if application restarts the user may need to re-authenticate
+    # I will add support for this outside of the scope of current PR.
+    def __init__(self, hostname, client_id, scopes):
+        self._hostname = self._normalize_host_name(hostname=hostname)
+        self._scope = scopes
```
pietern, line 59: Mix of singular and plural. If this is an array of scopes, it should be plural.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+
+import base64
+import hashlib
+import json
+import os
+import webbrowser
+
+from datetime import datetime, timedelta, tzinfo
+
+import click
```
pietern, line 33: The only use of click here is `click.echo`. Can be replaced with `print`'s to get rid of the dependency.

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -0,0 +1,48 @@
...
+        super().__init__(uri_or_host, port, path, cafile, cert_file, key_file, ssl_context)
+        self.__authenticator = authenticator
+
+    def setCustomHeaders(self, headers):
+        self._headers = headers
+        super().setCustomHeaders(headers)
+
+    def flush(self):
+        # TODO retry behaviour
+        # TODO locking for multiple concurrent refresh token
```
pietern, line 43: Would this client be used in a multi threaded context at all? I would think with methods like `read`, `write`, and `flush`, there is an implied expectation of the caller being single threaded, and multiple threads using multiple clients.

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -0,0 +1,48 @@
...
+        req = {}
+        self.__authenticator.add_auth_token(req)
+        self._headers['Authorization'] = req['Authorization']
```
pietern, line 46: Implied dependency of the function setting the `Authorization` header.

This could be more robust by naming the authenticator function `add_headers` and merging the entire return value into the `self._headers` dict. That would also allow for adding multiple headers out of the box.

pietern (commented):
I suspect you're targeting Python 3 only (as Python 2 is EOL), and if so, this code would benefit from having a couple more type annotations. I thought about this when seeing the `scope` vs `scopes` name. With an annotation, singular or plural would be obvious.

What are your plans for unit tests?

src/databricks/sql/auth/oauth.py
nfx: print is bad, `logging` must be used - https://docs.python.org/3/howto/logging.html

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+def get_client(client_id):
+    return oauthlib.oauth2.WebApplicationClient(client_id)
+
+
+def get_redirect_url(port=REDIRECT_PORT):
+    return "http://localhost:{port}".format(port=port)
+
+
+def fetch_well_known_config(idp_url):
+    known_config_url = "{idp_url}/.well-known/oauth-authorization-server".format(idp_url=idp_url)
```
nfx, line 89: please use f-strings https://realpython.com/python-f-strings/

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+                       "Verify it is a valid workspace URL and that OAuth is "
+                       "enabled on this account.".format(idp_url=idp_url))
+
+    if response.status_code != 200:
+        error_and_quit("Received status {status} OAuth configuration from "
+                       "{idp_url}.\n Verify it is a valid workspace URL and "
+                       "that OAuth is enabled on this account."
+                       .format(status=response.status_code, idp_url=idp_url))
+    try:
+        return json.loads(response.text)
```
nfx, line 103: use `response.json()`, it's a standard method, re-implementing it here only makes it more complicated

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+OIDC_REDIRECTOR_PATH = "oidc"
+REDIRECT_PORT = 8020
+UTC = UTCTimeZone()
+
+
+def get_client(client_id):
+    return oauthlib.oauth2.WebApplicationClient(client_id)
+
+
+def get_redirect_url(port=REDIRECT_PORT):
```
nfx, line 84: what about random port? what if other process is listening on 8020 on client machine?

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+import jwt
+from jwt import PyJWTError
+
+import oauthlib.oauth2
+from oauthlib.oauth2.rfc6749.errors import OAuth2Error
+
+import requests
+from requests.exceptions import RequestException
+
+from databricks_cli.utils import error_and_quit
```
nfx, line 44: you're not allowed to use it outside of cli package. it calls for `sys.exit`, which is end of the process. use proper exception re-wrapping.

https://github.com/databricks/databricks-cli/blob/main/databricks_cli/utils.py#L94-L100

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+def get_redirect_url(port=REDIRECT_PORT):
+    return "http://localhost:{port}".format(port=port)
+
+
+def fetch_well_known_config(idp_url):
+    known_config_url = "{idp_url}/.well-known/oauth-authorization-server".format(idp_url=idp_url)
+    try:
+        response = requests.request(method="GET", url=known_config_url)
+    except RequestException:
+        error_and_quit("Unable to fetch OAuth configuration from {idp_url}.\n"
```
nfx, line 93: replace with a dedicated exception, don't use this function, which is equal to panic.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+                       "Verify it is a valid workspace URL and that OAuth is "
+                       "enabled on this account.".format(idp_url=idp_url))
+
+    if response.status_code != 200:
+        error_and_quit("Received status {status} OAuth configuration from "
```
nfx, line 98: replace with a dedicated exception, don't use this function, which is equal to panic.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+def get_idp_url(host):
+    maybe_scheme = "https://" if not host.startswith("https://") else ""
+    maybe_trailing_slash = "/" if not host.endswith("/") else ""
+    return "{scheme}{host}{trailing}{path}".format(
+        scheme=maybe_scheme, host=host, trailing=maybe_trailing_slash, path=OIDC_REDIRECTOR_PATH)
+
+
+def get_challenge(verifier_string=token_urlsafe(32)):
+    digest = hashlib.sha256(verifier_string.encode('UTF-8')).digest()
+    challenge_string = base64.urlsafe_b64encode(digest).decode("UTF-8").replace('=', '')
```
nfx, line 119: nit: inconsistent use of single and double quotes in the same method.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+        return timedelta(0)
+
+
+# Some constant values
+OIDC_REDIRECTOR_PATH = "oidc"
+REDIRECT_PORT = 8020
+UTC = UTCTimeZone()
+
+
+def get_client(client_id):
```
nfx, line 80: this function is used only once and i've mistakenly assumed it as "databricks client". please remove the function and instantiate webapp client directly, for understandability and readability

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+    response = requests.request(method="POST", url=token_request_url, data=data, headers=headers)
+    oauth_response = json.loads(response.text)
+    return oauth_response
+
+
+def send_refresh_token_request(hostname, client_id, refresh_token):
+    idp_url = get_idp_url(hostname)
+    oauth_config = fetch_well_known_config(idp_url)
+    token_request_url = oauth_config['token_endpoint']
+    client = get_client(client_id=client_id)
```
nfx, line 218:
```suggestion
    client = oauthlib.oauth2.WebApplicationClient(client_id)
```

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+def check_and_refresh_access_token(hostname, access_token, refresh_token):
+    now = datetime.now(tz=UTC)
+    # If we can't decode an expiration time, this will be expired by default.
+    expiration_time = now
+    try:
+        # This token has already been verified and we are just parsing it.
+        # If it has been tampered with, it will be rejected on the server side.
+        # This avoids having to fetch the public key from the issuer and perform
+        # an unnecessary signature verification.
+        decoded = jwt.decode(access_token, options={"verify_signature": False})
```
nfx, line 239:
```suggestion
        decoded = json.loads(base64.standard_b64decode(access_token.split(".")[1]))
```

please rewrite it with something like that - don't depend on `PyJWT`, we don't need it here. jwt tokens are just dot-separated base-64-encoded jsons - https://jwt.io/

pyproject.toml
```diff
@@ -13,6 +13,9 @@ python = "^3.7.1"
...
+pyjwt=">=1.7.0"
```
nfx, line 17: please remove this dependency.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+    fresh_access_token, fresh_refresh_token = get_tokens_from_response(oauth_response)
+    return fresh_access_token, fresh_refresh_token, True
+
+
+def get_tokens(hostname, client_id, scope=None):
+    idp_url = get_idp_url(hostname)
+    oauth_config = fetch_well_known_config(idp_url)
+    # We are going to override oauth_config["authorization_endpoint"] use the
+    # /oidc redirector on the hostname, which may inject additional parameters.
+    auth_url = "{}oidc/v1/authorize".format(hostname)
```
nfx, line 265: use f-strings

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+        auth_response = get_authorization_code(
+            client,
+            auth_url,
+            redirect_url,
+            scope,
+            state,
+            challenge,
+            REDIRECT_PORT)
+    except OAuth2Error as err:
+        error_and_quit("OAuth Authorization Error: {error}".format(error=err.description))
```
nfx, line 280: rewrap as proper exceptions

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+    try:
+        auth_response = get_authorization_code(
+            client,
+            auth_url,
+            redirect_url,
+            scope,
+            state,
+            challenge,
+            REDIRECT_PORT)
+    except OAuth2Error as err:
```
nfx, line 279:
```suggestion
    except OAuth2Error as e:
```

is more idiomatic way of catching exceptions in python

src/databricks/sql/auth/thrift_http_client.py
nfx, line 46: please use visitor pattern or anything else, that will set more than one header. most of the authorizers do set more than one headers, sometimes three.

src/databricks/sql/client.py
```diff
@@ -90,25 +85,7 @@ def __init__(
...
-            ]
-        elif access_token:
-            authorization_header = [("Authorization", "Bearer {}".format(access_token))]
-        elif not (
-            kwargs.get("_use_cert_as_auth") and kwargs.get("_tls_client_cert_file")
-        ):
-            raise ValueError(
-                "No valid authentication settings. Please provide an access token."
-            )
+        authenticator = get_python_sql_connector_authenticator(server_hostname, **kwargs)
```
nfx, line 88: call it `auth`. or visitor. not authenticator, because it's not an established name in the unified client auth.

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,78 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from databricks.sql.auth.oauth import get_tokens, check_and_refresh_access_token
+import base64
+
+
+class Authenticator:
```
nfx, line 29:
```suggestion
# Private API: this is an evolving interface and it will change in the future.
# Please must not depend on it in your applications.
class CredentialsProvider:
```

(or `Authorizer` - authenticator is something different)

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,78 @@
...
+    def add_auth_token(self, request_headers):
+        pass
+
+
+class AccessTokenAuthenticator(Authenticator):
```
nfx, line 34:
```suggestion
# Private API: this is an evolving interface and it will change in the future.
# Please must not depend on it in your applications.
class AccessTokenAuthenticator(Authenticator):
```

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, access_token):
+        self.__authorization_header_value = "Bearer {}".format(access_token)
+
+    def add_auth_token(self, request_headers):
+        request_headers['Authorization'] = self.__authorization_header_value
+
+
+class UserPassAuthenticator(Authenticator):
```
nfx, line 42:
```suggestion
# Private API: this is an evolving interface and it will change in the future.
# Please must not depend on it in your applications.
class BasicAuthenticator(Authenticator):
```

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,78 @@
...
+        auth_credentials = "{username}:{password}".format(username, password).encode("UTF-8")
+        auth_credentials_base64 = base64.standard_b64encode(auth_credentials).decode("UTF-8")
+
+        self.__authorization_header_value = "Basic {}".format(auth_credentials_base64)
+
+    def add_auth_token(self, request_headers):
+        request_headers['Authorization'] = self.__authorization_header_value
+
+
+class OAuthAuthenticator(Authenticator):
```
nfx, line 53:
```suggestion
# Private API: this is an evolving interface and it will change in the future.
# Please must not depend on it in your applications.
class DatabricksOAuthAuthenticator(Authenticator):
```

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,78 @@
...
+    # TODO: moderakh the refresh_token is only kept in memory. not saved on disk
+    # hence if application restarts the user may need to re-authenticate
+    # I will add support for this outside of the scope of current PR.
+    def __init__(self, hostname, client_id, scopes):
```
nfx, line 57: hostname should not be a property of credentials provider and it's to be taken from configuration.

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,78 @@
...
+        self._refresh_token = refresh_token
+
+    def add_auth_token(self, request_headers):
+        check_and_refresh_access_token(hostname=self._hostname,
+                                       access_token=self._access_token,
+                                       refresh_token=self._refresh_token)
+        request_headers['Authorization'] = "Bearer {}".format(self._access_token)
+
+    @staticmethod
+    def _normalize_host_name(hostname):
```
nfx, line 71: this belongs to the layer that instantiates this class, not this concrete class

nfx (changes requested):
code smell: all functions of `src/databricks/sql/auth/oauth.py` are shuffled to a top-level namespace and it's difficult to understand which methods of  belong to Authorizer and which are related to web browser flow. please refactor this file to be a bit more object-oriented.

please remove `click` and `pyjwt` dependencies - you don't need them. perhaps even `databricks_cli` as dependency.

src/databricks/sql/auth/__init__.py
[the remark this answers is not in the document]
moderakh: thanks for pointing that out. fixed.

src/databricks/sql/auth/auth.py
moderakh: addressed

src/databricks/sql/auth/auth.py
moderakh: thanks, addressed.

src/databricks/sql/auth/authenticators.py
moderakh: thanks, addressed.

src/databricks/sql/auth/auth.py
moderakh: Added tests covering the other auth providers.
https://github.com/databricks/databricks-sql-python/blob/03ab454bdd4d9a39d306873d916fb3496d4496df/tests/test_auth.py

src/databricks/sql/auth/auth.py
moderakh: addressed.

pyproject.toml
moderakh: thanks addressed, replaced with logger

src/databricks/sql/auth/oauth.py
moderakh: the intention was to agree on the rest of code structure before I make changes to oauth class from cli.

already addressed.

src/databricks/sql/auth/oauth.py
moderakh: the intention was to agree on the rest of code structure before I make changes to oauth class from cli.

already addressed.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+def get_authorization_code(client, auth_url, redirect_url, scope, state, challenge, port):
+    #pylint: disable=unused-variable
+    (auth_req_uri, headers, body) = client.prepare_authorization_request(
+        authorization_url=auth_url,
+        redirect_url=redirect_url,
+        scope=scope,
+        state=state,
+        code_challenge=challenge,
+        code_challenge_method="S256")
+    click.echo("Opening {uri}".format(uri=auth_req_uri))
```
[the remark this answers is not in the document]
moderakh: addressed.

src/databricks/sql/auth/oauth.py
moderakh: thanks already addressed.

src/databricks/sql/auth/oauth.py
moderakh: thanks addressed.

src/databricks/sql/auth/oauth.py
moderakh: addressed.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+        authorization_code_response = \
+            client.parse_request_uri_response(full_redirect_url, state=state)
+    except OAuth2Error as err:
+        error_and_quit("OAuth Token Request error {error}".format(error=err.description))
+    return authorization_code_response
+
+
+def send_auth_code_token_request(client, token_request_url, redirect_url, code, verifier):
+    token_request_body = client.prepare_request_body(code=code, redirect_uri=redirect_url)
+    data = "{body}&code_verifier={verifier}".format(body=token_request_body, verifier=verifier)
```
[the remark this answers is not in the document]
moderakh: addressed.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+        click.echo("Listening for OAuth authorization callback at {uri}"
+                   .format(uri=redirect_url))
+        httpd.handle_request()
+
+    if not global_request_path:
+        error_and_quit("No path parameters were returned to the callback at {uri}"
+                       .format(uri=redirect_url))
+    # This is a kludge because the parsing library expects https callbacks
+    # We should probably set it up using https
+    full_redirect_url = "https://localhost:{port}/{path}".format(
```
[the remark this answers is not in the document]
moderakh: addressed.

src/databricks/sql/auth/oauth.py
moderakh: addressed.

src/databricks/sql/auth/oauth.py
moderakh: addressed.

src/databricks/sql/auth/oauth.py
moderakh: Thanks @nfx this was very helpful.

src/databricks/sql/client.py
moderakh: addressed, renamed to auth_provider

src/databricks/sql/auth/thrift_http_client.py
moderakh: thanks for pointing this out. addressed.

moderakh: thanks for the comment. addressed.

src/databricks/sql/auth/thrift_http_client.py
moderakh: thanks for the comment. correct. updated the comment accordingly.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,286 @@
...
+        set_request_path(self.path)
+
+    def log_message(self, format, *args):
+        #pylint: disable=redefined-builtin
+        #pylint: disable=unused-argument
+        return
+
+
+def get_authorization_code(client, auth_url, redirect_url, scope, state, challenge, port):
+    #pylint: disable=unused-variable
```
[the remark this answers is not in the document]
moderakh: addressed.

src/databricks/sql/auth/authenticators.py
moderakh: @nfx I am not sure if I understood what you meant in this comment.

`DatabricksOAuthProvider` hides the oauth implementation, so it needs to be aware of hostname or redirect url to initiate oauth protocol. 

in the python sql connector, the end point information is passed as following
```python 
connection = sql.connect(
      server_hostname='moderakh-test.cloud.databricks.com',
      http_path='sql/protocolv1/o/***',
      auth_type="databricks-oauth"
)
```

the implementation passed `server_hostname` down to `DatabricksOAuthProvider` to be used by oauth implementation, there is no independent configuration support.

moderakh:
> @pietern  I suspect you're targeting Python 3 only (as Python 2 is EOL), and if so, this code would benefit from having a couple more type annotations. I thought about this when seeing the scope vs scopes name. With an annotation, singular or plural would be obvious.
> What are your plans for unit tests?

thanks for the comment.
* We only support Python 3.
* I already added type annotation for scopes (thanks for pointing that out, will try to add for others as well). 
* I already added some unit test for other auth providers (access-token, user/pass here: https://github.com/databricks/databricks-sql-python/blob/410ef1642339aba1f681501b9911f57a4fd71ed6/tests/test_auth.py
* adding test for oauth will be a bit more complex, probably will add some mocking test, but that will be the last piece as I need everything else to be correctly structured to not do re-do refactor tests (I may do that one after this PR merges)

moderakh:
> @nfx code smell: all functions of src/databricks/sql/auth/oauth.py are shuffled to a top-level namespace and it's difficult to understand which methods of belong to Authorizer and which are related to web browser flow. please refactor this file to be a bit more object-oriented.
> 
> please remove click and pyjwt dependencies - you don't need them. perhaps even databricks_cli as dependency.

@nfx Thanks for the comment. I removed pyjwt and click dependencies. I will look at refactoring the code in oauth.py that we borrowed from databricks-cli. except this one and one or two other comments, all of your other comments are addressed.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,282 @@
...
+
+try:
+    from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+except ImportError:
+    from http.server import BaseHTTPRequestHandler, HTTPServer
+
+logger = logging.getLogger(__name__)
+
+
+# This could use 'import secrets' in Python 3
```
sweisdb, line 48: Do we need to support Python 2? If not, we can drop some of these workarounds for Python 2.

src/databricks/sql/auth/oauth.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+def set_request_path(path):
+    global global_request_path
+    global_request_path = path
+
+
+class SingleRequestHandler(BaseHTTPRequestHandler):
+    RESPONSE_BODY = """<html>
+<head>
```
sweisdb, line 130: Maybe put this into an external file where it will be easier to view and edit? (Yes, I know this is what we did for databricks-cli, but that was kind of treated as experimental.)

sweisdb (approved):
LGTM overall.

src/databricks/sql/auth/oauth.py
moderakh: Thanks for the comment. only python 3. I changed the code to use python 3 APIs. addressed here and in one other place.

src/databricks/sql/auth/oauth.py
moderakh: thanks for the comment addressed. also I got rid of the global variable.

src/databricks/sql/__init__.py
```diff
@@ -44,7 +44,6 @@ def TimestampFromTicks(ticks):
     return Timestamp(*time.localtime(ticks)[:6])
 
 
-def connect(server_hostname, http_path, access_token, **kwargs):
+def connect(server_hostname, http_path, experimental_oauth_persistence=None, **kwargs):
     from .client import Connection
-
-    return Connection(server_hostname, http_path, access_token, **kwargs)
+    return Connection(server_hostname, http_path, experimental_oauth_persistence, **kwargs)
```
pietern, line 49: This file can be reverted; no need to call out `experimental_oauth_persistence` here as it is carried forward by `**kwargs` just like all the other optional keyword arguments.

src/databricks/sql/auth/auth.py
```diff
@@ -0,0 +1,96 @@
...
+        return BasicAuthProvider(cfg.username, cfg.password)
+    elif cfg.use_cert_as_auth and cfg.tls_client_cert_file:
+        # no op authenticator. authentication is performed using ssl certificate outside of headers
+        return CredentialsProvider()
+    else:
+        raise RuntimeError("No valid authentication settings!")
+
+
+OAUTH_SCOPES = ["sql", "offline_access"]
+# TODO: moderakh to be changed once registered on the service side
```
pietern, line 78: Does this need to be addressed prior to merging?

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, auth_provider, uri_or_host, port=None, path=None, cafile=None, cert_file=None, key_file=None, ssl_context=None):
+        super().__init__(uri_or_host, port, path, cafile, cert_file, key_file, ssl_context)
+        self.__auth_provider = auth_provider
+
+    def setCustomHeaders(self, headers):
+        self._headers = headers
+        super().setCustomHeaders(headers)
+
+    def flush(self):
+        # TODO retry behaviour
```
pietern, line 41: If this is a real TODO, please refer to issue # here. Otherwise it should be removed.

src/databricks/sql/client.py
```diff
@@ -26,7 +22,7 @@ def __init__(
         self,
         server_hostname: str,
         http_path: str,
-        access_token: str,
+        oauth_persistence: OAuthPersistence = None,
```
pietern, line 25: Can you add back `access_token: str = None` here? Consistent with the documentation below.

If you don't want to mention authorization related arguments here but rather keep them all in `**kwargs`, both `access_token` and `oauth_persistence` shouldn't be mentioned. Having one but not the other is inconsistent.

src/databricks/sql/client.py
```diff
@@ -90,25 +86,7 @@ def __init__(
...
-            ]
-        elif access_token:
-            authorization_header = [("Authorization", "Bearer {}".format(access_token))]
-        elif not (
-            kwargs.get("_use_cert_as_auth") and kwargs.get("_tls_client_cert_file")
-        ):
-            raise ValueError(
-                "No valid authentication settings. Please provide an access token."
-            )
+        auth_provider = get_python_sql_connector_auth_provider(server_hostname, oauth_persistence, **kwargs)
```
pietern, line 89: Per command above, either pass all arguments explicitly, or rely on `**kwargs` for all.

src/databricks/sql/experimental/oauth_persistence.py
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        # Serializing json
+        json_object = json.dumps(dictionary, indent=4)
+
+        with open(self._file_path, "w") as outfile:
+            outfile.write(json_object)
+
+    def read(self) -> OAuthToken:
+        # TODO: validate the
```
pietern, line 48: Broken comment.

src/databricks/sql/auth/auth.py
```diff
@@ -0,0 +1,96 @@
...
+OAUTH_CLIENT_ID = "databricks-sql-python"
+OAUTH_REDIRECT_PORT_RANGE = range(8020, 8025)
+
+def get_python_sql_connector_auth_provider(hostname: str, oauth_persistence: OAuthPersistence = None, **kwargs):
```
pietern, line 82: Per the other comments,  `oauth_persistence` should be lifted from `**kwargs` as well.

pietern (commented):
Couple of comments around `**kwargs` vs explicitly named keyword arguments, for consistency.

src/databricks/sql/auth/oauth_http_handler.py
```diff
@@ -0,0 +1,42 @@
+from http.server import BaseHTTPRequestHandler
```
susodapop, line 1: I expect some of this will be explained in the design doc: but why does `databricks-sql-connector` need to run an HTTP server? Is this for the browser to redirect a callback URL after the user successfully authenticates?

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+# Private API: this is an evolving interface and it will change in the future.
+# Please must not depend on it in your applications.
+from databricks.sql.experimental.oauth_persistence import OAuthToken, OAuthPersistence
+
+
+class CredentialsProvider:
+    def add_headers(self, request_headers: Dict[str, str]):
+        pass
```
susodapop, line 36: Ditto previous comment. 
```suggestion
        raise NotImplementedError()
```

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+
+import thrift.transport.THttpClient
```
susodapop, line 25: Nit: It feels strange to use the fully qualified name in two places (here and line 30). You can instead import `thrift` here.
```suggestion
import thrift
```

or 
```suggestion
from thrift.transport.THttpClient import THttpClient
```

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -0,0 +1,46 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class THttpClient(thrift.transport.THttpClient.THttpClient):
```
susodapop, line 30: Nit: no type hints for this Class?

src/databricks/sql/auth/thrift_http_client.py
susodapop: I think retry behaviour is handled in `thrift_backend`, no?

src/databricks/sql/auth/auth.py
moderakh: thanks addressed. oauth client-id and redirect urls are registered in the service side. addressed.

src/databricks/sql/auth/oauth.py
moderakh: addressed. iterates through the first available port from [8020, 8025)

src/databricks/sql/experimental/oauth_persistence.py
```diff
@@ -0,0 +1,56 @@
...
+logger = logging.getLogger(__name__)
+
+
+class OAuthToken:
+    def __init__(self, access_token, refresh_token):
+        self._access_token = access_token
+        self._refresh_token = refresh_token
+
+    def get_access_token(self) -> str:
+        return self._access_token
```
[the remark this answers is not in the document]
moderakh: Thanks @susodapop this is very helpful.

src/databricks/sql/auth/oauth_http_handler.py
moderakh: correct. Interactive User to Machine OAuth code flow requires the oauth code to be returned on the redirect url where an http server is listening. This class provides the implementation which listens for oauth code on the configured redirect url (which should be http://localhsot:8020 or other ports if this one taken.)

src/databricks/sql/auth/thrift_http_client.py
moderakh: thanks addressed.

src/databricks/sql/__init__.py
moderakh: thanks for the comment. addressed.

src/databricks/sql/auth/auth.py
moderakh: thanks, addressed.

src/databricks/sql/experimental/oauth_persistence.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def get_refresh_token(self) -> str:
+        return self._refresh_token
+
+
+class OAuthPersistence:
+    def persist(self, oauth_token: OAuthToken):
+        pass
```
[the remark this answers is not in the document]
moderakh: This is actually not meant to be an interface but a base class with no op.

what's the use case?
the use case is when the existing `_use_cert_as_auth` and `_tls_client_cert_file` are used, in that scenario we don't add any auth additional header to the request.

src/databricks/sql/auth/authenticators.py
moderakh: ditto as above.

This is actually not meant to be an interface but a base class with no op.

what's the use case?
the use case is when the existing `_use_cert_as_auth` and `_tls_client_cert_file` are used, in that scenario we don't add any auth additional header to the request.

src/databricks/sql/client.py
moderakh: addressed. thanks.

src/databricks/sql/auth/thrift_http_client.py
moderakh: not a real TODO. removed thanks.

src/databricks/sql/client.py
moderakh: thanks for the comment moved both to `**kwargs`

src/databricks/sql/auth/authenticators.py
moderakh: done.

moderakh:
> It's unusual for files in this repository to all include the APACHE licence since this is already covered by our ./LICENSE file.

Thanks @susodapop , I removed the Apache license from the new files to be consistent with the existing files. Addressed.

src/databricks/sql/__init__.py
```diff
@@ -44,7 +44,7 @@ def TimestampFromTicks(ticks):
     return Timestamp(*time.localtime(ticks)[:6])
 
 
-def connect(server_hostname, http_path, access_token, **kwargs):
+def connect(server_hostname, http_path, **kwargs):
```
pietern, line 47: If folks have called this function without keyword arguments in the path, e.g.
```python
connect("hostname", "path", "token")
```

This modification would break that.

Might be fine, but worth checking. @susodapop WDYT?

pietern (approved):
Last comment, rest looks good. Accepting to unblock progress.

nfx (approved):
LGTM 👍

src/databricks/sql/__init__.py
moderakh: thanks @pietern for catching this. will fix it shortly.

moderakh: addressed

src/databricks/sql/__init__.py
```diff
@@ -44,7 +44,6 @@ def TimestampFromTicks(ticks):
     return Timestamp(*time.localtime(ticks)[:6])
 
 
-def connect(server_hostname, http_path, access_token, **kwargs):
+def connect(server_hostname, http_path, access_token=None, **kwargs):
```
benfleis, line 47: Thinking out loud -- this param being called `access_token` may conflate it with OAuth access tokens, and possibly appear to exclude PAT tokens.

If the earlier assertion is correct that it's used as a required positional param, is this the ideal moment to rename to `auth_token` to avoid the PAT conflation and make consistent with the attached docs/logs?

src/databricks/sql/auth/authenticators.py
```diff
@@ -0,0 +1,118 @@
...
+from typing import Dict, List
+
+from databricks.sql.auth.oauth import OAuthManager
+
+# Private API: this is an evolving interface and it will change in the future.
+# Please must not depend on it in your applications.
+from databricks.sql.experimental.oauth_persistence import OAuthToken, OAuthPersistence
+
+
+class CredentialsProvider:
```
sander-goos, line 12: nit: would `AuthProvider` be more consistent with the other names? The implementations all have this as suffix, and the field in which it is stored is also called: `auth_provider`

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -0,0 +1,35 @@
...
+        self.__auth_provider = auth_provider
+
+    def setCustomHeaders(self, headers: Dict[str, str]):
+        self._headers = headers
+        super().setCustomHeaders(headers)
+
+    def flush(self):
+        headers = dict(self._headers)
+        self.__auth_provider.add_headers(headers)
+        self._headers = headers
```
sander-goos, line 33: nit: setting `self._headers` is also done by `setCustomHeaders`

sander-goos (approved):
Great stuff, thanks for implementing this! Added some nits/questions

src/databricks/sql/auth/thrift_http_client.py
moderakh: `self.__auth_provider.add_headers()` adds the auth headers. `headers` here contains auth token and is not the same as what was set by the `setCustomHeaders()`

src/databricks/sql/client.py
```diff
@@ -36,15 +33,77 @@ def __init__(
...
+            :param catalog: An optional initial catalog to use. Requires DBR version 9.0+
+            :param schema: An optional initial schema to use. Requires DBR version 9.0+
+
+        Other Parameters:
+            auth_type: `str`, optional
+                `databricks-oauth` : to use oauth with fine-grained permission scopes, set to `databricks-oauth`.
+                This is currently in private preview for Databricks accounts on AWS.
+                This is a beta feature and supports User to Machine OAuth authentication for Databricks on AWS with
+                any IDP configured. This is only for interactive python applications and open a browser window.
+                Note this is beta (private preview)
```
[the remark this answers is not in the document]
moderakh: Thanks addressed.

src/databricks/sql/client.py
```diff
@@ -36,15 +33,77 @@ def __init__(
...
+
+            experimental_oauth_persistence: configures preferred storage for persisting oauth tokens.
```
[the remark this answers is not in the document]
moderakh: We do not have dependency on Databricks CLI. We rely on the end user to implement the token persistence layer. ecosystem dev team is planning to build configuration management in the python core sdk. once that's ready we can also use that as option.

src/databricks/sql/auth/auth.py
```diff
@@ -0,0 +1,96 @@
...
+    else:
+        raise RuntimeError("No valid authentication settings!")
+
+
+PYSQL_OAUTH_SCOPES = ["sql", "offline_access"]
+PYSQL_OAUTH_CLIENT_ID = "databricks-sql-python"
+PYSQL_OAUTH_REDIRECT_PORT_RANGE = list(range(8020, 8025))
+
+
+def normalize_host_name(hostname: str):
```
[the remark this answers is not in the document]
moderakh: that one slightly different constructs the oidc url

src/databricks/sql/auth/authenticators.py
moderakh: renamed.

src/databricks/sql/auth/thrift_http_client.py
sander-goos: I meant that `self.setCustomHeaders(self._headers)` is called at line 34 (after the auth headers have been added), which also sets `self._headers`

src/databricks/sql/auth/auth.py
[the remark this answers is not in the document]
sander-goos: I see. On a second look, this doesn't take into account the `port` (or `_connection_uri`) parameters. (Also, we can assume `server_hostname` never contains a scheme or trailing slash as otherwise the connection will fail: https://github.com/databricks/databricks-sql-python/blob/main/src/databricks/sql/thrift_backend.py#L105-L107 )

Brings up another question: why is this function actually needed at this point? It seems confusing as the field name is called `hostname` whereas it really is storing a uri. The usages of this field in `DatabricksOAuthProvider` (oauth_persistence, oauth_manager) seem to work fine with just a hostname (although again they would be missing the `port`).

PR #18, opened 2022-07-15
Make GetOperationStatus attempts retryable

susodapop:
## Description

Currently, we only retry attempts that returned a code 429 or 503 and include a `Retry-After` header. This pull request refactors the `extract_retry_delay` code to allow retrying of `GetOperationStatus` requests even if `Retry-After` header was not sent from the server. The configured retry policy is still honoured with regard to maximum attempts and max retry duration.

## Background

The reason we only retry 429 and 503 responses today is because retries must be idempotent. Otherwise the connector could cause unexpected or harmful consequences (data loss, excess resource utilisation etc.)

We know that 429/503 responses are idempotent because the attempt was halted before the server could execute it, regardless if the attempted operation was itself idempotent. 

By the same logic we know that `GetOperationStatus` is idempotent. No matter how many times it runs it does not change data on the server. Therefore it can be safely added to the allow list.

### DIFF
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -31,6 +31,8 @@ DATABRICKS_REASON_HEADER = "x-databricks-reason-phrase"
 
 TIMESTAMP_AS_STRING_CONFIG = "spark.thriftserver.arrowBasedRowSet.timestampAsString"
 
+DEFAULT_RETRY_DELAY_AFTER = 5
+
 # see Connection.__init__ for parameter descriptions.
 # - Min/Max avoids unsustainable configs (sane values are far more constrained)
 # - 900s attempts-duration lines up w ODBC/JDBC drivers (for cluster startup > 10 mins)
@@ -260,18 +262,36 @@ class ThriftBackend:
         def get_elapsed():
             return time.time() - t0
 
+        def make_bounded_delay(attempt: int, delay: int):
+            """Bound delay (seconds) by [min_delay*1.5^(attempt-1), max_delay]"""
+            delay = max(delay, self._retry_delay_min * math.pow(1.5, attempt - 1))
+            delay = min(delay, self._retry_delay_max)
+            return delay
+
         def extract_retry_delay(attempt):
-            # encapsulate retry checks, returns None || delay-in-secs
-            # Retry IFF 429/503 code + Retry-After header set
+            """Returns Retry-After header contents IFF 429/503 code + Retry-After header set"""
             http_code = getattr(self._transport, "code", None)
             retry_after = getattr(self._transport, "headers", {}).get("Retry-After")
             if http_code in [429, 503] and retry_after:
-                # bound delay (seconds) by [min_delay*1.5^(attempt-1), max_delay]
-                delay = int(retry_after)
-                delay = max(delay, self._retry_delay_min * math.pow(1.5, attempt - 1))
-                delay = min(delay, self._retry_delay_max)
-                return delay
+                return make_bounded_delay(attempt, int(retry_after))
             return None
+
+        def get_retry_delay(attempt):
+            """Encapsulate retry checks. Returns None || delay-in-secs
+
+            Retry if
+              - 429/503 code + Retry-After header set
+              - method is GetOperationStatus
+
+            Retry-After header is always used if present. If not present but method is GetOperationStatus
+            the DEFAULT_RETRY_DELAY_AFTER is used. This catches connection timeouts during GetOperationStatus.
+            """
+
+            header_delay = extract_retry_delay(attempt)
+            if method.__name__ == "GetOperationStatus" and not header_delay:
+                return make_bounded_delay(attempt, DEFAULT_RETRY_DELAY_AFTER)
+
+            return header_delay
 
         def attempt_request(attempt):
             # splits out lockable attempt, from delay & retry loop
@@ -285,7 +305,7 @@ class ThriftBackend:
                 logger.debug("Received response: {}".format(response))
                 return response
             except Exception as error:
-                retry_delay = extract_retry_delay(attempt)
+                retry_delay = get_retry_delay(attempt)
                 error_message = ThriftBackend._extract_error_message_from_headers(
                     getattr(self._transport, "headers", {})
                 )
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -15,6 +15,9 @@ from thrift.Thrift import TException
 
 from databricks.sql.thrift_api.TCLIService import TCLIService, ttypes
 from databricks.sql import *
+from databricks.sql.thrift_api.TCLIService.TCLIService import (
+    Client as TCLIServiceClient,
+)
 from databricks.sql.utils import (
     ArrowQueue,
     ExecuteResponse,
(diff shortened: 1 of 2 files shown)


tests/test_thrift_backend.py
```diff
@@ -968,6 +968,42 @@ def test_handle_execute_response_sets_active_op_handle(self):
...
+            443,
+            "path", [],
+            _retry_stop_after_attempts_count=2,
+            _retry_delay_max=0,
+            _retry_delay_min=0)
+
+        with self.assertRaises(RequestError) as cm:
+            thrift_backend.make_request(client.GetOperationStatus, req)
+
+        self.assertEqual(NoRetryReason.OUT_OF_ATTEMPTS.value, cm.exception.context["no-retry-reason"])
```
susodapop, line 1004: Prior to this change the test looked like this:
```suggestion
        self.assertEqual(NoRetryReason.NOT_RETRYABLE.value, cm.exception.context["no-retry-reason"])
```

susodapop:
I rebased off of main to resolve the merge conflict since we moved unit tests into a subdirectory of `./tests`

yunbodeng-db (approved):

src/databricks/sql/thrift_backend.py
```diff
@@ -31,6 +34,8 @@
 
 TIMESTAMP_AS_STRING_CONFIG = "spark.thriftserver.arrowBasedRowSet.timestampAsString"
 
+DEFAULT_RETRY_DELAY_AFTER = 5
```
benfleis, line 37: nit thought: should this be in the _retry_policy below? It's not bounds, but could be a `_retry_delay_default = (float, 5, None, None)` named entry. Placing it there gives a clear path for the savy/hacky user to set a new default in a blessed, per-connection way.

benfleis (approved):
1 nit thought, further LGTM

src/databricks/sql/thrift_backend.py
```diff
@@ -260,19 +265,40 @@ def make_request(self, method, request):
...
+              - 429/503 code + Retry-After header set
+              - method is GetOperationStatus
+
+            Retry-After header is always used if present. If not present but method is GetOperationStatus
+            the DEFAULT_RETRY_DELAY_AFTER is used. This catches connection timeouts during GetOperationStatus.
+            """
+
+            header_delay = extract_retry_delay(attempt)
+            if (
+                method.__name__ == TCLIServiceClient.GetOperationStatus.__name__
```
sander-goos, line 295: nit: perhaps it's slightly cleaner to check on the type of the request instead of reflecting on the method name?

tests/unit/test_thrift_backend.py
```diff
@@ -968,6 +968,42 @@ def test_handle_execute_response_sets_active_op_handle(self):
...
+            self, mock_retry_policy, t_transport_class):
+       
+        import thrift
+        from databricks.sql.thrift_api.TCLIService import TCLIService
+        from databricks.sql.exc import RequestError
+        from databricks.sql.utils import NoRetryReason
+
+        t_transport_instance = t_transport_class.return_value
+        t_transport_instance.code = 200
+        t_transport_instance.headers = {"foo": "bar"}
```
sander-goos, line 983: I might just miss it, but where do we make sure an error is raised? As that is the point of this test, I'd make that explicit.

src/databricks/sql/thrift_backend.py
```diff
@@ -260,19 +265,40 @@ def make_request(self, method, request):
         def get_elapsed():
             return time.time() - t0
 
+        def make_bounded_delay(attempt: int, delay: int):
```
sander-goos, line 268: Don't block on this, but IMO this `make_request` method is getting a bit unwieldy with all the various inner methods. It would be good if we could put this retry logic in a separate module/class in a follow up, WDYT?

src/databricks/sql/thrift_backend.py
```diff
@@ -260,19 +265,40 @@ def make_request(self, method, request):
...
+            """Bound delay (seconds) by [min_delay*1.5^(attempt-1), max_delay]"""
+            delay = max(delay, self._retry_delay_min * math.pow(1.5, attempt - 1))
+            delay = min(delay, self._retry_delay_max)
+            return delay
+
         def extract_retry_delay(attempt):
-            # encapsulate retry checks, returns None || delay-in-secs
-            # Retry IFF 429/503 code + Retry-After header set
+            """Returns Retry-After header contents IFF 429/503 code + Retry-After header set"""
```
sander-goos, line 275: As we have learned that the THttpClient doesn't seem reset the fields for code and headers before each request, we can read values from a previous request in case an exception occurs before `flush` is reached. See: https://github.com/apache/thrift/blob/7a5f29d48d6b99f14190d9d7bff0e63170965087/lib/py/src/transport/THttpClient.py#L191-L194
I think your fix will still work fine, but we might want to reset these explicitly to prevent potential confusion later.

sander-goos (approved):
Overall LTGM, left some non-blocking nits/questions

src/databricks/sql/thrift_backend.py
susodapop: I wondered about this but don't see an obvious way to assert it. The type of operation is not reflected in Python's type system. 

```python
>>> type(TCLIServiceClient.GetOperationStatus)
<class 'function'>

>>> type(method)
<class 'method'>

>>> isinstance(method, TCLIServiceClient.GetOperationStatus)
False
```

I don't know a cleaner way to check for this condition.

src/databricks/sql/thrift_backend.py
susodapop: Applied in [2e80e44](https://github.com/databricks/databricks-sql-python/pull/18/commits/2e80e4461822c382dbd40ccca24371daf3506fa4).

tests/unit/test_thrift_backend.py
susodapop: Good catch. I updated the test case in 6ed85df867b48905a8a22634ee307cb4b0084631

Technically, this fix does not require that `GetOperationStatus` raises an exception. Only that no `Retry-After` header is present in the response from HttpTransport. But since this change was initially prompted by errors due to connection timeouts it makes sense to introduce that as the mechanism by which the request fails.

src/databricks/sql/thrift_backend.py
susodapop: Thank you. I'll make a follow-up PR that does this.

src/databricks/sql/thrift_backend.py
susodapop: Agreed. I'll file a ticket for this.

susodapop:
I was prepared to merge this change until I ran our e2e test suite and found it caused our `test_cancel_during_execute` e2e test to fail. This happened because that test calls `GetOperationStatus` on a canceled query and receives a `THIS OBJECT DOES NOT EXIST` error. However, because this code change makes _all_ `GetOperationStatus` requests retryable, the client kept polling for a status.

This draws attention to a scneario I hadn't tested for: there is a set of exceptions that may be raised during `GetOperationStatus` that should not be retried. `THIS OBJECT DOES NOT EXIST` is one such exception. But I don't have a full list of the others.

So rather than add (incomplete) conditional logic to this implementation, I rewrote it to simply test explicitly for `socket.timeout` and `OSError` exceptions.

PR #24, opened 2022-07-28
Retry attempts that fail due to a connection timeout

susodapop:
### Note for reviewers

On 4 August 2022 I reverted all changes in this PR so I could reimplement and apply all your review feedback. This happened in fa1fd50c4388abd2ac41727b50a90d522e8e98c5. Every subsequent commit encapsulates one logical change to the code. Working through them one-at-a-time should be quite easy.

1. Isolate delay bounding bf65b81df48cd9ec718c84ff6f1d20e47864c546
2. Move error details scope up one level a0d340e659f8e142959cb11bbab7cb13bf03651f (H/T @moderakh )
3. Retry `GetOperationStatus` for `OSError` a55cf9dc1af85a750d1bc71ccb5dfe5be4dfbaf2 (thanks @sander-goos and @benfleis )
4. Log when we do it 38411a8c0ed8db59d403ab24bd82a4e1dee62da1 (H/T @benfleis )

The final four commits are simple code cleanup and fixing one warning in the test suite un-related to this change.

Since our e2e tests are not enabled via Github Actions yet I ran them locally and all passed.

## Description
Currently, we only retry attempts that returned a code 429 or 503 and include a Retry-After header. This pull request also allows `GetOperationStatus` requests to be retried if the request fails with an `OSError` exception. The configured retry policy is still honoured with regard to maximum attempts and max retry duration.

## Background
The reason we only retry 429 and 503 responses today is because retries must be idempotent. Otherwise the connector could cause unexpected or harmful consequences (data loss, excess resource utilisation etc.)

We know that 429/503 responses are idempotent because the attempt was halted before the server could execute it, regardless if the attempted operation was itself idempotent.

We also know that `GetOperationStatus` requests are idempotent because they do not modify data on the server. Therefore we can add an extra case to our retry allow list:
- If a request is a `GetOperationStatus` command
- If it fails due to an `OSError` such as `TimeoutError` or `ConnectionResetError`.
 
Previously we attempted this same behaviour by retrying `GetOperationStatus` requests, regardless the nature of the exception. But this change could not pass our e2e tests because there are valid cases when `GetOperationStatus` will raise an exception from within our own library code: for example, if an operation is canceled in a separate thread `GetOperationStatus` will raise a "DOES NOT EXIST" exception.

## Logging Behaviour

The connector will log whenever it retries an attempt because of an `OSError`. It will use log level `INFO` if the `OSError` is one we consider normal. It will use log level `WARNING` if the `OSError` seems unusual. The codes we consider normal are:

```python
                                            # | Debian | Darwin |
                    info_errs = [           # |--------|--------|         
                        errno.ESHUTDOWN,    # |   32   |   32   |
                        errno.EAFNOSUPPORT, # |   97   |   47   |
                        errno.ECONNRESET,   # |   104  |   54   |
                        errno.ETIMEDOUT,    # |   110  |   60   |
                    ]
```

The full set of `OSError` codes is platform specific. I wrote this patch to target a specific customer scenario when `GetOperationStatus` requests were retried after an operating system socket timeout exception. In this customer scenario the error code was `Errno 110: Connection timed out`. However that error code is specific to Linux. On a Darwin/Macos host the code would be `65` and on Windows it would be `10060`.

Rather than catch these specifically, I use Python's `errno` built-in to check for `errno.ETIMEDOUT` which will resolve to the platform-specific code at runtime. I tested this manually on Linux and MacOS (but not on Windows). @benfleis helped me pick what we consider "normal". 

We log all other `OSError` codes with `WARNING` is because it would be pretty unusual for a request to fail because of a "FileNotFound" or a system fault.

### References

I found [this article](https://aakinshin.net/posts/how-socket-error-codes-depend-on-runtime-and-operating-system/) extremely helpful while formulating this fix. The author is solving a very similar problem across platforms.

### DIFF
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -290,8 +290,17 @@ class ThriftBackend:
                 response = method(request)
                 logger.debug("Received response: {}".format(response))
                 return response
-            except (socket.timeout, OSError) as err:
+            except socket.timeout as err:
+                # socket.timeout means the connection was timed out by the socket package
                 retry_delay = self._retry_delay_default
+                error_message = str(err)
+                error = err
+            except OSError as err:
+                # OSError 110 means the connection was timed out by the operating system
+                if "Errno 110" in str(err):
+                    retry_delay = self._retry_delay_default
+                else:
+                    retry_delay = None
                 error_message = str(err)
                 error = err
             except Exception as err:
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -320,9 +320,7 @@ class ThriftBackend:
                     ]
 
                     # fmt: on
-                    log_string = (
-                        f"{gos_name} failed with code {err.errno} and will attempt to retry"
-                    )
+                    log_string = f"{gos_name} failed with code {err.errno} and will attempt to retry"
                     if err.errno in info_errs:
                         logger.info(log_string)
                     else:
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -247,7 +247,7 @@ class ThriftBackend:
     # FUTURE: Consider moving to https://github.com/litl/backoff or
     # https://github.com/jd/tenacity for retry logic.
     def make_request(self, method, request):
-        """Execute given request, attempting retries when receiving HTTP 429/503.
+        """Execute given request, attempting retries when TCP connection fils or when receiving HTTP 429/503.
 
         For delay between attempts, honor the given Retry-After header, but with bounds.
         Use lower bound of expontial-backoff based on _retry_delay_min,
@@ -265,8 +265,12 @@ class ThriftBackend:
             return time.time() - t0
 
         def extract_retry_delay(attempt):
-            # encapsulate retry checks, returns None || delay-in-secs
-            # Retry IFF 429/503 code + Retry-After header set
+            """
+            Encapsulate retry checks based on HTTP headers. Returns None || delay-in-secs
+
+            Retry IFF 429/503 code + Retry-After header set
+            """
+
             http_code = getattr(self._transport, "code", None)
             retry_after = getattr(self._transport, "headers", {}).get("Retry-After")
             if http_code in [429, 503] and retry_after:
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -311,6 +311,10 @@ class ThriftBackend:
                     # The built-in errno package encapsulates OSError codes, which are OS-specific.
                     # log.info for errors we believe are not unusual or unexpected. log.warn for
                     # for others like EEXIST, EBADF, ERANGE which are not expected in this context.
+                    #
+                    # I manually tested this retry behaviour using mitmweb and confirmed that 
+                    # GetOperationStatus requests are retried when I forced network connection
+                    # interruptions / timeouts / reconnects. See #24 for more info.
                                             # | Debian | Darwin |
                     info_errs = [           # |--------|--------|         
                         errno.ESHUTDOWN,    # |   32   |   32   |
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -284,24 +284,27 @@ class ThriftBackend:
             # - non-None method_return -> success, return and be done
             # - non-None retry_delay -> sleep delay before retry
             # - error, error_message always set when available
+            
+            error, error_message, retry_delay = None, None, None
             try:
                 logger.debug("Sending request: {}".format(request))
                 response = method(request)
                 logger.debug("Received response: {}".format(response))
                 return response
-            except Exception as error:
+            except Exception as err:
+                error = err
                 retry_delay = extract_retry_delay(attempt)
                 error_message = ThriftBackend._extract_error_message_from_headers(
                     getattr(self._transport, "headers", {})
                 )
-                return RequestErrorInfo(
-                    error=error,
-                    error_message=error_message,
-                    retry_delay=retry_delay,
-                    http_code=getattr(self._transport, "code", None),
-                    method=method.__name__,
-                    request=request,
-                )
+            return RequestErrorInfo(
+                error=error,
+                error_message=error_message,
+                retry_delay=retry_delay,
+                http_code=getattr(self._transport, "code", None),
+                method=method.__name__,
+                request=request,
+            )
 
         # The real work:
         # - for each available attempt:
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -1,4 +1,5 @@
 from decimal import Decimal
+import errno
 import logging
 import math
 import time
@@ -304,6 +305,27 @@ class ThriftBackend:
                 gos_name = TCLIServiceClient.GetOperationStatus.__name__
                 if method.__name__ == gos_name:
                     retry_delay = bound_retry_delay(attempt, self._retry_delay_default)
+
+                    # fmt: off
+                    # The built-in errno package encapsulates OSError codes, which are OS-specific.
+                    # log.info for errors we believe are not unusual or unexpected. log.warn for
+                    # for others like EEXIST, EBADF, ERANGE which are not expected in this context.
+                                            # | Debian | Darwin |
+                    info_errs = [           # |--------|--------|         
+                        errno.ESHUTDOWN,    # |   32   |   32   |
+                        errno.EAFNOSUPPORT, # |   97   |   47   |
+                        errno.ECONNRESET,   # |   104  |   54   |
+                        errno.ETIMEDOUT,    # |   110  |   60   |
+                    ]
+
+                    # fmt: on
+                    log_string = (
+                        f"{gos_name} failed with code {err.errno} and will attempt to retry"
+                    )
+                    if err.errno in info_errs:
+                        logger.info(log_string)
+                    else:
+                        logger.warning(log_string)
             except Exception as err:
                 error = err
                 retry_delay = extract_retry_delay(attempt)
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -5,6 +5,7 @@ import time
 import threading
 from uuid import uuid4
 from ssl import CERT_NONE, CERT_OPTIONAL, CERT_REQUIRED, create_default_context
+import socket
 
 import pyarrow
 import thrift.transport.THttpClient
@@ -279,24 +280,31 @@ class ThriftBackend:
             # - non-None method_return -> success, return and be done
             # - non-None retry_delay -> sleep delay before retry
             # - error, error_message always set when available
+
+            error = None
             try:
                 logger.debug("Sending request: {}".format(request))
                 response = method(request)
                 logger.debug("Received response: {}".format(response))
                 return response
-            except Exception as error:
+            except (socket.timeout, OSError) as err:
+                retry_delay = 5
+                error_message = str(err)
+                error = err
+            except Exception as err:
                 retry_delay = extract_retry_delay(attempt)
                 error_message = ThriftBackend._extract_error_message_from_headers(
                     getattr(self._transport, "headers", {})
                 )
-                return RequestErrorInfo(
-                    error=error,
-                    error_message=error_message,
-                    retry_delay=retry_delay,
-                    http_code=getattr(self._transport, "code", None),
-                    method=method.__name__,
-                    request=request,
-                )
+                error = err
+            return RequestErrorInfo(
+                error=error,
+                error_message=error_message,
+                retry_delay=retry_delay,
+                http_code=getattr(self._transport, "code", None),
+                method=method.__name__,
+                request=request,
+            )
 
         # The real work:
         # - for each available attempt:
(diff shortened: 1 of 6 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -279,24 +283,40 @@ def attempt_request(attempt):
...
+                # socket.timeout means the connection was timed out by the socket package
+                retry_delay = self._retry_delay_default
+                error_message = str(err)
+                error = err
+            except OSError as err:
+                # OSError 110 means the connection was timed out by the operating system
+                if "Errno 110" in str(err):
+                    retry_delay = self._retry_delay_default
+                else:
+                    retry_delay = None
```
yunbodeng-db, line 303: Can you initialize this at the beginning and can only be overridden on selected (retriable) errors?

src/databricks/sql/thrift_backend.py
```diff
@@ -279,24 +283,40 @@ def attempt_request(attempt):
...
+                error_message = str(err)
+                error = err
+            except Exception as err:
                 retry_delay = extract_retry_delay(attempt)
```
yunbodeng-db, line 307: What does this mean? It means retry on exceptions in general?

yunbodeng-db (approved):

src/databricks/sql/thrift_backend.py
susodapop: Sure. c76ee65ee674a3c1fcd1b45417b0493bcccdf899

src/databricks/sql/thrift_backend.py
susodapop: No, if `retry_delay` is `None` then the request is not retried. `extract_retry_delay` returns `None` in all cases except for a 429/503 code with `Retry-After` headers sent.

In general, requests are not considered retryable.

Meta point: this logic needs a full reworking for clarity as part of a follow-up ticket. The retry logic is only going to become more complicated from here as we unify the python connector behaviour with the ODBC behaviour

susodapop: The reason for these extra `except` statements is for the case when a request should be retried but won't have `Retry-After` headers because a response was never received from the server.

yunbodeng-db (approved):

src/databricks/sql/thrift_backend.py
```diff
@@ -279,24 +287,41 @@ def attempt_request(attempt):
...
+
+            # If retry_delay is None the request is treated as non-retryable
+            retry_delay = None
             try:
                 logger.debug("Sending request: {}".format(request))
                 response = method(request)
                 logger.debug("Received response: {}".format(response))
                 return response
-            except Exception as error:
+            except socket.timeout as err:
```
moderakh, line 300: do we retry on socket.timeout?

if so, how do we ensure idempotency of write requests?

src/databricks/sql/thrift_backend.py
```diff
@@ -279,24 +287,41 @@ def attempt_request(attempt):
...
+                # socket.timeout means the connection was timed out by the socket package
+                retry_delay = self._retry_delay_default
+                error_message = str(err)
+                error = err
+            except OSError as err:
```
moderakh, line 305: As this is Operating system error, is the semantic of the error (code, message, etc) the same on different OSs (linux, win)?

src/databricks/sql/thrift_backend.py
susodapop: I answered this exact question in the pull request description:

> By the same logic we know that requests which fail with socket.timeout are idempotent because no connection was made to the server. No matter how many times it runs it does not change data on the server. Therefore it can be safely added to the allow list

src/databricks/sql/thrift_backend.py
susodapop: This is not guaranteed for _all_ operating systems, but for UNIX and Windows the codes are the same. See [Microsoft's Errno documentation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/errno-constants?view=msvc-170)

> Some of the values in ERRNO.H are present to maintain compatibility with the UNIX family of operating systems. The errno values in a 32-bit Windows operating system are a subset of the values for errno in UNIX systems.

Therefore `Errno 110` is the correct `EHOSTUNREACH` code for any UNIX system + Windows

susodapop: After we add HTTP proxy support I would like to establish multi-platform e2e tests which effectively duplicate an `EHOSTUNREACH` code without any mocks. Until then, it is possible that platform-specific formatting of socket timeout exceptions could be missed by this Exception handler.

In such case, the default is to not retry.

src/databricks/sql/thrift_backend.py
susodapop: The more I've researched the more I thought about this I think there is still a gap, like you identified. 

Errno 110 always means EHOSTUNREACHABLE which is de facto idemptoent. 

But socket.timeout can be raised by any socket operation, not just connect(). So I'll need some further rewriting to specifically check what kind of socket operation failed.

If the socket operation that failed is a connection, then we can retry the request because we know nothing was transmitted to the server. 

But if it was a send / write operation, we can't guarantee a command wasn't sent to the server and therefore shouldn't retry.

src/databricks/sql/thrift_backend.py
sander-goos: Other question: in what cases do we _not_ want to retry this. E.g. can it also happen when a user specifies an incorrect host? In that case we probably don't want to retry (/hang) for 5*30 = 150s ?

src/databricks/sql/thrift_backend.py
sander-goos: +1
If figuring out what socket operation was done is hard, we could also restrict retries to GetOperationStatus here, like in the previous pr.

sander-goos (commented):
Left some comments. Could you please link in the description the resources that assures that the errors we catch are safe to retry (and platform independent).

src/databricks/sql/thrift_backend.py
susodapop: That's exactly what I've done @sander-goos. I updated the PR description but tl;dr we no retry under exactly three scenarios:

- HTTP Response 401 from server
- HTTP Response 503 from server
- `OSError` was raised by system during `GetOperationStatus` request

src/databricks/sql/thrift_backend.py
susodapop: Good points @sander-goos 

> Besides the fact that the errorcode will map to the same error, are we also sure that the string rendering of the error is the same? (As that is what is being checked with if "Errno 110" in str(err):)

After some rework, the code now uses Python's built-in `errno` to encapsulate the error code check. We check for `errno.ETIMEDOUT` which will evaluate to `110` on Linux, `65` on MacOS, and `10060` on Windows hosts.

> Other question: in what cases do we not want to retry this. E.g. can it also happen when a user specifies an incorrect host? In that case we probably don't want to retry (/hang) for 5*30 = 150s ?

The current implementation is not picky about which `OSError`s it retries. But it will log a warning if an unexpected code is found. In the precise situation you outline, only a `GetOperationStatus` would be retried. Any other request, such as opening a connection / session or creating a cursor would re-raise the exception back to the caller immediately.

tests/e2e/driver_tests.py
```diff
@@ -360,6 +360,42 @@ def execute_really_long_query():
...
+        
+
+        thrift_backend._retry_stop_after_attempts_count = ATTEMPTS_TO_TRY
+        req = ttypes.TGetOperationStatusReq(
+            operationHandle=op_handle,
+            getProgressUpdate=False,
+        )
+
+        
+        with self.assertRaises(OperationalError) as cm:
```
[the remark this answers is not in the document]
susodapop: I actually removed the e2e tests for this behaviour because they were no more useful than unit tests. I'll add an e2e test for this scenario after we merge support for http proxies. Then we can simulate real timeouts.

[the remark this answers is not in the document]
susodapop: In the unit tests asserts on `RequestError` btw.

tests/e2e/driver_tests.py
```diff
@@ -360,6 +360,42 @@ def execute_really_long_query():
...
+            with mock.patch("socket.create_connection") as mock_create_connection:
+                mock_create_connection.side_effect = OSError("[Errno 110]: Connection timed out")
+                thrift_backend.make_request(thrift_backend._client.GetOperationStatus, req)
+
+            self.assertIn("{0}/{0}".format(ATTEMPTS_TO_TRY), cm.exception.message_with_context())
```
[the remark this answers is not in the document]
susodapop: Thanks! In the unit tests I now assert on the `context` rather than the error message 👌

tests/unit/test_thrift_backend.py
```diff
@@ -984,6 +985,63 @@ def test_make_request_wont_retry_if_headers_not_present(self, t_transport_class)
...
+            "path",
+            [],
+            _retry_stop_after_attempts_count=2,
+            _retry_delay_default=0.25
+        )
+
+        with self.assertRaises(OperationalError) as cm:
+            thrift_backend.make_request(mock_method, Mock())
+
+            self.assertIn("2/2", cm.exception.message_with_context())
```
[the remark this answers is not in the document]
susodapop: Ditto 👌

src/databricks/sql/thrift_backend.py
```diff
@@ -79,6 +81,8 @@ def __init__(
         #   next calculated pre-retry delay would go past
         #   _retry_stop_after_attempts_duration, stop now.)
         #
+        # _retry_delay_default                   (default: 5)
+        #   used when Retry-After is not specified by the server
         # _retry_stop_after_attempts_count
```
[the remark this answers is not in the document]
susodapop: I believe this is now fixed.

susodapop:
@benfleis @sander-goos I updated the PR description to reflect the present state after this week's reviews / updates.

benfleis (approved):
looks good!

Which real connection tests did you perform to validate? Related, perhaps worth adding a note to PR (or near the retry code) explaining any manual E2E tests you used to validate behavior in "real conditions", so you (and others) won't have to think so hard about what's worth doing as a 1-off manual test.

susodapop:
> Which real connection tests did you perform to validate?

I used `mitmweb` (more [here](https://mitmproxy.org/)) to intercept traffic between the connector and a SQL Warehouse. `mitmweb` allows me to force a timeout and abort a connection. Both of which raise `OSError` in Python. The expected behaviour was met: `GetOperationStatus` requests were retried within the retry policy, but other requests were not such as `ExecuteStatement`, `FetchResults`, `OpenSession`, and `CloseSession`.

This process is very manual, so I'm not yet incorporating it into the `tests/e2e/driver_tests.py` file. I will follow-up with an outline of how to automate this testing in the future.

### How I ran the tests
The actual integration test I ran looked like this:

```python
def test_make_request_will_retry_GetOperationStatus(self):
    
    import thrift, errno
    from databricks.sql.thrift_api.TCLIService.TCLIService import Client
    from databricks.sql.exc import RequestError
    from databricks.sql.utils import NoRetryReason
    from databricks.sql import client

    from databricks.sql.thrift_api.TCLIService import ttypes

    with self.cursor() as cursor:
        cursor.execute("SELECT 1")
        op_handle = cursor.active_op_handle
    
    req = ttypes.TGetOperationStatusReq(
        operationHandle=op_handle,
        getProgressUpdate=False,
    )

    EXPECTED_RETRIES = 2

    with self.cursor({"_socket_timeout": 10, "_retry_stop_after_attempts_count": 2}) as cursor:
        _client = cursor.connection.thrift_backend._client

        with self.assertRaises(RequestError) as cm:
            breakpoint() # At this point I instructed mitmweb to intercept and suspend all requests
            cursor.connection.thrift_backend.make_request(_client.GetOperationStatus, req)

        self.assertEqual(NoRetryReason.OUT_OF_ATTEMPTS.value, cm.exception.context["no-retry-reason"])
        self.assertEqual(f'{EXPECTED_RETRIES}/{EXPECTED_RETRIES}', cm.exception.context["attempt"])
```

I used this Dockerfile to make this work on my local machine

```Dockerfile
FROM python:3.7-slim-buster

RUN useradd --create-home pysql


# Ubuntu packages
RUN apt-get update && \
  apt-get install -y --no-install-recommends \
    curl \
    gnupg \
    build-essential \
    pwgen \
    libffi-dev \
    sudo \
    git-core \
    # Additional packages required for data sources:
    libssl-dev \
    libsasl2-modules-gssapi-mit && \
    apt-get clean && \
     rm -rf /var/lib/apt/lists/*

RUN sudo mkdir /usr/local/share/ca-certificates/extra
RUN curl mitm.it/cert/pem > mitmcert.pem
RUN openssl x509 -in mitmcert.pem -inform PEM -out mitm.crt
RUN sudo cp mitm.crt /usr/local/share/ca-certificates/extra/mitm.crt
RUN sudo update-ca-certificates

RUN pip install poetry --user

COPY --chown=pysql . /pysql
RUN chown pysql /pysql

WORKDIR /pysql

RUN python -m poetry install
```

And configured Docker using its `config.json` file to forward all traffic from my containers through the `mitmweb` proxy that I installed on my local environment, like this:

```json
    "proxies":
    {
      "default":
      {
        "httpProxy": "http://<my local network ip address>:8080",
        "httpsProxy": "http://<my local network ip address>:8080",
        "noProxy": "pypi.org"
      }
    }
```

The `mitmweb` proxy must be running when building the Docker container.

> Related, perhaps worth adding a note to PR (or near the retry code) explaining any manual E2E tests you used to validate behavior in "real conditions", so you (and others) won't have to think so hard about what's worth doing as a 1-off manual test.

Done here https://github.com/databricks/databricks-sql-python/pull/24/commits/10016ea254359a8c3d6ad2d92c2ae86a0efe9c0f

sander-goos:
Thanks for the fix and the detailed explanation!

PR #30, opened 2022-08-10
Sqlalchemy dev POC

overcoil:
## What type of PR is this?  

- [X] Feature

## Description

This is a barebone dialect together with an expanded integration test that builds on the initial scaffolding. The main add are primary/foreign key support (to ignore them if they're specified) and explicit support for the numeric, string and date/time types.

## How is this tested?

- [ ] Unit tests 
- [ ] E2E Tests 
- [X ] Manually
- [ ] N/A

I expanded integration.py to cover the scenario of creating and dropping a table with the types. 

PECO-174


.github/workflows/sqlalchemy-dialect.yml
```diff
@@ -0,0 +1,178 @@
...
+      #----------------------------------------------
+      #       check-out repo and set-up python
+      #----------------------------------------------
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Set up python
+        id: setup-python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
moderakh, line 25: does this work on other versions of python3 as well? any limitation on the supported python version?

.github/workflows/sqlalchemy-dialect.yml
```diff
@@ -0,0 +1,178 @@
...
+        #       import secrets from the Github env
+        #----------------------------------------------
+        env:
+          DATABRICKS_SERVER_HOSTNAME: ${{ secrets.REPOSEC_DATABRICKS_SERVER_HOSTNAME }}
+          DATABRICKS_HTTP_PATH: ${{ secrets.REPOSEC_DATABRICKS_HTTP_PATH }}
+          DATABRICKS_TOKEN: ${{ secrets.REPOSEC_DATABRICKS_TOKEN }}
+          DATABRICKS_SCHEMA: ${{ secrets.REPOSEC_DATABRICKS_SCHEMA }}
+
+        run: |
+          echo y | dbsqlcli --hostname $DATABRICKS_SERVER_HOSTNAME --http-path $DATABRICKS_HTTP_PATH --access-token $DATABRICKS_TOKEN -e "USE $DATABRICKS_SCHEMA; DROP TABLE IF EXISTS t; DROP TABLE IF EXISTS tabletest; DROP TABLE IF EXISTS integer_table;"
```
moderakh, line 78: what happens if there are concurrent PRs? are they going to use the same database name and table name and hence one PR CI may fail because of other PR CI writing to the same table?

src/databricks/sqlalchemy/requirements.py
```diff
@@ -1,18 +0,0 @@
-# Following official SQLAlchemy guide:
```
moderakh, line 1: where did this class come from? do we already have some sql-alchemy related code in the main branch?

src/databricks/sqlalchemy/dialect.py
```diff
@@ -1,52 +1,301 @@
...
+            return "DECIMAL(%(precision)s, %(scale)s)" % {
+                "precision": type_.precision,
+                "scale": type_.scale,
+            }
+
+    def visit_NUMERIC(self, type_, **kw):
+        if type_.precision is None:
+            return "DECIMAL"
+        elif type_.scale is None:
+            return "DECIMAL(%(precision)s)" % {"precision": type_.precision}
```
moderakh, line 77: do we have test covering the methods of this class?

src/databricks/sqlalchemy/dialect.py
```diff
@@ -1,52 +1,301 @@
...
+    supports_alter = True
+    supports_pk_autoincrement = False
+    supports_default_values = False
+    supports_empty_insert = False
+    supports_native_decimal = True
+    supports_native_boolean = True
+    supports_unicode_statements = True
+    supports_unicode_binds = True
+    returns_unicode_strings = True
+    description_encoding = None
```
moderakh, line 181: what's description_encoding?

src/databricks/sqlalchemy/dialect.py
```diff
@@ -1,52 +1,301 @@
...
+
 from typing import AnyStr
 
-from sqlalchemy import util, exc, types
-from sqlalchemy.engine import default
 
+# provide a way to debug
+debugbreakpoint = os.getenv("DATABRICKS_DIALECT_DEBUG") or False
 
-class DatabricksDialect(default.DefaultDialect):
```
moderakh, line 8: seems `src/databricks/sqlalchemy/dialect.py` already existed before. Where did this come from? 

did we have some sort of support for sqlalchemy dialect previously?

overcoil: The pre-existing ``src/databricks/sqlalchemy/dialect.py`` was a rough scaffolding that Jesse and I put together earlier on (see commit 0378460e0aaa272f6a38193d47c5289cc5d2b98a). I built on this scaffolding.

src/databricks/sqlalchemy/requirements.py
overcoil: This was created initially from the rough scaffolding that Jesse and I put together earlier on (commit 0378460e0aaa272f6a38193d47c5289cc5d2b98a). I ended up tossing it after I figured out that I won't be using SA's test suite.

.github/workflows/sqlalchemy-dialect.yml
overcoil: It should. Note that I cribbed from code-quality-checks.yml for this action to test out some ideas. Now that I think about it, this action won't be operative since I've since removed the Pytest test. I will remove this action completely.

(The current plan is to *not* turn on Github Action in the OSS repo; contributors are encouraged to do so and to use these though.)

.github/workflows/sqlalchemy-dialect.yml
overcoil: Refer to my answers to your other comment on this action.

src/databricks/sqlalchemy/dialect.py
overcoil: According to what I found in SA's doc (https://github.com/sqlalchemy/sqlalchemy/blob/main/doc/build/changelog/changelog_05.rst), it "is used for encoding the column name when processing the metadata. This usually defaults to utf-8." I went with "None" as for all the other dialects I studied.

src/databricks/sqlalchemy/dialect.py
overcoil: According to what I found in SA's doc (https://github.com/sqlalchemy/sqlalchemy/blob/main/doc/build/changelog/changelog_05.rst), it "is used for encoding the column name when processing the metadata. This usually defaults to utf-8." I went with "None" as for all the other dialects I studied.

src/databricks/sqlalchemy/dialect.py
overcoil: Good catch. I will revisit integration.py to include both the default and specified scales.

overcoil: Rectified.

.github/workflows/sqlalchemy-dialect.yml
overcoil: Rectified

susodapop:
Closing in favour of https://github.com/databricks/databricks-sql-python/pull/57 although I will incorporate many of these tests into that PR.

PR #31, opened 2022-08-15
Test multiple python versions & revert python 3.10 not being supported

### LINKED ISSUE #26: Add support for Python 3.10
PySQL depends on Numpy which only added support for Python 3.10 on 22 June 22 in version 1.23.0. But we’re pinned to 1.21.1 . De facto, we don’t support Py3.10.

dbaxa:
Hi, we are making use of this library and have for some time been using it in combination with newer versions of `pandas` (1.4.3) & `numpy` (1.23.1), and pyarrow (`8.0.0` , however version `9.0.0` seems fine). While version 2.0.2 of this library works for us ... version 2.0.3 does not due to the changes to the dependency specification which break its usage in python 3.10.

This fixes #26 .

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,8 +11,8 @@ include = ["CHANGELOG.md"]
 [tool.poetry.dependencies]
 python = "^3.7.1"
 thrift = "^0.13.0"
-pyarrow = "^5.0.0"
 pandas = "^1.3.0"
+pyarrow = "^9.0.0"
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.1.2"
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -3,17 +3,20 @@ on: [push]
 jobs:
   run-unit-tests:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9, "3.10"]
     steps:
       #----------------------------------------------
       #       check-out repo and set-up python
       #----------------------------------------------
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -11,8 +11,8 @@ include = ["CHANGELOG.md"]
 [tool.poetry.dependencies]
 python = "^3.7.1"
 thrift = "^0.13.0"
-pyarrow = "^5.0.0"
 pandas = "^1.3.0"
```
dbaxa, line 14: IMHO this library should drop support for python 3.7 & the update pandas to version `1.4.3`.

dbaxa:
cc @susodapop

susodapop (approved):
Excellent work. Thanks for adding the test matrix for different versions!

pyproject.toml
susodapop: Understandable. Because many Databricks users connect from older Pythons we prefer to maintain as much backwards compatibility as possible. For your use-case, is there any reason you cannot install a newer pandas? Or does `databricks-sql-connector` use pandas methods that were deprecated in newer versions?

susodapop:
I'm running the e2e test suite against this change now.

susodapop:
All e2e tests pass in the following Python versions:

- 3.7.9
- 3.8.13
- 3.9.13
- 3.10.5

pyproject.toml
dbaxa: No -^ it isn't an issue for us. It is rather that many libraries have dropped support for python 3.7 & you may wish to update pandas to a newer version at some point which will mean dropping support for python 3.7.

PR #32, opened 2022-08-15
Specify newer versions of numpy be used

dbaxa:
While pandas makes use of https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg , we need not hold the version of `numpy` in use back. By requiring a recent release - there should be no issues with regards to architectures such as arm64 on darwin or loongarch64. For python 3.7, we can use 1.21.6.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,8 +11,8 @@ include = ["CHANGELOG.md"]
 [tool.poetry.dependencies]
 python = "^3.7.1"
 thrift = "^0.13.0"
-pyarrow = "^5.0.0"
 pandas = "^1.3.0"
+pyarrow = "^9.0.0"
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.1.2"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,8 @@ include = ["CHANGELOG.md"]
 [tool.poetry.dependencies]
 python = "^3.7.1"
 thrift = "^0.13.0"
-pandas = "^1.3.0"
+pandas = "^1.3.5"
+numpy = [{version="^1.23.2", python=">=3.8"}, {version="^1.21.6", python="<3.8"}]
 pyarrow = "^9.0.0"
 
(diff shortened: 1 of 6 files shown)


dbaxa:
This follows on from https://github.com/databricks/databricks-sql-python/pull/31 .

dbaxa:
Test execution can be seen at https://github.com/dbaxa/databricks-sql-python/runs/7837710866?check_suite_focus=true .

pyproject.toml
```diff
@@ -11,7 +11,8 @@ include = ["CHANGELOG.md"]
 [tool.poetry.dependencies]
 python = "^3.7.1"
 thrift = "^0.13.0"
-pandas = "^1.3.0"
+pandas = "^1.3.5"
+numpy = [{version="^1.23.2", python=">=3.8"}, {version="^1.21.6", python="<3.8"}]
```
dbaxa, line 15: I am not sure I have done what I want here. I want to specify "use at least this version of numpy for python >= 3.8". But this ends up as 

```{':python_version < "3.8"': ['numpy>=1.21.6,<2.0.0'],
 ':python_version >= "3.8"': ['numpy>=1.23.2,<2.0.0']}
```

Which is likely fine but pandas 1.3.5 for example ends up with

```install_requires = 
        numpy>=1.17.3; platform_machine!='aarch64' and platform_machine!='arm64' and python_version<'3.10'
        numpy>=1.19.2; platform_machine=='aarch64' and python_version<'3.10'
        numpy>=1.20.0; platform_machine=='arm64' and python_version<'3.10'
        numpy>=1.21.0; python_version>='3.10'
```

dbaxa:
I am going to close this as it is likely better to use a minimum version of `numpy` according to what `pandas` wants.

pyproject.toml
dbaxa: This can be addressed by using `>=$version`, but I still feel that obtaining numpy version information via pandas is better.

PR #33, opened 2022-08-17
Add a marker to indicate Python 3.10 isn't supported

kaxil:
Currently, Python 3.10 isn't supported (GitHub Issue: https://github.com/databricks/databricks-sql-python/issues/26), but as this marker here says it works for all Python version, if a user tries to install it on Py 3.10 it fails with unhelpful messages.


susodapop:
Thanks for your contribution! To satisfy the DCO policy in our               [contributing guide](https://github.com/databricks/databricks-sql-python/blob/main/CONTRIBUTING.md)               every commit message must include a sign-off message. One or more of your commits is missing this message.               You can reword previous commit messages with an interactive rebase (`git rebase -i main`).

susodapop:
Also, agreed this should have been included in #27. That said, Python 3.10 support will be incorporated in the next release to Pypi. So changing this here isn't exactly necessary.

kaxil:
Thanks @susodapop , updated with signed commit message, Airflow Commiter and PMC member here 👋 . We are using this as dep for the Airflow-Databricks provider

>Also, agreed this should have been included in https://github.com/databricks/databricks-sql-python/pull/27. That said, Python 3.10 support will be incorporated in the next release to Pypi. So changing this here isn't exactly necessary.

Ideally, we you can release v2.0.4 with this change, this should help the current user making this mistake and then you can remove it when you actually add support for 3.10 in the next release.

susodapop:
NTMU 👋 

Sorry what I meant to say is that we're working on releasing Python 3.10 support in v2.0.4 within a number of _hours_. We're just running the e2e tests for the change now. Making a special release with your proposed one-line change wouldn't happen measurably sooner.

kaxil:
> NTMU 👋
> 
> Sorry what I meant to say is that we're working on releasing Python 3.10 support in v2.0.4 within a number of _hours_. We're just running the e2e tests for the change now. Making a special release with your proposed one-line change wouldn't happen measurably sooner.

Oh, that's awesome ❤️ . I didn't know you were planning 2.0.4 in few hours. That's good to know, we will hold off our release for https://github.com/astronomer/astronomer-providers/pull/588 too ❤️

susodapop:
v2.0.4 is now live on Pypi 🚀

kaxil:
> v2.0.4 is now live on Pypi 🚀

I can confirm it works for us now, thanks for the quick turnaround

PR #36, opened 2022-08-17
On Pypi, display the "Project Links" sidebar.

susodapop:
This PR makes it so our bug tracker and homepage are explicitly called out on the Pypi listing. I tested by pushing to test.pypi.org. It looks like this:

CleanShot 2022-08-17 at 14 47 28@2x

You can see the sample here: https://test.pypi.org/project/databricks-sql-connector/2.0.5.dev1/

**note**: test.pypi.org is periodically erased and rebuilt so the above link may break in the future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,10 @@ pytest = "^7.1.2"
 mypy = "^0.950"
 black = "^22.3.0"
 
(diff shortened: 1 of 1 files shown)


moderakh (approved):
LGTM. thanks Jesse

PR #38, opened 2022-08-22
[ES-402013] Close cursors before closing connection

susodapop:
## Description

This PR adds a unit test to confirm that when we call `close()` on a `Connection`, all of the active cursors for that connection are also closed.


susodapop:
I have confirmed all e2e tests pass locally.

src/databricks/sql/client.py
```diff
@@ -719,8 +718,8 @@ def close(self) -> None:
                 and self.connection.open
             ):
                 self.thrift_backend.close_command(self.command_id)
+                self.has_been_closed_server_side = True
```
moderakh, line 721: if we just change the order of closing connection and closing cursors (the change that you already made in _close()) that should be enough right?

why the change to `self.has_been_closed_server_side` needed here?

andreid-db: Unless there's a good reason to move it, I would leave has_been_closed_server_side in the finally block.

If you get an exception from thrift_backend.close_command, you don't know what happened. Maybe the rpc was sent and it was a timeout, maybe the connection was timed out on the backend and it is rejecting the close rpc.

andreid-db (approved):

src/databricks/sql/client.py
susodapop: If we don't move this logic there's no way to test whether the cursor was closed server side because the cursor will _always_ return `True` for this attribute. I can see wrapping it in another try block. The alternative is to not test the change at all, which isn't adviseable.

Case in point: if I move this line back to its original place then the e2e tests passes even without our change.

susodapop: Just to reiterate: setting `self.has_been_closed_server_side` is plain incorrect here. The existing behaviour sets it `True` even if we aren't certain the connection was closed. For example, if the connection has been closed we effectively assume that the cursor was closed. Current experience demonstrates this isn't a safe assumption.

moderakh: changing the business logic code for sake of testing may cause some issues. I think it is safer to have `self.has_been_closed_server_side` in the finally block. If there is no reason other than testing, I would suggest to undo this change. I wonder if there is any other way we can test this which doesn't require changing the business logic? maybe subclassing client/mocking?

andreid-db: Could you maybe have another local variable 'last_backend_exception' or something like that, and just use that to verify in test?

susodapop: I can of course sink more time into building a more intricate testing apparatus that can workaround this incorrect behaviour. It just seems nonsensical to continue setting a value that we are _certain is false for a nontrivial set of users_. If there are edge-cases that depend on this bug then we can address them separately. Now does not seem like the time to double-down on a design oversight from the past.

susodapop: > changing the business logic code for sake of testing may cause some issues.

This isn't for the sake of testing. It's squashing a bug that has been hiding in plain sight for a few months. It just so happens that a unit test was necessary to identify this bug.

> If there is no reason other than testing, I would suggest to undo this change

The main reason (besides testing) is that we are positive this value is being incorrectly set in the wild. It doesn't make any sense to double down on this bug considering that all unit + e2e tests pass. Any other option will require a more significant re-write of the business logic or constructing a series of mocks which don't effectively reproduce a live environment.

andreid-db: I think the disconnect here is because has_been_closed_server_side doesn't have a clear definition. 

I see it being used in a bunch of places as a way to early exit. And I think that is correct as is (i.e., if the connection is already closed, the cursor is effectively useless). If those early exits were using self.op_state, then it would probably be fine to have has_been_closed_server_side set in the try section.

Why do we even have has_been_closed_server_side and self.op_state?

susodapop: > Why do we even have has_been_closed_server_side and self.op_state?

It hasn't been clearly defined. I think digging through the eccentricities of this logic is beyond the scope of this work. We should revisit it later.

Since I'm outnumbered on this I took a different route with my test. Now I issue a request to the server to verify the server's `operationState` rather than trusting the client to give an accurate answer.

As before, I wrote the test into one commit _before_ the fix. The test fails when run against `main` and passes after the change in b80cb7120c4bd9433da9c9641a3113f0a3a5c7bb is applied.

andreid-db (approved):

moderakh (approved):
Thanks @susodapop for addressing the fix and addressing the comments.

src/databricks/sql/client.py
moderakh: Thanks Jesse for addressing the comment

PR #40, opened 2022-08-23
Bump version to 2.0.5 and update CHANGELOG

susodapop:
## Description

This prepares for the `2.0.5` release. You can see an example of this release on test.pypi.org here:

https://test.pypi.org/project/databricks-sql-connector/2.0.5/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 2.0.x (Unreleased)
 
+## 2.0.5 (2022-08-23)
+
+- Fix: closing a connection now closes any open cursors from that connection at the server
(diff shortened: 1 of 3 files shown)


moderakh (approved):
LGTM.

(this comment is outside of the scope of this PR as we have different format for other changelog entries and this is not a blocking comment we should think about it for future re-formatting of the changelog)

We should try to use consistent wording for all entries (Fix vs Fixed, Add, Added) I see we have different wording in the file.

one sample changelog:
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/CHANGELOG.md

PR #41, opened 2022-08-23
Bump main to 2.0.6-dev

susodapop:


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## 2.0.x (Unreleased)
 
+## 2.0.5 (2022-08-23)
+
+- Fix: closing a connection now closes any open cursors from that connection at the server
```
moderakh, line 7: (this comment is outside of the scope of this PR as we have different format for other changelog entries and this is not a blocking comment we should think about it for future re-formatting of the changelog)

We should try to use consistent wording for all entries (Fix vs Fixed, Add, Added) I see we have different wording in the file.

one sample changelog:
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/CHANGELOG.md

moderakh (approved):
LGTM.

@susodapop 
(this comment is outside of the scope of this PR as we have different format for other changelog entries and this is not a blocking comment we should think about it for future re-formatting of the changelog)

We should try to use consistent wording for all entries (Fix vs Fixed, Add, Added) I see we have different wording in the file.

one sample changelog:
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/cosmos/azure-cosmos/CHANGELOG.md

susodapop:
Closing this PR since we've now bumped to 2.1

PR #46, opened 2022-08-26
Test parameter escaping

susodapop:
## Background

`databricks-sql-connector` adheres to PEP-249 which defines a parameter binding API. For supported databases, parameter binding avoids SQL-injection vulnerabilities because parameter values are always escaped at runtime. The hive thrift server doesn't support safe parameter binding (yet!). This means that all queries are passed as strings and are therefore vulnerable to SQL-injection. It's the client's responsibility to safely handle parameterised input before calling `cursor.execute(<some query text>)`.

`databricks-sql-connector` automatically sanitises inputs that follow the PEP-249 calling convention. This pull request adds unit tests to demonstrate the sanitisation behaviour. It also sets a baseline we can use for testing server-side parameter binding once it's supported by the Databricks runtime.

In a follow-up PR, I will add examples that use the PEP-249 parameter binding interface. Once Databricks runtime supports proper server-side binding, users of `databricks-sql-connector` will benefit from its enhanced safety "for free", without needing to rewrite their client code.

### DIFF
--- a/tests/unit/test_param_escaper.py
+++ b/tests/unit/test_param_escaper.py
@@ -24,11 +24,13 @@ class TestIndividualFormatters(object):
 
         assert pe.escape_string("golly bob howdy") == "'golly bob howdy'"
 
-    def test_escape_string_that_includes_quotes(self):
-        # Databricks queries support just one special character: a single quote mark
-        # These are escaped by doubling: 
+    def test_escape_string_that_includes_single_quotes(self):
         # e.g. INPUT: his name was 'robert palmer'
         # e.g. OUTPUT: 'his name was ''robert palmer'''
+
+        assert pe.escape_string("his name was 'robert palmer'") == "'his name was ''robert palmer'''"
+
+    def test_escape_string_that_includes_special_characters(self):
 
         assert pe.escape_string("his name was 'robert palmer'") == "'his name was ''robert palmer'''"
 
--- a/tests/unit/test_param_escaper.py
+++ b/tests/unit/test_param_escaper.py
@@ -1,0 +1,108 @@
+from datetime import date, datetime
+import unittest, pytest
+
+from databricks.sql.utils import ParamEscaper, inject_parameters
+
+pe = ParamEscaper()
+
+class TestIndividualFormatters(object):
+
+    # Test individual type escapers
+    def test_escape_number_integer(self):
+        """This behaviour falls back to Python's default string formatting of numbers
+        """
+        assert pe.escape_number(100) == 100
+
+    def test_escape_number_float(self):
+        """This behaviour falls back to Python's default string formatting of numbers
+        """
+        assert pe.escape_number(100.1234) == 100.1234
+
+    def test_escape_string_normal(self):
+        """
+        """
+
+        assert pe.escape_string("golly bob howdy") == "'golly bob howdy'"
+
+    def test_escape_string_that_includes_quotes(self):
+        # Databricks queries support just one special character: a single quote mark
+        # These are escaped by doubling: 
+        # e.g. INPUT: his name was 'robert palmer'
+        # e.g. OUTPUT: 'his name was ''robert palmer'''
+
+        assert pe.escape_string("his name was 'robert palmer'") == "'his name was ''robert palmer'''"
+
+    def test_escape_date_time(self):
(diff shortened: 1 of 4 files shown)


tests/unit/test_param_escaper.py
```diff
@@ -0,0 +1,108 @@
...
+
+        assert pe.escape_string("golly bob howdy") == "'golly bob howdy'"
+
+    def test_escape_string_that_includes_quotes(self):
+        # Databricks queries support just one special character: a single quote mark
+        # These are escaped by doubling: 
+        # e.g. INPUT: his name was 'robert palmer'
+        # e.g. OUTPUT: 'his name was ''robert palmer'''
+
+        assert pe.escape_string("his name was 'robert palmer'") == "'his name was ''robert palmer'''"
```
moderakh, line 33: could you please add a test for string which contains other special characters as well: e.g, '\\', '/' , " etc

tests/unit/test_param_escaper.py
```diff
@@ -0,0 +1,108 @@
+from datetime import date, datetime
+import unittest, pytest
+
+from databricks.sql.utils import ParamEscaper, inject_parameters
+
+pe = ParamEscaper()
+
+class TestIndividualFormatters(object):
```
moderakh, line 8: do we need to add more tests for other supported types from here?
https://docs.databricks.com/sql/language-manual/sql-ref-datatypes.html

binary? struct? map?

moderakh (approved):
Thanks @susodapop LGTM.

I added a few non-blocking comments for adding more tests.

tests/unit/test_param_escaper.py
kravets-levko, line 33: Is this enough for escaping? I mean - will it work fine if string contains, say, null character, emoji, slash/backslash, etc.?

tests/unit/test_param_escaper.py
```diff
@@ -0,0 +1,108 @@
...
+        assert pe.escape_number(100.1234) == 100.1234
+
+    def test_escape_string_normal(self):
+        """
+        """
+
+        assert pe.escape_string("golly bob howdy") == "'golly bob howdy'"
+
+    def test_escape_string_that_includes_quotes(self):
+        # Databricks queries support just one special character: a single quote mark
```
mcannamela, line 28: i don't understand this comment, per the [spark docs](https://spark.apache.org/docs/latest/sql-ref-literals.html) backslash is used for escaping (so it too is a special character)

you might also try this query in Databricks SQL:  
```
select 'cat\'s meow' 
union all
select 'cat\\\'s meow'
union all
select '\U0001F44D \u3042'
```

see also https://github.com/databricks/databricks-sql-python/issues/51

susodapop: Good callout. Removing this comment in dd25478c4e94fddb33884f6d206fd62427e5076c

tests/unit/test_param_escaper.py
susodapop: I've now added tests for these cases.

tests/unit/test_param_escaper.py
susodapop: I don't think so, because this has to do with mapping user input into a SQL text query. Members of structs would be covered by a different type (string or number) (ditto for maps) and binary is represented as a string (already covered).

tests/unit/test_param_escaper.py
```diff
@@ -0,0 +1,108 @@
...
+        # These are escaped by doubling: 
+        # e.g. INPUT: his name was 'robert palmer'
+        # e.g. OUTPUT: 'his name was ''robert palmer'''
+
+        assert pe.escape_string("his name was 'robert palmer'") == "'his name was ''robert palmer'''"
+
+    def test_escape_date_time(self):
+        INPUT = datetime(1991,8,3,21,55)
+        OUTPUT = "1991-08-03 21:55:00"
```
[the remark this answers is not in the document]
susodapop: Let's push this work to a later PR.  Right now, the parameter escaper doesn't handle timezones (it silently drops them). Since the goal of this PR is to document the existing functionality, the existing test is sufficient. And after the Databricks thrift server supports server-side parameter binding it won't be relevant. 

If users run into issues around this we can consider improving the binding behaviour. But it's low priority for now.

PR #47, opened 2022-08-26
Github workflows: run checks on pull requests from forks

susodapop:
## Description

Prior to this change, the unit test workflow only runs when changes were `push`'d to a branch of the main repository. What we want instead is to run the unit tests for any pull request, including from forks.


moderakh (approved):
thanks @susodapop for quick fix.

PR #48, opened 2022-09-06
Automate deploys to Pypi

susodapop:
Borrowing the github actions from databricks/databricks-sql-cli.

This introduces two GitHub actions:
- test publish will push a new experimental version to test.pypi.org for every pull request. This verifies that a successful publish is possible as of the tip of that branch.
- publish will new push a new version to pypi.org whenever we cut a new GitHub release (using the /releases sub-path of this repository). The version number will be the highest tagged version.

### DIFF
--- a/.github/workflows/publish-test.yml
+++ b/.github/workflows/publish-test.yml
@@ -1,0 +1,64 @@
+name: Publish to PyPI [Test]
+on: [push]
+jobs:
+  test-pypi:
+    name: Create patch version number and push to test-pypi
+    runs-on: ubuntu-latest
+    steps:
+      #----------------------------------------------
+      #       check-out repo and set-up python
+      #----------------------------------------------
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Set up python
+        id: setup-python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      #----------------------------------------------
+      #  -----  install & configure poetry  -----
+      #----------------------------------------------
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+          installer-parallel: true
+      #----------------------------------------------
+      #       load cached venv if cache exists
+      #----------------------------------------------
+      - name: Load cached venv
+        id: cached-poetry-dependencies
+        uses: actions/cache@v2
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ github.event.repository.name }}-${{ hashFiles('**/poetry.lock') }}
+      #----------------------------------------------
+      # install dependencies if cache does not exist
+      #----------------------------------------------
+      - name: Install dependencies
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction --no-root
+      #----------------------------------------------
+      # Get the current version and increment it (test-pypi requires a unique version number)
+      #----------------------------------------------
+      - name: Get next version
+        uses: reecetech/version-increment@2022.2.4
+        id: version
+        with:
+          scheme: semver
+          increment: patch
+      #----------------------------------------------
+      # Tell poetry to update the version number
+      #----------------------------------------------
+      - name: Update pyproject.toml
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          scheme: semver
+          increment: patch
+      #-----------------------------------------------------------------------------
+      # Tell poetry to use the `current-version` that was found by the previous step
+      #-----------------------------------------------------------------------------
+      - name: Update pyproject.toml
+        run: poetry version ${{ steps.version.outputs.current-version }}
+      #----------------------------------------------
+      # Attempt push to test-pypi
```
moderakh, line 59: which action triggers a publish to test-pypi?
do we publish on every change on main branch? or does someone need to take a manual action to trigger publish to test-pypi?

susodapop: It's based on the `release.published` [attribute](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release). So the `publish.yml` only runs when we cut a release on GitHub. 

The `publish-test.yml` runs on every push to a branch on this repository (`main` or even feature branches).

.github/workflows/publish-test.yml
```diff
@@ -0,0 +1,64 @@
...
+      # install dependencies if cache does not exist
+      #----------------------------------------------
+      - name: Install dependencies
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction --no-root
+      #----------------------------------------------
+      # Get the current version and increment it (test-pypi requires a unique version number)
+      #----------------------------------------------
+      - name: Get next version
+        uses: reecetech/version-increment@2022.2.4
```
moderakh, line 46: does this auto increment the version after the release if so to which version?

how does it know to increment the minor version or the major version?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,64 @@
...
+      # install dependencies if cache does not exist
+      #----------------------------------------------
+      - name: Install dependencies
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction --no-root
+      #------------------------------------------------------------------------------------------------
+      # Here we use version-increment to fetch the latest tagged version (we won't increment it though)
+      #------------------------------------------------------------------------------------------------
+      - name: Get next version
+        uses: reecetech/version-increment@2022.2.4
```
moderakh, line 48: we have this in the other publish yaml file to publish to test pypi. 

How do we handle this scenario:
1. we publish to test pypi, hence publish test action increments the version
2. we want to publish the same binary in step 1 to pypi (do we have to roll back the version manually after step 1. incremented it?)

.github/workflows/publish.yml
moderakh: Do we have a doc how we are doing the release? 

> So the publish.yml only runs when we cut a release on GitHub.
what constitutes cutting a release? what would you do to cut a release?

moderakh (commented):
@susodapop an internal doc would help to describe the proposed process for releasing. That would answer many of the discussion/questions in this PR.

Otherwise we have to reverse engineer the yaml file to understand the flow.

.github/workflows/publish.yml
susodapop: Internally, yes.

.github/workflows/publish-test.yml
susodapop: This step has no side-effects.

Understand this is for test-pypi where versions are ephemeral (periodically deleted). The only requirement for the version we push is that it must be _unique_. This step combines with the following step to generate a guaranteed unique version number for testing purposes.

.github/workflows/publish.yml
susodapop: Pypi does not allow rollbacks. test-pypi is entirely different from Pypi. Pushing a version to one has zero affect on the other.

Note that the output from this step is never used as described in the comments. 

The "one true version" is derived from git tags. We're using the `reecetech/version-increment` action as a simple shortcut to parse that the version tags.

.github/workflows/publish.yml
susodapop: The basic steps to create a release are:
1. tag a commit with a semantic version number
2. Create a github release using that tag

moderakh (approved):

PR #49, opened 2022-09-13
[SC-110400] Enabling compression in Python SQL Connector

mohitsingla-db:
This patch enables support for handling compressed results in the python sql connector.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -153,6 +153,7 @@ class Connection:
         self.host = server_hostname
         self.port = kwargs.get("_port", 443)
         self.disable_pandas = kwargs.get("_disable_pandas", False)
+        self.lz4_compression = kwargs.get("enable_query_result_lz4_compression", True)
 
         auth_provider = get_python_sql_connector_auth_provider(
             server_hostname, **kwargs
@@ -208,7 +209,6 @@ class Connection:
         self,
         arraysize: int = DEFAULT_ARRAY_SIZE,
         buffer_size_bytes: int = DEFAULT_RESULT_BUFFER_SIZE_BYTES,
-        lz4_compression: bool = True,
     ) -> "Cursor":
         """
         Return a new Cursor object using the connection.
@@ -223,7 +223,6 @@ class Connection:
             self.thrift_backend,
             arraysize=arraysize,
             result_buffer_size_bytes=buffer_size_bytes,
-            lz4_compression=lz4_compression,
         )
         self._cursors.append(cursor)
         return cursor
@@ -254,7 +253,6 @@ class Cursor:
         thrift_backend: ThriftBackend,
         result_buffer_size_bytes: int = DEFAULT_RESULT_BUFFER_SIZE_BYTES,
         arraysize: int = DEFAULT_ARRAY_SIZE,
-        lz4_compression: bool = True,
     ) -> None:
         """
         These objects represent a database cursor, which is used to manage the context of a fetch
@@ -268,7 +266,6 @@ class Cursor:
         self.buffer_size_bytes = result_buffer_size_bytes
         self.active_result_set: Union[ResultSet, None] = None
         self.arraysize = arraysize
-        self.lz4_compression: bool = lz4_compression
         # Note that Cursor closed => active result set closed, but not vice versa
         self.open = True
         self.executing_command_id = None
@@ -323,7 +320,7 @@ class Cursor:
             session_handle=self.connection._session_handle,
             max_rows=self.arraysize,
             max_bytes=self.buffer_size_bytes,
-            lz4_compression=self.lz4_compression,
+            lz4_compression=self.connection.lz4_compression,
             cursor=self,
         )
         self.active_result_set = ResultSet(
@@ -595,9 +592,6 @@ class Cursor:
     def setoutputsize(self, size, column=None):
         """Does nothing by default"""
         pass
-
-    def setLZ4Compression(self, lz4_compression):
-        self.lz4_compression = lz4_compression
 
 
 class ResultSet:
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -1,4 +1,5 @@
 from typing import Dict, Tuple, List, Optional, Any, Union
+from xmlrpc.client import boolean
 
 import pandas
 import pyarrow
@@ -251,6 +252,7 @@ class Cursor:
         thrift_backend: ThriftBackend,
         result_buffer_size_bytes: int = DEFAULT_RESULT_BUFFER_SIZE_BYTES,
         arraysize: int = DEFAULT_ARRAY_SIZE,
+        lz4_compression: bool = True,
     ) -> None:
         """
         These objects represent a database cursor, which is used to manage the context of a fetch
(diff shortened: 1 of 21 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -940,8 +954,11 @@ def fetch_results(
                     expected_row_start_offset, resp.results.startRowOffset
                 )
             )
+        are_arrow_results_compressed = (
+            resp.resultSetMetadata and resp.resultSetMetadata.lz4Compressed
```
CJStuart, line 958: I don't think `resultSetMetadata` exists here. We'll have to capture it from the ExecuteStatement or GetOperationMetadata response and store it as cursor state.

mohitsingla-db: I think it exists as a field in response. Below is the class in ttypes.

class TFetchResultsResp(object):
    """
    Attributes:
     - status
     - hasMoreRows
     - results
     - resultSetMetadata
    """

I am not sure if you meant that this field won't be set in the response.

sander-goos: This shortcut was added in Thrift V5. However, I don't think it is needed here as the `TResultSetMetadataResp` is already fetched at this point in `_results_message_to_execute_response`. We can capture the `lz4Compressed` field there and store it in `ResultSet` so it can be used here.

We had a discussion before to utilise this shortcut to save an extra roundtrip, but it would require more refactoring for not much value as it only is relevant for non-direct results.

mohitsingla-db: Done.

moderakh (commented):
thanks @mohitsingla-db for the contribution.

Could you please add a little bit more detail to the PR description on the end goal.

Couple of questions for you.
1. what are the pros and cons of compression?
1.1 what's the impact of compression on end to end latency
1.2 what's the impact of the compression on the client side CPU utilization?
3. is compression only applicable to the arrow batch result? what if result is returned row by row?
4. is there a config either on the service side or the sdk side to disable/enable this feature?
5. any impact on the end user? or change on the API? or is this merely implementation detail?

src/databricks/sql/client.py
```diff
@@ -614,6 +614,7 @@ def __init__(
         self.has_been_closed_server_side = execute_response.has_been_closed_server_side
         self.has_more_rows = execute_response.has_more_rows
         self.buffer_size_bytes = result_buffer_size_bytes
+        self.lz4_compressed = execute_response.lz4_compressed
```
bogdanghit, line 617: Can we expose a conf that will allow users to disable compression?

tests/e2e/common/large_queries_mixin.py
```diff
@@ -80,6 +80,7 @@ def test_long_running_query(self):
         scale0 = 10000
         scale_factor = 1
         with self.cursor() as cursor:
+            cursor.execute("SET use_cached_result=false")
```
bogdanghit, line 83: We should add tests with and without caching versus compression on and off.

bogdanghit (approved):
LGTM, but let's wait for @moderakh to approve.

src/databricks/sql/client.py
```diff
@@ -251,6 +254,7 @@ def __init__(
         thrift_backend: ThriftBackend,
         result_buffer_size_bytes: int = DEFAULT_RESULT_BUFFER_SIZE_BYTES,
         arraysize: int = DEFAULT_ARRAY_SIZE,
+        lz4_compression: bool = True,
```
mohitsingla-db, line 257: @moderakh

src/databricks/sql/client.py
```diff
@@ -590,6 +596,9 @@ def setoutputsize(self, size, column=None):
         """Does nothing by default"""
         pass
 
+    def setLZ4Compression(self, lz4_compression):
```
mohitsingla-db, line 599: @moderakh. This will allow to change compression without initialising new cursor or connection. Do we want to give such flexibility?

src/databricks/sql/client.py
moderakh: @mohitsingla-db similar to the JDBC/ODBC driver this should be passed to the client upfront, rather than per cursor or per query

```
sql.connect(host, httpath, your_new_option)
```

`your_new_option` should be passed as kwargs

src/databricks/sql/client.py
moderakh: @mohitsingla-db similar to the JDBC/ODBC driver this should be passed to the client upfront, rather than per cursor or per query

something like this, at this layer.

```python
sql.connect(host, httpath, your_new_option)
```

`your_new_option` should be passed as kwargs
and for its naming please check what the equivalent option is named in jdbc/odbc and here name it consistently

moderakh (changes requested):
@mohitsingla-db similar to the JDBC/ODBC driver the option should be passed to the client upfront, rather than per cursor or per query

something like this, at this layer.
```python
sql.connect(host, httpath, your_new_option)
```
your_new_option should be passed as kwargs
and for its naming please check what the equivalent option is named in jdbc/odbc and here name it consistently

bogdanghit:
> @mohitsingla-db similar to the JDBC/ODBC driver the option should be passed to the client upfront, rather than per cursor or per query
> 
> something like this, at this layer.
> 
> ```python
> sql.connect(host, httpath, your_new_option)
> ```
> 
> your_new_option should be passed as kwargs and for its naming please check what the equivalent option is named in jdbc/odbc and here name it consistently

@moderakh is this conf really necessary or can we drop it?

moderakh:
> > @mohitsingla-db similar to the JDBC/ODBC driver the option should be passed to the client upfront, rather than per cursor or per query
> > something like this, at this layer.
> > ```python
> > sql.connect(host, httpath, your_new_option)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > your_new_option should be passed as kwargs and for its naming please check what the equivalent option is named in jdbc/odbc and here name it consistently
> 
> @moderakh is this conf really necessary or can we drop it?

@bogdanghit I thought it would be good to have parity with the JDBC/ODBC driver. do you have any concern for exposing this config in the pysql?

moderakh (approved):
LGTM. thanks for your contribution @mohitsingla-db Please get sign off from @susodapop

tests/unit/test_thrift_backend.py
```diff
@@ -358,7 +381,7 @@ def test_get_status_uses_display_message_if_available(self, tcli_service_class):
 
         thrift_backend = ThriftBackend("foobar", 443, "path", [], auth_provider=AuthProvider())
         with self.assertRaises(DatabaseError) as cm:
-            thrift_backend.execute_command(Mock(), Mock(), 100, 100, Mock())
+            thrift_backend.execute_command(Mock(), Mock(), 100, 100, Mock(), Mock())
```
susodapop, line 384: Not unique to this PR: but wow is this test inscrutable 😂

susodapop (approved):
LGTM.

Approved with one non-critical change request / nit.

I will run the e2e tests on this branch locally and comment if there are any e2e failures.

susodapop (approved):
🚀

susodapop:
I ran the e2e tests locally and all pass.

PR #52, opened 2022-09-28
[PECO-205] Add functional examples

susodapop:
This pull request adds functional example scripts to demonstrate usage of `databricks-sql-connector`. Each example is a self-contained Python script. Most of this is not _new_ work. It is either copied from the public Databricks documentation or adapted from existing unit or integration test.


examples/persistent_oauth.py
```diff
@@ -0,0 +1,69 @@
...
+
+  1. https://docs.databricks.com/administration-guide/account-settings-e2/single-sign-on.html 
+  2. https://docs.databricks.com/dev-tools/python-sql-connector.html
+"""
+
+import os
+from typing import Optional
+
+from databricks import sql
+from databricks.sql.experimental.oauth_persistence import OAuthPersistence, OAuthToken, DevOnlyFilePersistence
```
moderakh, line 33: could you please also mention that the API (interface) for OAuthPersistence are experimental and evolve over time. I think the package name `databricks.sql.experimental.oauth_persistence` indicates that but it would be great if we could also explicitly say so.

moderakh (approved):
I love these samples. thanks @susodapop

examples/persistent_oauth.py
susodapop: Updated in [f8a81e8](https://github.com/databricks/databricks-sql-python/pull/52/commits/f8a81e8605448c48f65e769da79f8086d4d3d82b)

Thanks!

PR #56, opened 2022-10-06
Escape single quotes with 2 backslashes

### LINKED ISSUE #51: single quote not properly escaped
## What
`ParamEscaper`'s `escape_string()` gives incorrect behavior on Databricks SQL and in Databricks notebooks. 

It [replaces  a single quote `'` with `''`](https://github.com/databricks/databricks-sql-python/blob/2a638c417edbf4e5187ad1db93f50e552a9474b6/src/databricks/sql/utils.py#L149), but the correct way to escape `'` is[ with a backslash](https://spark.apache.org/docs/latest/sql-ref-literals.html), like `\'`. 

You can verify in PySpark with: 
```
assert spark.sql("select 'cat''s meow' as my_col").head(1)[0]['my_col'] == "cats meow"
assert spark.sql("select 'cat\\'s meow' as my_col").head(1)[0]['my_col'] == "cat's meow"
```
Note that because it starts as a Python literal, we need two backslashes `\\` to get Python to first escape `\\'` to `\'` and then Spark escapes to `'`. 

I don't know what the motivation for this implementation was, but the result seems to be concatenation instead of escaping the quote character.

### Reproduction in databricks-sql-python 
The following demonstrates the issue in version ~1.2.2~ 2.0.5 of databricks-sql-python against a serverless SQL warehouse in Azure, v2022.30, plus an implementation without parameter substitution showing an escape treatment that does work :  

```python
from typing import List

from databricks import sql
import os

from databricks.sql import ServerOperationError

server_hostname= os.environ.get('DBT_DATABRICKS_HOST')
http_path=f'/sql/1.0/endpoints/{os.environ.get("DBT_DATABRICKS_ENDPOINT")}'
access_token=os.environ.get('DBT_DATABRICKS_TOKEN')
user_agent_entry = "dbt-databricks/1.2.2"
connection = sql.connect(
  server_hostname=server_hostname,
  http_path=http_path,
  access_token=access_token,
  _user_agent_entry=user_agent_entry
)


cursor = connection.cursor()


def get_result_using_parameter_bindings(p:List[str]):
  try:
    cursor.execute('select %s as my_col', p)
    result = list(cursor.fetchall())[0]['my_col']
  except ServerOperationError as exc:
    result = exc.message.strip()[:20] + '...'
  return result

def get_result_using_fstring(p:List[str]):
  try:
    escaped = p[0].replace('\\','\\\\' ).replace("'", "\\'")
    cursor.execute(f"select '{escaped}' as my_col")
    result = list(cursor.fetchall())[0]['my_col']
  except ServerOperationError as exc:
    result = exc.message.strip()[:20] + '...'
  return result


params = [
  ["cat's meow"],
  ["cat\'s meow"],
  ["cat\\'s meow"],
  ["cat''s meow"],
]



for p in params:
  # using dbt-databricks-sql's parameter substitution
  param_binding_result = get_result_using_parameter_bindings(p)

  # using manually built and escaped query
  f_string_result = get_result_using_fstring(p)



  print('\nparameter value:', p[0], 'parameter-binding result:', param_binding_result, 'round-trip ok?', p[0]==param_binding_result)
  print('parameter value:', p[0], 'f-string result:', f_string_result, 'round-trip ok?',
        p[0] == f_string_result
        )
  assert p[0] == f_string_result


cursor.close()
connection.close()

```

The output is: 
```
bash_1  | parameter value: cat's meow parameter-binding result: cats meow round-trip ok? False
bash_1  | parameter value: cat's meow f-string result: cat's meow round-trip ok? True
bash_1  | 
bash_1  | parameter value: cat's meow parameter-binding result: cats meow round-trip ok? False
bash_1  | parameter value: cat's meow f-string result: cat's meow round-trip ok? True
bash_1  | 
bash_1  | parameter value: cat\'s meow parameter-binding result: [PARSE_SYNTAX_ERROR]... round-trip ok? False
bash_1  | parameter value: cat\'s meow f-string result: cat\'s meow round-trip ok? True
bash_1  | 
bash_1  | parameter value: cat''s meow parameter-binding result: cats meow round-trip ok? False
bash_1  | parameter value: cat''s meow f-string result: cat''s meow round-trip ok? True
```

## Expected results
String parameters with single quotes and backslashes should be properly reproduced:
`"cat's meow"` would be escaped as  `"cat\\'s meow"` and the resulting SQL would return `cat's meow`
`"cat\\'s meow"` would escape to `"cat\\\\\\'s meow"` and the SQL would return `cat\'s meow`

## Suggested fix
I'm not sure how this is usually implemented, but in my example just doing `param.replace('\\','\\\\' ).replace("'", "\\'")` at least preserves single quotes and backslashes, which are probably the most common cases.  It would also leave alone escaped unicode literals like `\U0001F44D`. 

## How I encountered it
I'm using dbt with Databricks and noticed on upgrading from dbt-databricks 1.0 to 1.2.2 that single quotes started disappearing from our "seeds" (csv files loaded as Delta tables). Code had changed in dbt-databricks to use the new parameter binding functionality in this library, whereas (I assume) before it must have been injecting the values as literals into the SQL.

courtneyholcomb:
Fixes #51 by escaping single quotes with backslashes.

### DIFF
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -144,9 +144,9 @@ class ParamEscaper:
         if isinstance(item, bytes):
             item = item.decode("utf-8")
         # This is good enough when backslashes are literal, newlines are just followed, and the way
-        # to escape a single quote is to put two single quotes.
+        # to escape a single quote is to put precede it with 2 backslashes.
         # (i.e. only special character is single quote)
-        return "'{}'".format(item.replace("'", "''"))
+        return "'{}'".format(item.replace("'", "\\'"))
 
     def escape_sequence(self, item):
         l = map(str, map(self.escape_item, item))
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -144,7 +144,7 @@ class ParamEscaper:
         if isinstance(item, bytes):
             item = item.decode("utf-8")
         # This is good enough when backslashes are literal, newlines are just followed, and the way
-        # to escape a single quote is to put precede it with 2 backslashes.
+        # to escape a single quote is to precede it with 2 backslashes.
         # (i.e. only special character is single quote)
         return "'{}'".format(item.replace("'", "\\'"))
 
--- a/tests/e2e/driver_tests.py
+++ b/tests/e2e/driver_tests.py
@@ -288,6 +288,20 @@ class PySQLCoreTestSuite(SmokeTestMixin, CoreTestMixin, DecimalTestsMixin, Times
                 for table in table_names:
                     cursor.execute('DROP TABLE IF EXISTS {}'.format(table))
 
+    def test_escape_single_quotes(self):
+        with self.cursor({}) as cursor:
+            table_name = 'table_{uuid}'.format(uuid=str(uuid4()).replace('-', '_'))
+            # Test escape syntax directly
+            cursor.execute("CREATE TABLE IF NOT EXISTS {} AS (SELECT 'you\\'re' AS col_1)".format(table_name))
+            cursor.execute("SELECT * FROM {} WHERE col_1 LIKE 'you\\'re'".format(table_name))
+            rows = cursor.fetchall()
+            assert rows[0]["col_1"] == "you're"
+
+            # Test escape syntax in parameter
+            cursor.execute("SELECT * FROM {} WHERE {}.col_1 LIKE %(var)s".format(table_name, table_name), parameters={"var": "you're"})
+            rows = cursor.fetchall()
+            assert rows[0]["col_1"] == "you're"
+
     def test_get_schemas(self):
         with self.cursor({}) as cursor:
             database_name = 'db_{uuid}'.format(uuid=str(uuid4()).replace('-', '_'))


susodapop:
Thanks for this contribution! We're actually bundling a few changes around this behaviour into a larger pull request here shortly. I think per the spec defined in the comments on #51, the escape sequence should be a single back stroke `\` rather than doubled. But it's confounded because there's actually two escape characters: the Python escape character and the SQL escape character XD. No matter what, we'll incorporate a proper usage example in the `/examples` directory as well.

I'm going to keep this PR open while we finish up our internal version. This way we can incorporate your e2e test onto our branch and make sure it still passes (we've been using unit tests defined in #46. This way you'll also get contributor credit on the repo :)

src/databricks/sql/utils.py
```diff
@@ -144,9 +144,9 @@ def escape_string(self, item):
         if isinstance(item, bytes):
             item = item.decode("utf-8")
         # This is good enough when backslashes are literal, newlines are just followed, and the way
-        # to escape a single quote is to put two single quotes.
+        # to escape a single quote is to precede it with 2 backslashes.
         # (i.e. only special character is single quote)
-        return "'{}'".format(item.replace("'", "''"))
+        return "'{}'".format(item.replace("'", "\\'"))
```
mcannamela, line 149: i know that per @susodapop this will be fixed elsewhere, but want to point out that the backslash character itself also needs to be escaped

i think the correct replacement would be: `item.replace('\\','\\\\' ).replace("'", "\\'")`

tests/e2e/driver_tests.py
```diff
@@ -288,6 +288,20 @@ def test_get_columns(self):
                 for table in table_names:
                     cursor.execute('DROP TABLE IF EXISTS {}'.format(table))
 
+    def test_escape_single_quotes(self):
```
mcannamela, line 291: this is great, probably want the same type of test for a string with a backslash in it, as in 

```python
test_str = "Many Linux users are upset that Windows uses \\ to separate path elements"
```

tests/e2e/driver_tests.py
```diff
@@ -288,6 +288,20 @@ def test_get_columns(self):
...
+        with self.cursor({}) as cursor:
+            table_name = 'table_{uuid}'.format(uuid=str(uuid4()).replace('-', '_'))
+            # Test escape syntax directly
+            cursor.execute("CREATE TABLE IF NOT EXISTS {} AS (SELECT 'you\\'re' AS col_1)".format(table_name))
```
susodapop, line 295: Just a note: this e2e test is really testing how DBSQL behaves rather than our connector. The connector escape behaviour is only called to escape parameters passed to the Connection.execute() method.

This is a good test for demonstrating how to work with DBSQL. But the dynamic is not unique to `databricks-sql-connector` as we observe the same with NodeJS, Go, and ODBC/JDBC.

susodapop:
I've incorporated your e2e test from this PR into https://github.com/databricks/databricks-sql-python/pull/46 which includes your change plus a bunch of additional test cases. Closing this PR in favour of that one. Thank you for this contribution!

PR #67, opened 2022-11-15
SI: Implement put operations

susodapop:
## Description

This pull request implements basic `PUT` operations. 

- [x] PUT is supported with tests
- [x] GET is supported with tests
- [x] REMOVE is supported with tests

Forthcoming: Based on our internal design doc we need to restrict which local files may be PUT using `uploadsBasePath`. This will come in a follow-up PR.

The scope of PECO-397 was to implement PUT only. But I think we should include implementing GET in this PR so we can verify in the e2e test that the file we attempted to PUT was successfully captured on the server. Right now we're just checking for a 200 status code. Which is better than nothing but only means that the server said everything is okay. But maybe the server is wrong...

**Update**: I've now implemented GET in this pull request so we can verify that a PUT operation succeeded.

**Update**: I've applied all PR feedback minus adding a usage example. To make the example I need to know a bit more about how we want to address multi-end-user cases (do we need more than one `uploads_base_path`, should we rename `uploads_base_path`, and should we illustrate how to perform retries of the staging operations or simply retry them automatically?)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,6 +112,7 @@ export access_token=""
 There are several e2e test suites available:
 - `PySQLCoreTestSuite`
 - `PySQLLargeQueriesSuite`
+- `PySQLStagingIngestionTestSuite`
 - `PySQLRetryTestSuite.HTTP503Suite` **[not documented]**
 - `PySQLRetryTestSuite.HTTP429Suite` **[not documented]**
 - `PySQLUnityCatalogTestSuite` **[not documented]**
@@ -121,6 +122,12 @@ To execute the core test suite:
 ```bash
 poetry run python -m pytest tests/e2e/driver_tests.py::PySQLCoreTestSuite
 ```
+
+The `PySQLCoreTestSuite` namespace contains tests for all of the connector's basic features and behaviours. This is the default namespace where tests should be written unless they require specially configured clusters or take an especially long-time to execute by design.
+
+The `PySQLLargeQueriesSuite` namespace contains long-running query tests and is kept separate. In general, if the `PySQLCoreTestSuite` passes then these tests will as well.
+
+The `PySQLStagingIngestionTestSuite` namespace requires a cluster running DBR version > 13.x which supports staging ingestion commands.
 
 The suites marked `[not documented]` require additional configuration which will be documented at a later time.
 ### Code formatting
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -127,7 +127,7 @@ The `PySQLCoreTestSuite` namespace contains tests for all of the connector's bas
 
 The `PySQLLargeQueriesSuite` namespace contains long-running query tests and is kept separate. In general, if the `PySQLCoreTestSuite` passes then these tests will as well.
 
-The `PySQLStagingIngestionTestSuite` namespace requires a cluster running DBR version > 13.x which supports staging ingestion commands.
+The `PySQLStagingIngestionTestSuite` namespace requires a cluster running DBR version > 12.x which supports staging ingestion commands.
 
 The suites marked `[not documented]` require additional configuration which will be documented at a later time.
 ### Code formatting
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,10 @@ pyarrow = "^9.0.0"
 lz4 = "^4.0.2"
 requests=">2.18.1"
 oauthlib=">=3.1.0"
+numpy = [
+    {version = "1.21.1", python = ">=3.7,<3.8"},
+    {version = "1.23.4", python = ">=3.8"}
+]
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.1.2"
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -300,7 +300,9 @@ class Cursor:
         if not self.open:
             raise Error("Attempting operation on closed cursor")
 
-    def _handle_staging_operation(self, staging_allowed_local_path: Union[None, str, List[str]]):
+    def _handle_staging_operation(
+        self, staging_allowed_local_path: Union[None, str, List[str]]
+    ):
         """Fetch the HTTP request instruction from a staging ingestion command
         and call the designated handler.
 
@@ -317,7 +319,9 @@ class Cursor:
                 "You must provide at least one staging_allowed_local_path when initialising a connection to perform ingestion commands"
             )
 
-        abs_staging_allowed_local_paths = [os.path.abspath(i) for i in _staging_allowed_local_paths]
+        abs_staging_allowed_local_paths = [
+            os.path.abspath(i) for i in _staging_allowed_local_paths
+        ]
 
         assert self.active_result_set is not None
         row = self.active_result_set.fetchone()
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -30,7 +30,7 @@ class Connection:
         session_configuration: Dict[str, Any] = None,
         catalog: Optional[str] = None,
         schema: Optional[str] = None,
-        **kwargs
+        **kwargs,
     ) -> None:
         """
         Connect to a Databricks SQL endpoint or a Databricks cluster.
@@ -175,7 +175,7 @@ class Connection:
             http_path,
             (http_headers or []) + base_headers,
             auth_provider,
-            **kwargs
+            **kwargs,
         )
 
         self._session_handle = self.thrift_backend.open_session(
@@ -305,17 +305,18 @@ class Cursor:
 
         row = self.active_result_set.fetchone()
 
-
         # TODO: Experiment with DBR sending real headers.
         # The specification says headers will be in JSON format but the current null value is actually an empty list []
         handler_args = {
             "operation": row.operation,
             "presigned_url": row.presignedUrl,
             "local_file": row.localFile,
-            "headers": json.loads(row.headers or "{}")
+            "headers": json.loads(row.headers or "{}"),
         }
-        
-        logger.debug(f"Attempting staging operation indicated by server: {row.operation} - {row.localFile}")
+
+        logger.debug(
+            f"Attempting staging operation indicated by server: {row.operation} - {row.localFile}"
+        )
 
         # TODO: Create a retry loop here to re-attempt if the request times out or fails
         if row.operation == "GET":
@@ -326,12 +327,15 @@ class Cursor:
             # Local file isn't needed to remove a remote resource
             handler_args.pop("local_file")
             return self._handle_staging_remove(**handler_args)
-        else: 
-            raise Error(f"Operation {row.operation} is not supported. " +
-                    "Supported operations are GET, PUT, and REMOVE")
+        else:
+            raise Error(
+                f"Operation {row.operation} is not supported. "
+                + "Supported operations are GET, PUT, and REMOVE"
+            )
 
-
-    def _handle_staging_put(self, operation:str, presigned_url:str, local_file:str, headers:dict=None):
+    def _handle_staging_put(
+        self, operation: str, presigned_url: str, local_file: str, headers: dict = None
+    ):
         """Make an HTTP PUT request
 
         Raise an exception if request fails. Returns no data.
@@ -354,14 +358,19 @@ class Cursor:
         # fmt: on
 
         if r.status_code not in [OK, CREATED, NO_CONTENT, ACCEPTED]:
-            raise Error(f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}")
-        
-        if r.status_code == ACCEPTED:
-            logger.debug(f"Response code {ACCEPTED} from server indicates ingestion command was accepted " +
-                "but not yet applied on the server. It's possible this command may fail later.")
-        
+            raise Error(
+                f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}"
+            )
 
-    def _handle_staging_get(self, operation:str, local_file:str, presigned_url:str, headers:dict=None):
+        if r.status_code == ACCEPTED:
+            logger.debug(
+                f"Response code {ACCEPTED} from server indicates ingestion command was accepted "
+                + "but not yet applied on the server. It's possible this command may fail later."
+            )
+
+    def _handle_staging_get(
+        self, operation: str, local_file: str, presigned_url: str, headers: dict = None
+    ):
         """Make an HTTP GET request, create a local file with the received data
 
         Raise an exception if request fails. Returns no data.
@@ -373,16 +382,17 @@ class Cursor:
             # response.ok verifies the status code is not between 400-600.
             # Any 2xx or 3xx will evaluate r.ok == True
             if not r.ok:
-                raise Error(f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}")
-                
-            fp.write(r.content)
-        
-        
+                raise Error(
+                    f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}"
+                )
 
-    def _handle_staging_remove(self, operation, presigned_url, local_file, headers=None):
+            fp.write(r.content)
+
+    def _handle_staging_remove(
+        self, operation, presigned_url, local_file, headers=None
+    ):
         pass
 
-    
     def execute(
         self, operation: str, parameters: Optional[Dict[str, str]] = None
     ) -> "Cursor":
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -4,6 +4,7 @@ import pandas
 import pyarrow
 import requests
 import json
+import os
 
 from databricks.sql import __version__
 from databricks.sql import *
@@ -301,14 +302,22 @@ class Cursor:
 
     def _handle_staging_operation(self, uploads_base_path: str):
         """Fetch the HTTP request instruction from a staging ingestion command
-        and call the designated handler."""
+        and call the designated handler.
+        
+        Raise an exception if localFile is specified by the server but the localFile
+        is not descended from uploads_base_path.
+        """
 
         if uploads_base_path is None:
             raise Error(
                 "You must provide an uploads_base_path when initialising a connection to perform ingestion commands"
             )
-
+        
         row = self.active_result_set.fetchone()
+
+        if getattr(row, "localFile", None):
+            if os.path.commonpath([row.localFile, uploads_base_path]) != uploads_base_path:
+                raise Error("Local file operations are restricted to paths within the configured uploads_base_path")
 
         # TODO: Experiment with DBR sending real headers.
         # The specification says headers will be in JSON format but the current null value is actually an empty list []
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -313,6 +313,7 @@ class Cursor:
 
         operation_map = {
             "PUT": requests.put,
+            "GET": requests.get,
         }
 
         if operation not in operation_map:
@@ -344,6 +345,10 @@ class Cursor:
                     resp.status_code, resp.text
                 )
             )
+
+        if operation == "GET":
+            with open(local_file, "wb") as fp:
+                fp.write(resp.content)
 
     def execute(
         self, operation: str, parameters: Optional[Dict[str, str]] = None
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -392,16 +392,16 @@ class Cursor:
         if local_file is None:
             raise Error("Cannot perform GET without specifying a local_file")
 
+        r = requests.get(url=presigned_url, headers=headers)
+
+        # response.ok verifies the status code is not between 400-600.
+        # Any 2xx or 3xx will evaluate r.ok == True
+        if not r.ok:
+            raise Error(
+                f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}"
+            )
+
         with open(local_file, "wb") as fp:
-            r = requests.get(url=presigned_url, headers=headers)
-
-            # response.ok verifies the status code is not between 400-600.
-            # Any 2xx or 3xx will evaluate r.ok == True
-            if not r.ok:
-                raise Error(
-                    f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}"
-                )
-
             fp.write(r.content)
 
     def _handle_staging_remove(self, presigned_url: str, headers: dict = None):
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -312,18 +312,23 @@ class Cursor:
             raise Error(
                 "You must provide an uploads_base_path when initialising a connection to perform ingestion commands"
             )
-        
+
         row = self.active_result_set.fetchone()
 
+        # Must set to None in cases where server response does not include localFile
+        abs_localFile = None
+
         if getattr(row, "localFile", None):
-            if os.path.commonpath([row.localFile, uploads_base_path]) != uploads_base_path:
+            abs_localFile = os.path.abspath(row.localFile)
+            abs_uploads_base_path = os.path.abspath(uploads_base_path)
+            if os.path.commonpath([abs_localFile, abs_uploads_base_path]) != abs_uploads_base_path:
                 raise Error("Local file operations are restricted to paths within the configured uploads_base_path")
 
         # TODO: Experiment with DBR sending real headers.
         # The specification says headers will be in JSON format but the current null value is actually an empty list []
         handler_args = {
             "presigned_url": row.presignedUrl,
-            "local_file": getattr(row, "localFile", None),
+            "local_file": abs_localFile,
             "headers": json.loads(row.headers or "{}"),
         }
 
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -300,24 +300,24 @@ class Cursor:
         if not self.open:
             raise Error("Attempting operation on closed cursor")
 
-    def _handle_staging_operation(self, uploads_base_path: Union[List, str]):
+    def _handle_staging_operation(self, staging_allowed_local_path: Union[List, str]):
         """Fetch the HTTP request instruction from a staging ingestion command
         and call the designated handler.
 
         Raise an exception if localFile is specified by the server but the localFile
-        is not descended from uploads_base_path.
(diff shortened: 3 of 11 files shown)


src/databricks/sql/client.py
```diff
@@ -297,6 +298,53 @@ def _check_not_closed(self):
...
+
+        # TODO: Handle headers. What format will gateway send? json? plaintext?
+        operation, presigned_url, local_file, headers = (
+            row.operation,
+            row.presignedUrl,
+            row.localFile,
+            None,
+        )
+
+        operation_map = {
```
susodapop, line 314: This will be expanded to include REMOVE and GET

src/databricks/sql/client.py
```diff
@@ -297,6 +298,53 @@ def _check_not_closed(self):
...
+        rq_func_args = dict(url=presigned_url, data=raw_data)
+
+        logger.debug(
+            "Attempting staging operation: {} - {}".format(operation, local_file)
+        )
+
+        # Call the function
+        resp = req_func(**rq_func_args)
+
+        if resp.status_code != 200:
```
susodapop, line 341: I'd like to e2e test this behaviour. Just need a way to force requests to throw a 501 or similar error. Probably easiest to do it with a mock.

src/databricks/sql/client.py
```diff
@@ -331,6 +379,10 @@ def execute(
             self.buffer_size_bytes,
             self.arraysize,
         )
+
+        if execute_response.is_staging_operation:
```
susodapop, line 383: Question for reviewers: is there any specifically desired end-state for the `cursor` after a staging operation? Maybe we return a new NamedTuple `StagingOperationResult` with properties of `.successful:boolean` and perhaps a copy of the `operation` and `localFile` that were used?

tests/e2e/driver_tests.py
```diff
@@ -630,6 +632,53 @@ def test_initial_namespace(self):
...
+
+        with open(fh, 'wb') as fp:
+            fp.write("hello world!".encode("utf-8"))
+
+        with self.connection() as conn:
+            cursor = conn.cursor()
+            query = f"PUT '{temp_path}' INTO 'stage://tmp/{self.staging_ingestion_user}/tmp/11/15/file1.csv' OVERWRITE"
+            cursor.execute(query)
+
+        os.remove(temp_path)
```
susodapop, line 658: temp files created with `mkstemp()` must be manually removed per the Python docs

src/databricks/sql/client.py
moderakh: as this is constant I wonder if we should define this outside of the method? or is the python culture do this inside method?

src/databricks/sql/client.py
```diff
@@ -297,6 +298,58 @@ def _check_not_closed(self):
...
+
+        operation_map = {
+            "PUT": requests.put,
+            "GET": requests.get,
+        }
+
+        if operation not in operation_map:
+            raise Error(
+                "Operation {} is not supported. Supported operations are {}".format(
+                    operation, ",".join(operation_map.keys())
```
moderakh, line 322: any reason we are not using f strings?
https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python

sounds a better approach comparing to `.format()` to me.

src/databricks/sql/client.py
```diff
@@ -297,6 +298,58 @@ def _check_not_closed(self):
...
+
+        if local_file:
+            raw_data = open(local_file, "rb")
+        else:
+            raw_data = None
+
+        rq_func_args = dict(url=presigned_url, data=raw_data)
+
+        logger.debug(
+            "Attempting staging operation: {} - {}".format(operation, local_file)
```
moderakh, line 336: ditto, here and for other `.format()`

src/databricks/sql/client.py
```diff
@@ -297,6 +298,58 @@ def _check_not_closed(self):
...
+        )
+
+        # Call the function
+        resp = req_func(**rq_func_args)
+
+        if resp.status_code != 200:
+            raise Error(
+                "Staging operation over HTTP was unsuccessful: {}-{}".format(
```
moderakh, line 344: there are different failure scenarios:
1. non retriable errors (e.g., 401 or 403, file system access error)
2. retryable errors (e.g., socket connection failure)

we should consider retrying retryable failures. how do we retry idempotent query failures? what is our retry policy for query failures?

CONTRIBUTING.md
```diff
@@ -122,6 +123,12 @@ To execute the core test suite:
 poetry run python -m pytest tests/e2e/driver_tests.py::PySQLCoreTestSuite
 ```
 
+The `PySQLCoreTestSuite` namespace contains tests for all of the connector's basic features and behaviours. This is the default namespace where tests should be written unless they require specially configured clusters or take an especially long-time to execute by design.
+
+The `PySQLLargeQueriesSuite` namespace contains long-running query tests and is kept separate. In general, if the `PySQLCoreTestSuite` passes then these tests will as well.
+
+The `PySQLStagingIngestionTestSuite` namespace requires a cluster running DBR version > 13.x which supports staging ingestion commands.
```
xiaonanyang-db, line 130: Spark driver supports staging commands in 12.x as well

src/databricks/sql/client.py
```diff
@@ -297,6 +298,58 @@ def _check_not_closed(self):
...
+        """Make HTTP request using instructions provided by server"""
+
+        row = self.active_result_set.fetchone()
+
+        # TODO: Handle headers. What format will gateway send? json? plaintext?
+        operation, presigned_url, local_file, headers = (
+            row.operation,
+            row.presignedUrl,
+            row.localFile,
+            None,
```
xiaonanyang-db, line 311: Why are headers always None here? Did we parse the headers from the spark driver response?

src/databricks/sql/client.py
```diff
@@ -297,6 +298,58 @@ def _check_not_closed(self):
...
+            )
+
+        req_func = operation_map[operation]
+
+        if local_file:
+            raw_data = open(local_file, "rb")
+        else:
+            raw_data = None
+
+        rq_func_args = dict(url=presigned_url, data=raw_data)
```
xiaonanyang-db, line 333: may need to add headers here

src/databricks/sql/client.py
HSCandYH: headers could be non-empty from Azure Cloud, please make sure we take care of this and considering integration tests that could run across cloud.

src/databricks/sql/client.py
susodapop: Python actually has three interchangeable string formatting syntaxes. 

1. `f"This is a {value}"`
2. `"This is a {}".format(value)`
3. `"This is a %s" % value`

PySQL uses all three of them in different places. I'll switch these to f-strings if that's what you like :)

src/databricks/sql/client.py
```diff
@@ -297,6 +298,58 @@ def _check_not_closed(self):
...
+
+        logger.debug(
+            "Attempting staging operation: {} - {}".format(operation, local_file)
+        )
+
+        # Call the function
+        resp = req_func(**rq_func_args)
+
+        if resp.status_code != 200:
```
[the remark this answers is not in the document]
susodapop: Good catch. Will update

src/databricks/sql/client.py
susodapop: Good questions. I don't think this should block merge here though. We can add it with a follow-up PR subsequent.

PySQL has robust retry behaviour specified for **thrift operations**, including unit and integration tests. But these staging operations are not performed over Thrift. So if we implement retry behaviour here we'll be doing so from scratch and will should probably send it through a one-decision doc + review because each cloud provider may have different needs / response headers etc. 

For now, the http request to the cloud storage provider is tried exactly once. In a follow-up we can implement retry behaviour for the ingestion HTTP Requests (and probably even borrow the config used for thrift retries)

CONTRIBUTING.md
susodapop: Thank you! I'll update this to reflect.

src/databricks/sql/client.py
susodapop: The `None` here is a placeholder. I wrote above this block in the comments that I don't know what format the `headers` field will take. But to send them to `requests` I need to turn them into a Python `dict()`. To do that I need to know what format they will arrive in.

The sample notebook from @xiaonanyang-db doesn't include any values in the headers field that I saw. Does the Spark Driver send them as JSON?

src/databricks/sql/client.py
susodapop: Yep, just need to know what format the headers will arrive in so I can reformat them in the way `requests` expects.

src/databricks/sql/client.py
```diff
@@ -297,6 +298,58 @@ def _check_not_closed(self):
...
+            raise Error(
+                "Operation {} is not supported. Supported operations are {}".format(
+                    operation, ",".join(operation_map.keys())
+                )
+            )
+
+        req_func = operation_map[operation]
+
+        if local_file:
+            raw_data = open(local_file, "rb")
```
HSCandYH, line 329: local_file is super interesting here, customer might try different schemes I bet.
dbfs://bla, https://bla, files://bla

Each of those schemes has a related "open" function and client side shall try to understand their ask and support/decline their request. 

I also would like to know how much of those schemes we plan to support in near term and how this function would grow

src/databricks/sql/client.py
xiaonanyang-db: @susodapop got it. Yes, it would look like a JSON string like `{"h1": "v1"}`

src/databricks/sql/client.py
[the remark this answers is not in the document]
HSCandYH: check the implementation of response.ok [here](https://src.dev.databricks.com/databricks/universe/-/blob/redash/managed_redash/redash/handlers/ingestion.py?L455) might give you some good idea.

src/databricks/sql/client.py
xiaonanyang-db: You can reference this test suite https://src.dev.databricks.com/databricks/runtime@afacc75d9c5436209fbac057d1f09caf9e30d196/-/blob/sql/hive-thriftserver/src/test/scala/com/databricks/sql/hive/thriftserver/StagingOperationSQLQuerySuite.scala?L65

src/databricks/sql/client.py
HSCandYH: please change this line according to [stack-over flow](https://stackoverflow.com/questions/2342579/http-status-code-for-update-and-delete).

I have a bug like that before on Azure cloud, their success is 201.

```
For a PUT request: HTTP 200, HTTP 204 should imply "resource updated successfully". HTTP 201 if the PUT request created a new resource.

For a DELETE request: HTTP 200 or HTTP 204 should imply "resource deleted successfully".

HTTP 202 can also be returned by either operation and would imply that the instruction was accepted by the server, but not fully applied yet. It's possible that the operation fails later, so the client shouldn't fully assume that it was success.

A client that receives a status code it doesn't recognize, but it's starting with 2 should treat it as a 200 OK.
```

src/databricks/sql/client.py
HSCandYH: The error is more precise if we specify it's the pre-signed url over HTTP and we should anticipate this can fail.
However, for retryable errors, I am wondering how much the client lib here should retry.

it's possible that the user of this client lib will retry as well, we should be consistent with other places to handle the transit error.

FYI,
PUT url is createOrUpdate, so it's ok to retry.
Delete url can be issued multiple times and only take effect when file is found.

src/databricks/sql/client.py
HSCandYH: I don't quite get this question, but the cursor for now will return just one row and we should have reached the end of this cursor.

src/databricks/sql/client.py
susodapop: Preliminary spec is to only support upload of local files. Nothing in dbfs or from an arbitrary URL. That restriction isn't implemented here because it's part of a separate ticket. The basic idea is that uploads will only be possible when a user configures an `uploadsBasePath` pointing to a mounted volume.

I agree that we need a way to hook this behaviour for other file origins, however. I'm going to noodle how we can make this sufficiently generic for now.

cc: @moderakh

src/databricks/sql/client.py
HSCandYH: REMOVE is supported, just making sure it will be a separate PR or we will add it here ?

susodapop: Yeah sorry I know it's confusing. The way the tickets for this were first made it seemed like we wanted separate PR's for:

- PUT
- GET
- REMOVE
- Restrict PUT to `baseUploadsPath`

But I think it should really just be one PR which keeps it from being so confusing / wasting reviewer time.

susodapop: In 55525cbf8660c7d47a628a2c072610b45c212ee8 I refactored to stub in the GET and REMOVE behaviours. All these will be included in this PR.

src/databricks/sql/client.py
susodapop: Applied these changes to GET and PUT in 0739ccc9f76e123e40e72eb697c5f868a58ce6b8. Will incorporate into the REMOVE logic shortly.

src/databricks/sql/client.py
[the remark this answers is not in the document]
susodapop: Done here d3a3651b9a8fdfd5f6cf3a897234f7e335a8d5f0

Thanks for the pointer.

CONTRIBUTING.md
susodapop: Fixed here: 72f917ee2a0182e1c2a4a35e2774adf197ab94af

src/databricks/sql/client.py
susodapop: Interestingly, the response from the server today (with no headers) evaluates to an empty Python list `[]`. If it's supposed to be an empty JSON I would expect it to be `{}`. Which is I coerce it to starting in this commit: 55525cbf8660c7d47a628a2c072610b45c212ee8 (see line 315 in client.py)

susodapop:
I've pushed in the `uploadsBasePath` restriction and associated tests. I have a few questions for the group though:

1. Should we allow `uploads_base_path` to be a list of allowed paths?
2. Do we need to require `uploads_base_path` for REMOVE operations?
3. Should we re-name `uploads_base_path` to `allowed_local_dirs` since it's technically restricting not only what data we PUT but what data we GET?
4. Right now we require`uploads_base_path` be provide and be a path-like string. Do we need any other explicit constraints? For example we could prevent users from specifying the root directory `'/'` as their `uploads_base_path`. Is that desireable?

cc @moderakh @HSCandYH

src/databricks/sql/client.py
```diff
@@ -297,6 +300,117 @@ def _check_not_closed(self):
...
+        if uploads_base_path is None:
+            raise Error(
+                "You must provide an uploads_base_path when initialising a connection to perform ingestion commands"
+            )
+        
+        row = self.active_result_set.fetchone()
+
+        if getattr(row, "localFile", None):
+            if os.path.commonpath([row.localFile, uploads_base_path]) != uploads_base_path:
+                raise Error("Local file operations are restricted to paths within the configured uploads_base_path")
```
susodapop, line 320: This may need to be more sophisticated. I'm considering what valid `uploads_base_path` could be provided where the longest common path shared by both files _is not_ identical to `uploads_base_path`. The idea here is that if the safe path is `/path/to/file` and the server specifies `/path/to/file/somewhere/on/disk` then the longest common path is `/path/to/file`.

src/databricks/sql/client.py
```diff
@@ -297,6 +300,117 @@ def _check_not_closed(self):
...
+
+    def _handle_staging_get(
+        self, local_file: str, presigned_url: str, headers: dict = None
+    ):
+        """Make an HTTP GET request, create a local file with the received data
+
+        Raise an exception if request fails. Returns no data.
+        """
+
+        with open(local_file, "wb") as fp:
```
xiaonanyang-db, line 392: nit: would it be better to open an outputStream after we get a successful response?

xiaonanyang-db, line 392: nit: would it be better to open an outputStream after we get a successful response?

xiaonanyang-db, line 392: nit: should we do a similar not-None check on `local_file` as the put handler?

tests/e2e/driver_tests.py
```diff
@@ -630,6 +632,115 @@ def test_initial_namespace(self):
             cursor.execute("select current_database()")
             self.assertEqual(cursor.fetchone()[0], table_name)
 
+class PySQLStagingIngestionTestSuite(PySQLTestCase):
```
xiaonanyang-db, line 635: Can you add more tests?
- Two Puts to the same staging path should fail the second one
- Put/Get/Delete files in another user's staging location should fail with permission-denied errors
- Empty `localPath` argument in the query should fail with parsing errors on server side
- Invalid `stagingPath` argument in the query should fail with parsing errors on server side
  - Valid staging path should be a valid URI starting with `stage://tmp/`

cc @HSCandYH

xiaonanyang-db (approved):
LGTM except some minor comments and tests request

src/databricks/sql/client.py
HSCandYH: I think we should do that from a usability perspective before merge. the use case here is

User 1:
```
PUT /path/to/file/$userID1/2022-12-05 into $PSL_USER1

COPY INTO $deltaTable FROM $PSL_USER1
```

User 2:
```
PUT /path/to/file/$userID2/2022-12-05 into $PSL_USER2

COPY INTO $deltaTable2 FROM $PSL_USER2
```

We would like 2 user to use uploads_base_path and provide some sort of isolation among them.

WDYT ?

HSCandYH: Also, we should add some tests like this, we need to make sure the resolved path still start with uploads_base_path
```
PUT $uploads_base_path/../a.file into $PSL_USER2
```

pyproject.toml
```diff
@@ -16,6 +16,10 @@ pyarrow = "^9.0.0"
 lz4 = "^4.0.2"
 requests=">2.18.1"
 oauthlib=">=3.1.0"
+numpy = [
+    {version = "1.21.1", python = ">=3.7,<3.8"},
+    {version = "1.23.4", python = ">=3.8"}
+]
```
moderakh, line 22: why is numpy changes are relevant to staging?

src/databricks/sql/client.py
```diff
@@ -297,6 +300,117 @@ def _check_not_closed(self):
...
+        Raise an exception if localFile is specified by the server but the localFile
+        is not descended from uploads_base_path.
+        """
+
+        if uploads_base_path is None:
+            raise Error(
+                "You must provide an uploads_base_path when initialising a connection to perform ingestion commands"
+            )
+        
+        row = self.active_result_set.fetchone()
```
moderakh, line 316: I know `self.active_result_set` is introduced in this PR.
so this is merely a generic question rather that specific to staging.

if we are using a field member `self.active_result_set` for keeping a state that means we won't be able to support multi threading in an application which concurrently uses pysql. is this understanding correct?

src/databricks/sql/client.py
```diff
@@ -297,6 +300,117 @@ def _check_not_closed(self):
...
+
+        if getattr(row, "localFile", None):
+            if os.path.commonpath([row.localFile, uploads_base_path]) != uploads_base_path:
```
moderakh, line 319: what happens if `uploads_base_path = /Users/user1` and `row.localFile = /Users/user1/../user2` what does this method return?

moderakh: could you please add some tests for that?

src/databricks/sql/client.py
moderakh: @susodapop could you please with an a sample code explain how this will provide different experience to the end user?

src/databricks/sql/thrift_backend.py
```diff
@@ -733,6 +735,8 @@ def _results_message_to_execute_response(self, resp, operation_state):
             .to_pybytes()
         )
         lz4_compressed = t_result_set_metadata_resp.lz4Compressed
+        # TODO: will this fail if metadata doesn't include `isStagingOperation`?
```
moderakh, line 738: is the TODO still relevant?

tests/e2e/driver_tests.py
moderakh: +1

moderakh (approved):
Thanks @susodapop high level looks good.

could you please verify what happens if `uploads_base_path = /Users/user1` and `row.localFile = /Users/user1/../user2` maybe add a test.

also please add a sample code for the cursor usage after staging operation.

also could you please once a comment is addressed, resolve the comment. not clear to me which comments are addressed and which one not addressed in the PR.

pyproject.toml
susodapop: This one is a bit tricky to explain. tl;dr in this PR we're making explicit a Python-version-specific numpy dependency that was _implicit_ when we added Python 3.,10 support in #31.

Implicit here means that we didn't specify the numpy versions in `pyproject.toml` but Poetry automatically figured it out during dependency resolution and captured that in `poetry.lock`. Since then, poetry dependency resolution changed. In the process of building this PR, Poetry found itself in a dependency resolution recursive loop. To resolve the issue I explicitly pinned the `numpy` version so that `poetry install` would run properly on Python 3.10.

We _could have done this during #31_ but it wasn't necessary because Poetry worked at that time and all our unit and e2e tests passed.

AFAIK this change doesn't actually affect how the package is installed. It's here to help Poetry know how to resolve its dependencies.

susodapop: Just adding a note here to say that this is an _excellent_ question to ask for this PR. If you peek at the `main` branch you'll see we've been adding more explicit Python-version-specific dependency declarations because of `pyarrow` too.

src/databricks/sql/client.py
susodapop: I'm confused by the first part of your question:

> I know self.active_result_set is introduced in this PR.

I don't believe this is correct. `active_result_set` has been present since the first version of this library. It's present on `main` right now.

> if we are using a field member self.active_result_set for keeping a state that means we won't be able to support multi threading in an application which concurrently uses pysql

You're pulling on a valid thread. But I disagree with this assessment. In general pysql works fine with multi-threading. In fact, multi-threading is _required_ if you want to cancel a running query (which is reflected in `PySQLCoreTestSuite.test_cancel_during_execute`).

The specific scenario where `active_result_set` state would affect multi-threaded applications is if multiple threads are working _with the same cursor_. Is that a desirable usage pattern? I think there is usually one cursor per thread, in which case there's no issue with shared state.

src/databricks/sql/client.py
susodapop: > We would like 2 user to use uploads_base_path and provide some sort of isolation among them.
> 
> WDYT ?

I don't think we have a clean way to limit user2 from accessing user1's `uploadsBasePath`. If `uploadsBasePath` is configured for a connection, then anyone with access to that connection can PUT files from that location. If isolation is required by the client then they should maintain a separate `Client()` for each user.

cc: @moderakh

src/databricks/sql/thrift_backend.py
susodapop: Resolved f6127950d0aaa68575eb24cb0bcceb066b6830e5

src/databricks/sql/client.py
susodapop: Resolved in e609ef3d843a6474b55162f990b8969d5d5633e6

susodapop: Resolved here cdbe2d6b9c560209391aef497ec3798ede8408d3

src/databricks/sql/client.py
susodapop: Good catch.

tl;dr I updated the code in 34a0362a81de1c5ceeeaca681d084fb2e9ed5249 so that it resolves any relative paths _before_ checking for their common_path. I added a test to prove this.

## Before

`/Users/user1` and `/Users/user1/../user2` show a common path of `/Users/user1` which is **wrong**.

## After

`/Users/user1` and `/Users/user1/../user2` show a common path of `/Users` which is correct.

susodapop: @moderakh Are there other cases we should consider?

src/databricks/sql/client.py
susodapop: > Also, we should add some tests like this, we need to make sure the resolved path still start with uploads_base_path

@HSCandYH Done as part of 34a0362a81de1c5ceeeaca681d084fb2e9ed5249 👌

tests/e2e/driver_tests.py
susodapop: Done!

> Two Puts to the same staging path should fail the second one: c8a64c78d8d9ac0477e2dbf582d03d16239cb7b5
> Put/Get/Delete files in another user's staging location should fail with permission-denied errors: d48d3f367392fd4cf1f166210c45c609a28d3002
> Empty localPath argument in the query should fail with parsing errors on server side: e0037e0a8a5e5b224f3d031ec6de6a5f330470e8
> Invalid stagingPath argument in the query should fail with parsing errors on server side: 3fa5d84b7e3f4b25e3ce8826e7e2bc0b66e2bdee

moderakh:
thanks @susodapop my answers inline:

> I've pushed in the `uploadsBasePath` restriction and associated tests. I have a few questions for the group though:
> 
> 1. Should we allow `uploads_base_path` to be a list of allowed paths?

yes

> 2. Do we need to require `uploads_base_path` for REMOVE operations?

yes

> 3. Should we re-name `uploads_base_path` to `allowed_local_dirs` since it's technically restricting not only what data we PUT but what data we GET?

I thinks we should make this plural but `allowed_local_dirs` is not clear to what feature we are referring.
maybe `uploads_base_paths` ?

> 4. Right now we require`uploads_base_path` be provide and be a path-like string. Do we need any other explicit constraints? For example we could prevent users from specifying the root directory `'/'` as their `uploads_base_path`. Is that desireable?

I don't think we can do that that requires knowledge of specific OS filesystem structure and home user path. We rely on the end user to configure that.

> 
> cc @moderakh @HSCandYH

src/databricks/sql/client.py
moderakh: I consider this outside of the scope of this PR. we should follow up on retry behaviour later after this PR.

src/databricks/sql/client.py
moderakh: ingestion only supports uploading from local file system not elsewhere. anything else must fail.

PR #60, opened 2022-10-26
Support Python 3.11

susodapop:
## Description

**Status**: This PR is on-hold awaiting PyArrow to release their pre-built wheels for Python 3.11.

This PR adds Python 3.11 to the unit testing matrix.

Closes https://github.com/databricks/databricks-sql-python/issues/59

PR #70, opened 2022-11-28
Python 3.11 support: bump version to 2.2.1 

susodapop:
After #60 we can cut a release with a patch version increment.


moderakh (approved):
LGTM. thanks

PR #71, opened 2022-12-13
Add none check on _oauth_persistence in DatabricksOAuthProvider

jackyhu-db:
Add `None` check on `_oauth_persistence` in `DatabricksOAuthProvider` to avoid app crash when `_oauth_persistence` is `None`.


moderakh (approved):
Thanks @jackyhu-db for the fix.

PR #74, opened 2022-12-15
Less strict dependencies

Tom-Newton:
Currently this library has quite strict requirements for `pyarrow` and `numpy` versions which significantly limits the usefulness. As far as I can tell this library does not require any particular features of `numpy` or `pyarrow` added in recent versions. 

Signed-off-by: Thomas Newton <thomas.w.newton@gmail.com>


pyproject.toml
```diff
@@ -13,15 +13,15 @@ python = "^3.7.1"
...
+    {version = ">=6.0.0", python = ">=3.7,<3.11"},
     {version = ">=10.0.1", python = ">=3.11"}
 ]
 lz4 = "^4.0.2"
 requests="^2.18.1"
 oauthlib="^3.1.0"
 numpy = [
-    {version = "1.21.1", python = ">=3.7,<3.8"},
-    {version = "1.23.4", python = ">=3.8"}
+    {version = ">=1.16.6", python = ">=3.7,<3.8"},
```
Tom-Newton, line 23: It looks like `numpy` is barely used in this library so I think pretty much any version should work. I chose `>=1.16.6` because that is what the latest version of `pyarrow` requires.

pyproject.toml
```diff
@@ -13,15 +13,15 @@ python = "^3.7.1"
...
     {version = ">=10.0.1", python = ">=3.11"}
 ]
 lz4 = "^4.0.2"
 requests="^2.18.1"
 oauthlib="^3.1.0"
 numpy = [
-    {version = "1.21.1", python = ">=3.7,<3.8"},
-    {version = "1.23.4", python = ">=3.8"}
+    {version = ">=1.16.6", python = ">=3.7,<3.11"},
+    {version = ">=1.23.4", python = ">=3.11"}
```
Tom-Newton, line 24: It looks like the special requirement for python 3.11 was added recently in https://github.com/databricks/databricks-sql-python/pull/60 so I assume that is important but I changed it to accept >= to this version. 

I also think we should only enforce this requirement on python 3.11

pyproject.toml
```diff
@@ -13,15 +13,15 @@ python = "^3.7.1"
 thrift = "^0.16.0"
 pandas = "^1.3.0"
 pyarrow = [
-    {version = ">=9.0.0", python = ">=3.7,<3.11"},
+    {version = ">=6.0.0", python = ">=3.7,<3.11"},
```
Tom-Newton, line 16: Probably we could allow even older versions but I think this should make most people happy.

Tom-Newton:
Looks like there is one failure on the CI checks 

```
Download action repository 'actions/checkout@v2' (SHA:dc323e67f16fb5f7663d20ff7941f27f5809e9b6)
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/dc[32](https://github.com/databricks/databricks-sql-python/actions/runs/4168062376/jobs/7214961091#step:1:36)3e67f16fb5f7663d20ff7941f27f5809e9b6'. Error: Resource temporarily unavailable (api.github.com:443)
Warning: Back off 22.447 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/dc323e67f16fb5f7663d20ff7941f27f5809e9b6'. Error: Resource temporarily unavailable (api.github.com:443)
Warning: Back off 24.765 seconds before retry.
Error: Resource temporarily unavailable (api.github.com:443)
```
I don't think this can be because of my changes. I guess its just a transient thing but I don't think I can re-trigger it.

Tom-Newton:
Thanks for triggering the checks @susodapop. It looks like all the unittests passed for all the different python versions. One of the code quality checks is failing though. From the error it looks like it just needs a retry. 

If someone could retry that check and review the PR that would be much appreciated.

Tom-Newton:
@susodapop sorry to be impatient but please could you review this? I think this is about to start causing us some significant problems if its not fixed.

susodapop:
Not impatient at all! We just needed to freeze our dependencies to get v2.4.0 out the door. Else this would have merged already.

Tom-Newton:
Superseded by https://github.com/databricks/databricks-sql-python/pull/90

PR #64, opened 2022-11-03
Regenerate TCLIService using latest TCLIService.thrift from DBR

susodapop:
This is a duplicate of #61 but targeting the `si` branch (which I just forked off `main`). I have regenerated the TCLIService from the latest TCLIService.thrift within DBR.

PR #78, opened 2023-01-09
Merge staging ingestion into main

susodapop:
This is a follow-up to:

- https://github.com/databricks/databricks-sql-python/pull/67
- https://github.com/databricks/databricks-sql-python/pull/64

It merges the staging ingestion feature into `main` so we can release it to Pypi.

We need to re-lock our dependencies because the `poetry.lock` file is different between these branches. To do this:

1. Merge `main` back into `si` (taking the version of `poetry.lock` from `main`). At this point, the lock file and `pyproject.toml` disagree with one another and will raise an error.
2. `rm poetry.lock`
3. `poetry install` which regenerates a new lockfile from `pyproject.toml`


susodapop:
e2e tests pass

moderakh (approved):
Thanks @susodapop

Issue #22, opened 2022-07-25
Issues with using http proxy


andresebastian-moelle-ext:
Hi there,

i am not able to use the sql.connect() with a http proxy:

```
os.environ['http_proxy'] = f"http://{user}:{pwd}@10.185.190.100:8080"
os.environ['https_proxy'] = f"http://{user}:{pwd}@10.185.190.100:8080"
self.connection = sql.connect(server_hostname, http_path, access_token)
```

I am getting the following error:

```
File "C:\Users\User\Documents\GitHub\eve_data_ingestion\3.8.10 64 bit\lib\site-packages\databricks\sql\__init__.py", line 48, in connect
    return Connection(server_hostname, http_path, access_token, **kwargs)
  File "C:\Users\User\Documents\GitHub\eve_data_ingestion\3.8.10 64 bit\lib\site-packages\databricks\sql\client.py", line 109, in __init__
    self.thrift_backend = ThriftBackend(self.host, self.port, http_path,
  File "C:\Users\User\Documents\GitHub\eve_data_ingestion\3.8.10 64 bit\lib\site-packages\databricks\sql\thrift_backend.py", line 115, in __init__
    self._transport = thrift.transport.THttpClient.THttpClient(
  File "C:\Users\User\Documents\GitHub\eve_data_ingestion\3.8.10 64 bit\lib\site-packages\thrift\transport\THttpClient.py", line 86, in __init__
    self.proxy_auth = self.basic_proxy_auth_header(parsed)
  File "C:\Users\User\Documents\GitHub\eve_data_ingestion\3.8.10 64 bit\lib\site-packages\thrift\transport\THttpClient.py", line 101, in basic_proxy_auth_header
    cr = base64.b64encode(ap).strip()
  File "C:\Users\User\AppData\Local\Programs\Python\Python38\lib\base64.py", line 58, in b64encode
    encoded = binascii.b2a_base64(s, newline=False)

TypeError: a bytes-like object is required, not 'str'
```

Is this a known issue? Is there another way to use proxies?
_______________
Python 3.8.10 (64bit Windows)
databricks-sql-connector 2.0.2

susodapop:
Proxies aren't supported yet. We have an internal ticket to implement soon.

Related: https://github.com/databricks/dbt-databricks/issues/111

andresebastian-moelle-ext:
@susodapop Thanks for the quick answer! Is there any workaround possible at the moment? Or any eta when this will be implemented? :)

[...]

PR #81, opened 2023-01-12
[PECO-244] Make http proxies work

susodapop:
## Description

This fix makes the connector work with standard proxies using Python's [environment variable configuration spec](https://docs.python.org/3/library/urllib.request.html#urllib.request.getproxies). This fix is taken directly from @pspeter's pull request against the `thrift` library here: https://github.com/apache/thrift/pull/2565

Since we are already overriding the `THttpClient`, we don't need to wait for the fix to merge into `thrift` upstream to implement it locally.

Closes https://github.com/databricks/databricks-sql-python/issues/22

### DIFF
--- a/src/databricks/sql/auth/thrift_http_client.py
+++ b/src/databricks/sql/auth/thrift_http_client.py
@@ -39,7 +39,6 @@ class THttpClient(thrift.transport.THttpClient.THttpClient):
 
     @staticmethod
     def basic_proxy_auth_header(proxy):
-        logger.warning("Hey I was called!")
         if proxy is None or not proxy.username:
             return None
         ap = "%s:%s" % (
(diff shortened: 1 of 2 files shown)


andrefurlan-db (approved):
looks good to me

rcypher-databricks (approved):

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -33,3 +36,15 @@ def flush(self):
         self._headers = headers
         self.setCustomHeaders(self._headers)
         super().flush()
+
+    @staticmethod
+    def basic_proxy_auth_header(proxy):
+        logger.warning("Hey I was called!")
```
arikfr, line 42:
```suggestion
```

Probably leftover from debugging?

PR #84, opened 2023-01-19
[ES-556959] Add pkgutil-style for the package

lu-wang-dl:
Since the package is under databricks namespace. pip install this package will cause issue importing other packages under the same namespace like automl and feature store.

Adding `pkgutil` style to resolve the issue.


susodapop (approved):
LGTM. Thanks!

PR #86, opened 2023-01-30
[PECO-626] Support OAuth flow for Databricks Azure

jackyhu-db:
## Summary

Support OAuth flow for Databricks Azure

## Background

Some OAuth endpoints (e.g. Open ID Configuration) and scopes are different between Databricks Azure and AWS. Current code only supports OAuth flow on Databricks in AWS

## What changes are proposed in this pull request?

- Change `OAuthManager` to decouple Databricks AWS specific configuration from OAuth flow
- Add `sql/auth/endpoint.py` that implements cloud specific OAuth endpoint configuration
- Change `DatabricksOAuthProvider` to work with the OAuth configurations in different Databricks cloud (AWS, Azure)
- Add the corresponding unit tests


src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+# Infer cloud type from Databricks SQL instance hostname
+def infer_cloud_from_host(hostname: str) -> Optional[CloudType]:
+    # normalize
+    host = hostname.lower().replace("https://", "").split("/")[0]
+
+    if host.endswith(".azuredatabricks.net"):
+        return CloudType.AZURE
+    elif host.endswith(".gcp.databricks.com"):
```
susodapop, line 34: Should we raise an exception for GCP since that cloud doesn't support OAuth even experimentally at this time?

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+    elif host.endswith("cloud.databricks.com"):
+        return CloudType.AWS
+    else:
+        return None
+
+
+def get_databricks_oidc_url(hostname: str):
+    maybe_scheme = "https://" if not hostname.startswith("https://") else ""
+    maybe_trailing_slash = "/" if not hostname.endswith("/") else ""
+    return f"{maybe_scheme}{hostname}{maybe_trailing_slash}{OIDC_REDIRECTOR_PATH}"
```
susodapop, line 45: Can we use Python's built-in URL parsing for this? The built-in is called [`urllib.parse`](https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse).

susodapop: Actually it looks like this was moved from the previous implementation so no change is needed here.

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+        raise NotImplementedError()
+
+    # Endpoint for well-known openid configuration e.g https://idp.example.com/oauth2/.well-known/openid-configuration
+    @abstractmethod
+    def get_openid_config_endpoint(self, hostname: str) -> str:
+        raise NotImplementedError()
+
+
+class OAuthEndpointsAzure(OAuthEndpoints):
+    SCOPE_USER_IMPERSONATION = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d/user_impersonation"
```
susodapop, line 65: Is this supposed to be hard-coded?

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def get_scopes_mapping(self, scopes: List[str]) -> List[str]:
+        # There is no corresponding scopes in Azure, instead, access control will be delegated to Databricks
+        # To support scope in dev, it can also be set in the environment variable DATABRICKS_AZURE_SCOPE
+        azure_scope = (
+            os.getenv("DATABRICKS_AZURE_SCOPE")
+            or OAuthEndpointsAzure.SCOPE_USER_IMPERSONATION
+        )
+        ret_scopes = [azure_scope]
```
susodapop, line 74: What does `ret_scopes` stand for? Clearer naming might help.

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+        return None
+
+
+def get_databricks_oidc_url(hostname: str):
+    maybe_scheme = "https://" if not hostname.startswith("https://") else ""
+    maybe_trailing_slash = "/" if not hostname.endswith("/") else ""
+    return f"{maybe_scheme}{hostname}{maybe_trailing_slash}{OIDC_REDIRECTOR_PATH}"
+
+
+class OAuthEndpoints(ABC):
```
susodapop, line 48: It's unusual to me to see Python class names expressed as plurals. If this object exposes a collection then that should be demonstrated by the name. Something like `OAuthEndpointCollection` or `OAuthEndpoingList` (note that both are singular).

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+    def get_authorization_endpoint(self, hostname: str) -> str:
+        raise NotImplementedError()
+
+    # Endpoint for well-known openid configuration e.g https://idp.example.com/oauth2/.well-known/openid-configuration
+    @abstractmethod
+    def get_openid_config_endpoint(self, hostname: str) -> str:
+        raise NotImplementedError()
+
+
+class OAuthEndpointsAzure(OAuthEndpoints):
```
susodapop, line 64: Likewise for class naming convention: I would name this something like `AzureOAuthEndpointCollection` so that the attribute which makes it unique `Azure` appears early in the name rather than later.

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+
+def get_databricks_oidc_url(hostname: str):
+    maybe_scheme = "https://" if not hostname.startswith("https://") else ""
+    maybe_trailing_slash = "/" if not hostname.endswith("/") else ""
+    return f"{maybe_scheme}{hostname}{maybe_trailing_slash}{OIDC_REDIRECTOR_PATH}"
+
+
+class OAuthEndpoints(ABC):
+    @abstractmethod
+    def get_scopes_mapping(self, scopes: List[str]) -> List[str]:
```
susodapop, line 50: Is there anywhere that explains what this function is supposed to return? It accepts a list of strings and returns a list of strings. Typically in Python a mapping is represented as a `dict()`.

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+    def get_authorization_endpoint(self, hostname: str):
+        # We need get account specific url, which can be redirected by databricks unified oidc endpoint
+        return f"{get_databricks_oidc_url(hostname)}/oauth2/v2.0/authorize"
+
+    def get_openid_config_endpoint(self, hostname: str):
+        return "https://login.microsoftonline.com/organizations/v2.0/.well-known/openid-configuration"
+
+
+class OAuthEndpointsAws(OAuthEndpoints):
+    def get_scopes_mapping(self, scopes: List[str]) -> List[str]:
```
susodapop, line 88: Related to the above comment, is it normal for this function to return one of its arguments untouched?

src/databricks/sql/auth/endpoint.py
susodapop: Elsewhere when GCP is introduced those functions return `None`. I think this should probably do the same.

src/databricks/sql/auth/oauth.py
```diff
@@ -14,17 +14,19 @@
...
-    OIDC_REDIRECTOR_PATH = "oidc"
-
-    def __init__(self, port_range: List[int], client_id: str):
+    def __init__(
+        self, port_range: List[int], client_id: str, idp_endpoint: OAuthEndpoints
+    ):
         self.port_range = port_range
         self.client_id = client_id
         self.redirect_port = None
+        self.idp_endpoint = idp_endpoint
```
susodapop, line 29: By changing this function signature does this break the implementation for anyone using AWS after they upgrade their version of this connector?

tests/unit/test_auth.py
```diff
@@ -1,7 +1,24 @@
...
 
-from databricks.sql.auth.auth import AccessTokenAuthProvider, BasicAuthProvider, AuthProvider
+from databricks.sql.auth.auth import AccessTokenAuthProvider, BasicAuthProvider, AuthProvider, DatabricksOAuthProvider
 from databricks.sql.auth.auth import get_python_sql_connector_auth_provider
+from databricks.sql.auth.oauth import OAuthManager
+from databricks.sql.auth.endpoint import OAuthEndpointsAzure, OAuthEndpointsAws, CloudType
+from databricks.sql.experimental.oauth_persistence import OAuthPersistence, OAuthToken
+
+
+class OAuthPersistenceCache(OAuthPersistence):
```
susodapop, line 13: Is there any reason we _shouldn't_ just make this type of cache available to end-users @moderakh? Right now our examples simply tell users to roll their own cache. This in-memory demo could be useful for anyone starting out.

If not, I think we may as well name it `TestOAuthPersistenceCache` to make it obvious this is only for testing purposes.

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def get_authorization_endpoint(self, hostname: str):
+        # We need get account specific url, which can be redirected by databricks unified oidc endpoint
+        return f"{get_databricks_oidc_url(hostname)}/oauth2/v2.0/authorize"
+
+    def get_openid_config_endpoint(self, hostname: str):
+        return "https://login.microsoftonline.com/organizations/v2.0/.well-known/openid-configuration"
+
+
+class OAuthEndpointsAws(OAuthEndpoints):
```
susodapop, line 87: Ditto comments on class names
```suggestion
class AwsOAuthEndpointCollection(OAuthEndpoints):
```

tests/unit/test_endpoint.py
```diff
@@ -0,0 +1,56 @@
...
+import os
+import pytest
+
+from unittest.mock import patch
+
+from databricks.sql.auth.endpoint import infer_cloud_from_host, CloudType, get_oauth_endpoints, OAuthEndpointsAzure
+
+aws_host = "foo-bar.cloud.databricks.com"
+azure_host = "foo-bar.1.azuredatabricks.net"
+gcp_host = "foo-bar.gcp.databricks.com"
```
susodapop, line 11: Ditto we shouldn't be testing GCP if it's not supported at all.

src/databricks/sql/auth/endpoint.py
susodapop: Looking at the tests, it seems like this function returns a filtered version of the list it receives. In which case it may be better to name it `get_allowed_scopes` rather than `get_scopes_mapping` imo.

susodapop (commented):
I have a lot of questions but most of them are small. The big question is whether this implementation continues to work for existing AWS public preview users. Some manual testing in this regard will be fine. You can use the demo code shown in the `examples` directory to make a quick POC.

src/databricks/sql/auth/endpoint.py
jackyhu-db: Yes, this is a constant value.

src/databricks/sql/auth/endpoint.py
jackyhu-db: I will rename it. Thanks.

src/databricks/sql/auth/oauth.py
jackyhu-db: `OAuthManager` is only internally used by the PySQL. The consumer of the SDK uses the API `connect` in `databricks/sql/client.py`, and it does not have `endpoint` parameter

tests/unit/test_auth.py
jackyhu-db: This is only for unit test. Yes, we can name it to `TestOAuthPersistenceCache`

src/databricks/sql/auth/endpoint.py
jackyhu-db: I renamed it to `mapped_scopes`

src/databricks/sql/auth/endpoint.py
jackyhu-db: I changed it by copying the original scopes.

tests/unit/test_auth.py
jackyhu-db: I moved `OAuthPersistenceCache` to `oauth_persistence.py` so it can also be reused by others.

tests/unit/test_endpoint.py
jackyhu-db: Removed.

jackyhu-db:
> I have a lot of questions but most of them are small. The big question is whether this implementation continues to work for existing AWS public preview users. Some manual testing in this regard will be fine. You can use the demo code shown in the `examples` directory to make a quick POC.

It does not break existing AWS OAuth flow. It has been tested via unit test and end-to-end test (`examples/interactive_oauth.py`) on both AWS and Azure

susodapop (approved):
LGTM. I'd like to get @moderakh's approval also.

tests/unit/test_auth.py
susodapop: SGTM

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+# Infer cloud type from Databricks SQL instance hostname
+def infer_cloud_from_host(hostname: str) -> Optional[CloudType]:
+    # normalize
+    host = hostname.lower().replace("https://", "").split("/")[0]
+
+    if host.endswith(".azuredatabricks.net"):
+        return CloudType.AZURE
+    elif host.endswith("cloud.databricks.com") or host.endswith("dev.databricks.com"):
```
moderakh, line 33: how is this impacted a proxy is used? or if someone defines a dns CNAME for databricks endpoint?

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,104 @@
...
+    AWS = "aws"
+    AZURE = "azure"
+
+
+# Infer cloud type from Databricks SQL instance hostname
+def infer_cloud_from_host(hostname: str) -> Optional[CloudType]:
+    # normalize
+    host = hostname.lower().replace("https://", "").split("/")[0]
+
+    if host.endswith(".azuredatabricks.net"):
```
moderakh, line 31: there are other azure endpoints as well, do we support them all?

- ".azuredatabricks.net"
- ".databricks.azure.cn"
- ".databricks.azure.us"

src/databricks/sql/auth/auth.py
```diff
@@ -79,6 +81,14 @@ def normalize_host_name(hostname: str):
     return f"{maybe_scheme}{hostname}{maybe_trailing_slash}"
 
 
+def get_client_id(hostname: str):
+    return (
+        PYSQL_OAUTH_CLIENT_ID
+        if infer_cloud_from_host(hostname) == CloudType.AWS
+        else PYSQL_OAUTH_CLIENT_ID_AZURE
```
moderakh, line 88: what happens if other clouds are used? GCP? should we throw error?

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class AzureOAuthEndpointCollection(OAuthEndpointCollection):
+    SCOPE_USER_IMPERSONATION = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d/user_impersonation"
+
+    def get_scopes_mapping(self, scopes: List[str]) -> List[str]:
+        # There is no corresponding scopes in Azure, instead, access control will be delegated to Databricks
+        # To support scope in dev, it can also be set in the environment variable DATABRICKS_AZURE_SCOPE
+        azure_scope = (
+            os.getenv("DATABRICKS_AZURE_SCOPE")
```
moderakh, line 68: why do we need this additional env variable? why can't we always return 
`AzureOAuthEndpointCollection.SCOPE_USER_IMPERSONATION`

it is not clear what this maybe we should have "OAUTH" in the name.

It would be the best if we don't need it, we remove `DATABRICKS_AZURE_SCOPE`?

moderakh (approved):
@jackyhu-db please take a. look at the comments. other than that LGTM.

src/databricks/sql/auth/endpoint.py
jackyhu-db: Good to know. I will add them.

src/databricks/sql/auth/auth.py
jackyhu-db: The error will be thrown when it tires to get the endpoint configuration if cloud type is neither AWS nor Azure.

src/databricks/sql/auth/endpoint.py
jackyhu-db: The scope is different in our internal dev/test system as the directory is different.

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,112 @@
...
+class AzureOAuthEndpointCollection(OAuthEndpointCollection):
+    DATATRICKS_AZURE_TENANT_ID = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"
+
+    def get_scopes_mapping(self, scopes: List[str]) -> List[str]:
+        # There is no corresponding scopes in Azure, instead, access control will be delegated to Databricks
+        tenant_id = os.getenv(
+            "DATABRICKS_AZURE_TENANT_ID",
+            AzureOAuthEndpointCollection.DATATRICKS_AZURE_TENANT_ID,
+        )
+        azure_scope = f"{tenant_id}/user_impersonation"
```
moderakh, line 78: seems for tableau we are getting many other scopes as well:
https://github.com/databricks/tableau-connector/blob/master/tableau-databricks/oauth-config.xml#L16-L20

any idea why these are not required here?

src/databricks/sql/auth/endpoint.py
```diff
@@ -0,0 +1,112 @@
...
+        raise NotImplementedError()
+
+    # Endpoint for well-known openid configuration e.g https://idp.example.com/oauth2/.well-known/openid-configuration
+    @abstractmethod
+    def get_openid_config_url(self, hostname: str) -> str:
+        raise NotImplementedError()
+
+
+class AzureOAuthEndpointCollection(OAuthEndpointCollection):
+    DATATRICKS_AZURE_TENANT_ID = "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"
```
nfx, line 70: this is not a tenant, but is Enterprise App instead. Is misleading for those reading the code

nfx: this one also won't work for our staging environment

src/databricks/sql/auth/auth.py
```diff
@@ -70,6 +71,7 @@ def get_auth_provider(cfg: ClientContext):
 
 PYSQL_OAUTH_SCOPES = ["sql", "offline_access"]
 PYSQL_OAUTH_CLIENT_ID = "databricks-sql-python"
+PYSQL_OAUTH_AZURE_EXPERIMENTAL_CLIENT_ID = "a743d78c-536a-4ffc-b110-edfb231e90dc"
```
nfx, line 74: is it Azure Enterprise Application? it's not visible in the AAD.

nfx (changes requested):
Please strongly consider depending on https://pypi.org/project/databricks-sdk/ instead

src/databricks/sql/auth/endpoint.py
jackyhu-db: updated.

src/databricks/sql/auth/endpoint.py
jackyhu-db: `openid profile email` are OpenID scopes and only used to generate the `id_token` which is not required by PySQL

src/databricks/sql/auth/auth.py
jackyhu-db: Updated the client_id with the published Azure app ID

susodapop (approved):

susodapop:
As as final ask, will you please update the CHANGELOG?

PR #92, opened 2023-02-27
Fix syntax in examples in root readme.

shea-parkes:
Do this because the environment variable pulls did not have closing quotes on their string literals.


susodapop (approved):
LGTM. Nice catch thanks!

PR #93, opened 2023-03-07
Fix logging statements in oauth.py

jbylund:
* fix a couple logging statements (lazy parameters, right number of args), left one as a string interpolation because the message is also used in an exception
* add some timeouts to http requests

### DIFF
--- a/src/databricks/sql/auth/oauth.py
+++ b/src/databricks/sql/auth/oauth.py
@@ -17,6 +17,7 @@ from databricks.sql.auth.oauth_http_handler import OAuthHttpSingleRequestHandler
 
 logger = logging.getLogger(__name__)
 
+# pylint: disable=invalid-name
 
 class OAuthManager:
     OIDC_REDIRECTOR_PATH = "oidc"
@@ -38,12 +39,12 @@ class OAuthManager:
     def __fetch_well_known_config(idp_url: str):
         known_config_url = f"{idp_url}/.well-known/oauth-authorization-server"
         try:
-            response = requests.get(url=known_config_url)
+            response = requests.get(url=known_config_url, timeout=10)
         except RequestException as e:
             logger.error(
-                f"Unable to fetch OAuth configuration from {idp_url}.\n"
+                "Unable to fetch OAuth configuration from %s.\n"
                 "Verify it is a valid workspace URL and that OAuth is "
-                "enabled on this account."
+                "enabled on this account.", idp_url
             )
             raise e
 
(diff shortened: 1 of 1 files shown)


src/databricks/sql/auth/oauth.py
```diff
@@ -38,12 +40,13 @@ def __get_redirect_url(redirect_port: int):
     def __fetch_well_known_config(idp_url: str):
         known_config_url = f"{idp_url}/.well-known/oauth-authorization-server"
         try:
-            response = requests.get(url=known_config_url)
+            response = requests.get(url=known_config_url, timeout=10)
         except RequestException as e:
             logger.error(
-                f"Unable to fetch OAuth configuration from {idp_url}.\n"
+                "Unable to fetch OAuth configuration from %s.\n"
```
susodapop, line 46: Why the change from f-strings throughout?

src/databricks/sql/auth/oauth.py
```diff
@@ -38,12 +40,13 @@ def __get_redirect_url(redirect_port: int):
     def __fetch_well_known_config(idp_url: str):
         known_config_url = f"{idp_url}/.well-known/oauth-authorization-server"
         try:
-            response = requests.get(url=known_config_url)
+            response = requests.get(url=known_config_url, timeout=10)
```
susodapop, line 43: Generally we like to keep these timeouts configurable. Perhaps when instantiating the OAuthManager?

susodapop (changes requested):
Sorry if I'm being foggy-brained here, but apart from the timeouts I'm not clear what this PR actually changes? The log messages appear to be the same, but using the older string interpolation syntax. For this repository f-strings are the preferred direction.

jbylund:
> Sorry if I'm being foggy-brained here, but apart from the timeouts I'm not clear what this PR actually changes? The log messages appear to be the same, but using the older string interpolation syntax. For this repository f-strings are the preferred direction.

I had assumed %-formatting and passing args to the logging call would be preferred because that's the recommended pattern for logging. I think there are some benefits to using %-formatting for logging calls (it's lazy which helps performance and makes writing log handlers easier).

I don't need this pr any longer (and it's got a conflict now), so unless you want to change the logging style I'll just close this. If you would like to change the logging formatting repo wide I'd be happy to take that on and maybe add a check to the ci pipeline.

src/databricks/sql/auth/oauth.py
```diff
@@ -96,24 +100,25 @@ def __get_authorization_code(self, client, auth_url, scope, state, challenge):
...
                     httpd.handle_request()
                 self.redirect_port = port
                 break
             except OSError as e:
                 if e.errno == 48:
-                    logger.info(f"Port {port} is in use")
+                    logger.info("Port %d is in use", port)
                     last_error = e
             except Exception as e:
-                logger.error("unexpected error", e)
```
jbylund, line 113: This is probably the issue that brought me here in the first place. The format string here has no positions for arguments, but one argument was provided.

susodapop:
I think for now we should certainly fix the failed logging statement on line 113 (marker included but no value provided). I can do that with this current PR so you get the contribution credit or we can file it as a follow-up. Any preference?

susodapop:
Closing in favour of #269. This change will be part of the next release.

PR #101, opened 2023-04-07
Add external auth provider

andrefurlan-db:
Using Databricks Python SDK as external API reference.

This allows for skipping authentication in the driver and using any auth mechanism available - custom or from databricks-sdk.

### DIFF
--- a/examples/custom_cred_provider.py
+++ b/examples/custom_cred_provider.py
@@ -1,3 +1,5 @@
+# please install databricks-sdk prior to running this example.
+
 from databricks import sql
 from databricks.sdk.oauth import OAuthClient
 import os
@@ -16,7 +18,7 @@ with sql.connect(server_hostname = os.getenv("DATABRICKS_SERVER_HOSTNAME"),
                  http_path       = os.getenv("DATABRICKS_HTTP_PATH"),
                  credentials_provider=creds) as connection:
 
-    for x in range(1, 100):
+    for x in range(1, 5):
         cursor = connection.cursor()
         cursor.execute('SELECT 1+1')
         result = cursor.fetchall()
--- a/examples/custom_cred_provider.py
+++ b/examples/custom_cred_provider.py
@@ -1,4 +1,4 @@
-# please pip install databricks-sdk prior to running this example.
+# please install databricks-sdk prior to running this example.
 
 from databricks import sql
 from databricks.sdk.oauth import OAuthClient
--- a/examples/custom_cred_provider.py
+++ b/examples/custom_cred_provider.py
@@ -1,0 +1,27 @@
+from databricks import sql
+from databricks.sdk.oauth import OAuthClient
+import os
+
+oauth_client = OAuthClient(host=os.getenv("DATABRICKS_SERVER_HOSTNAME"),
+                           client_id=os.getenv("DATABRICKS_CLIENT_ID"),
+                           client_secret=os.getenv("DATABRICKS_CLIENT_SECRET"),
+                           redirect_url=os.getenv("APP_REDIRECT_URL"),
+                           scopes=['all-apis', 'offline_access'])
+
+consent = oauth_client.initiate_consent()
+
+creds = consent.launch_external_browser()
+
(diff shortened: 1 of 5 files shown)


examples/custom_cred_provider.py
```diff
@@ -0,0 +1,27 @@
+from databricks import sql
+from databricks.sdk.oauth import OAuthClient
```
susodapop, line 2: This import only works if databricks-sdk is installed, which is not currently part of `pyproject.toml`. I can show you how to add this requirement and regenerate poetry.lock. Else we should note in a docstring or comment that users should `pip install databricks-sdk`

examples/custom_cred_provider.py
```diff
@@ -0,0 +1,27 @@
...
+import os
+
+oauth_client = OAuthClient(host=os.getenv("DATABRICKS_SERVER_HOSTNAME"),
+                           client_id=os.getenv("DATABRICKS_CLIENT_ID"),
+                           client_secret=os.getenv("DATABRICKS_CLIENT_SECRET"),
+                           redirect_url=os.getenv("APP_REDIRECT_URL"),
+                           scopes=['all-apis', 'offline_access'])
```
susodapop, line 9: Are these scopes documented anywhere?

examples/custom_cred_provider.py
```diff
@@ -0,0 +1,27 @@
...
+
+consent = oauth_client.initiate_consent()
+
+creds = consent.launch_external_browser()
+
+with sql.connect(server_hostname = os.getenv("DATABRICKS_SERVER_HOSTNAME"),
+                 http_path       = os.getenv("DATABRICKS_HTTP_PATH"),
+                 credentials_provider=creds) as connection:
+
+    for x in range(1, 100):
```
susodapop, line 19:
```suggestion
    for x in range(1, 5):
```

It seems excessive for demo code to make 100 roundtrips to prove it works 🤷

src/databricks/sql/auth/authenticators.py
```diff
@@ -14,6 +15,22 @@ def add_headers(self, request_headers: Dict[str, str]):
...
+HeaderFactory = Callable[[], Dict[str, str]]
+
+# In order to keep compatibility with SDK
+class CredentialsProvider(abc.ABC):
+    """CredentialsProvider is the protocol (call-side interface)
+    for authenticating requests to Databricks REST APIs"""
+
+    @abc.abstractmethod
+    def auth_type(self) -> str:
+        ...
```
susodapop, line 27: What does ellipsis `...` do here? I've never seen that syntax in Python

tests/unit/test_auth.py
```diff
@@ -37,6 +38,22 @@ def test_noop_auth_provider(self):
         self.assertEqual(len(http_request.keys()), 1)
         self.assertEqual(http_request['myKey'], 'myVal')
 
+    def test_external_provider(self):
+        class MyProvider(CredentialsProvider):
+                def auth_type(self) -> str:
+                    return "mine"
+
+                def __call__(self, *args, **kwargs) -> HeaderFactory:
+                    return lambda: {"foo": "bar"}
```
susodapop, line 47: Clever trick!

susodapop (approved):
Looks good. Couple comments.

examples/custom_cred_provider.py
andrefurlan-db: I added a comment in the file and readme. Thanks for pointing it out!

examples/custom_cred_provider.py
andrefurlan-db: yes, they are. But this is just the example, so I went with the basic ones

examples/custom_cred_provider.py
andrefurlan-db: yeap, you're right

examples/custom_cred_provider.py
```diff
@@ -1,3 +1,5 @@
+# please install databricks-sdk prior to running this example.
```
susodapop, line 1:
```suggestion
# please pip install databricks-sdk prior to running this example.
```

susodapop (approved):
Might as well show the full command here.

PR #105, opened 2023-04-15
Automatically run integration tests

susodapop:
This pull request turns on automatic integration test execution following the pattern of dbt-databricks.

- Tests will run automatically against our DECO workspace whenever changes are pushed to a branch _on this repository_ but not on pushes to branches of forks.

Work still required:

- [ ] DECO team needs to make their environment secrets available to this repository (all tests fail as of 15 April 2023)
- [ ] CONTRIBUTING guide needs to be updated similar to dbt-databricks
- [ ] Some of the tests should be temporarily marked as skipped because they've been failing for awhile (not a pysql issue -- just changes to thrift server)

Notes:

* I moved the staging ingestion tests into their own mixin so they would match the pattern of the other driver tests.
* I'd like to add a real integration test that proves our http proxies work as well

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,61 @@
+name: Integration Tests
+on:
+  push
+jobs:
+  run-integration-tests:
+    runs-on: ubuntu-latest
+    environment: azure-prod-pecou
+    strategy:
+      matrix:
+        # python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
+          python-version: [3.7, 3.8]
+    env:
+      host: ${{ secrets.DATABRICKS_HOST }}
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -0,0 +1,61 @@
+name: Integration Tests
+on:
+  push
+jobs:
+  run-integration-tests:
+    runs-on: ubuntu-latest
+    environment: azure-prod-pecou
+    strategy:
+      matrix:
+        # python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
```
susodapop, line 10: Waiting to enable for the whole matrix until I can see it working for 3.7 and 3.8

susodapop:
Closing this PR as we've significantly re-architected some of these tests. Will give it another go once we have our new, dedicated infra for this.

PR #113, opened 2023-04-28
Relax sqlalchemy requirement

susodapop:
## Description

The existing sqlalchemy dependency specified version was the _current_ release of sqlalchemy at the time we developed our dialect. But it blocks some Databricks customers from using the connector if they have an older sqlalchemy installed. This PR relaxes our requirement since none of our core features are that picky about minor versions or patches.

I also added a segment to CONTRIBUTING.md that shows how to make changes like this in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 2.5.x (Unreleased)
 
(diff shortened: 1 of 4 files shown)


susodapop:
Passes e2e tests run locally on my machine.

PR #116, opened 2023-05-03
Relax panda and alembic requirements

bkyryliuk:
Pandas 1.3+ has a regression: https://github.com/pandas-dev/pandas/issues/42549
This PR proposes relaxing alembic and panda requirements


susodapop (approved):

susodapop:
I've mirrored this change in #119 so I can verify it builds. This change will be part of v2.5.2 releasing shortly.

PR #117, opened 2023-05-08
Fix SQLAlchemy timestamp converter + docs

susodapop:
This is to unblock langchain work. Tests coming in subsequent PR.

Resolves the case of reading a table with a snippet like this:

```python
tables = Table("system.information_schema.tables", metadata_obj, autoload_with=db_engine)
cursor = db_engine.execute(select(tables))
cursor.fetchone()
```

Which would fail with this error:

```
ValueError: Couldn't parse datetime string 'datetime.datetime(2023, 1, 5, 19, 8, 2, 16000)' - value is not a string.
```

This happened because DatabricksTimestamp assumed that TIMESTAMP types were returned as strings from the thrift connector. However TIMESTAMPS are actually returned as `DateTime` objects, so calling `str_to_datetime` was superfluous and unecessary.

We missed this in the initial development because we didn't test for it. I will make a follow-up PR that adds proper type mapping tests that we can build-on as we continue development of the SQLA dialect.

### DIFF
--- a/src/databricks/sqlalchemy/dialect/__init__.py
+++ b/src/databricks/sqlalchemy/dialect/__init__.py
@@ -59,7 +59,7 @@ class DatabricksDate(types.TypeDecorator):
     impl = types.DATE
 
     def process_result_value(self, value, dialect):
-        return processors.str_to_date(value)
+        return value
 
     def adapt(self, impltype, **kwargs):
         return self.impl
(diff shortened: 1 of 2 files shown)


src/databricks/sqlalchemy/dialect/__init__.py
```diff
@@ -47,7 +47,7 @@ class DatabricksTimestamp(types.TypeDecorator):
     impl = types.TIMESTAMP
 
     def process_result_value(self, value, dialect):
-        return processors.str_to_datetime(value)
+        return value
```
gengliangwang, line 50: what about date columns?

susodapop: Fixed in c520103f309275f28850ab6c5b3b8d4cfa3ff245

andrefurlan-db (approved):

gengliangwang (approved):

susodapop:
Existing alembic e2e tests pass.

PR #122, opened 2023-05-12
OAuth: don't override auth headers with contents of .netrc file

### LINKED ISSUE #121: OAuth breaks if a bad .netrc file is present
While fixing https://github.com/databricks/dbt-databricks/issues/337 I found that the same bug plagueing dbt-databricks affects pysql as well. 

To reproduce this issue:

1. Add an intentionally bad `~/.netrc` to your workstation, like this:

```
machine <my-workspace>.cloud.databricks.com
login token
password <expired_token>
```

2. Try to run the `interactive_oauth.py` example using the same host name specified in the `.netrc` file.
3. You'll receive this exception:

```
    access_token = oauth_response["access_token"]
KeyError: 'access_token'
```

## The fix

Straightforward: force `requests` to not use the `.netrc` file when making requests to Databricks OAuth endpoints. These requests are **unauthenticated** (no auth header is required). The bug here is that if you include an auth header in the request to `https://****.staging.cloud.databricks.com/oidc/v1/token` the Databricks runtime will return an error response. `oauth.py` looks for `access_token` in this response and doesn't find one, so it raises an Exception.

If `.netrc` is present, `requests` always uses it. Even for these requests that are supposed to be unauthenticated. So we need to  force it to not do this.

susodapop:
## Description

This pull request applies the fix described in #121.

Closes #121

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add support for Cloud Fetch
 - Fix: Revised SQLAlchemy dialect and examples for compatibility with SQLAlchemy==1.3.x
+- Fix: oauth would fail if expired credentials appeared in ~/.netrc
 
 ## 2.7.0 (2023-06-26)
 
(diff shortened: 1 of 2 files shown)


susodapop:
All our e2e tests pass with this change.

susodapop:
I've synced this branch with `main` and updated the changelog

mattdeekay (approved):

PR #127, opened 2023-05-16
Cloud Fetch download handler

mattdeekay:
CloudFetch download handler. Each handler will run on a thread and a CloudFetch download manager will maintain a thread pool.

### DIFF
--- a/src/databricks/sql/cloudfetch/downloader.py
+++ b/src/databricks/sql/cloudfetch/downloader.py
@@ -81,20 +81,19 @@ class ResultSetDownloadHandler(threading.Thread):
             else:
                 if self.settings.is_lz4_compressed:
                     compressed_data = response.content
-                    uncompressed_data = lz4.frame.decompress(compressed_data)
+                    uncompressed_data, bytes_read = lz4.frame.decompress(compressed_data, return_bytes_read=True)
+                    if bytes_read < len(compressed_data):
+                        d_context = lz4.frame.create_decompression_context()
+                        start = 0
+                        uncompressed_data = bytearray()
+                        while start < len(compressed_data):
+                            data, num_bytes, e = lz4.frame.decompress_chunk(d_context, compressed_data[start:])
+                            uncompressed_data += data
+                            start += num_bytes
                     self.result_file = uncompressed_data
-
-                    if len(uncompressed_data) != self.result_link.bytesNum:
-                        self.is_file_downloaded_successfully = False
-                    else:
-                        self.is_file_downloaded_successfully = True
-
                 else:
                     self.result_file = response.content
-                    if len(self.result_file) != self.result_link.bytesNum:
-                        self.is_file_downloaded_successfully = False
-                    else:
-                        self.is_file_downloaded_successfully = True
+                self.is_file_downloaded_successfully = len(self.result_file) == self.result_link.bytesNum
         except:
             self.is_file_downloaded_successfully = False
 
--- a/src/databricks/sql/cloudfetch/downloader.py
+++ b/src/databricks/sql/cloudfetch/downloader.py
@@ -100,7 +100,9 @@ class ResultSetDownloadHandler(threading.Thread):
             self.is_download_finished.set()
 
     def _reset(self):
-        # Reset download-related flags for every retry of run()
+        """
+        Reset download-related flags for every retry of run()
+        """
         self.is_file_downloaded_successfully = False
         self.is_link_expired = False
         self.is_download_timedout = False
--- a/src/databricks/sql/cloudfetch/downloader.py
+++ b/src/databricks/sql/cloudfetch/downloader.py
@@ -38,9 +38,10 @@ class ResultSetDownloadHandler(threading.Thread):
             if not self.is_download_finished.wait(timeout=timeout):
                 self.is_download_timedout = True
                 logger.debug(
-                    "Cloud fetch download timed out after {} seconds for url: {}".format(
+                    "Cloud fetch download timed out after {} seconds for link representing rows {} to {}".format(
                         self.settings.download_timeout,
-                        self.result_link.fileLink,
+                        self.result_link.startRowOffset,
+                        self.result_link.startRowOffset + self.result_link.rowCount,
                     )
                 )
                 return False
--- a/src/databricks/sql/cloudfetch/downloader.py
+++ b/src/databricks/sql/cloudfetch/downloader.py
@@ -1,0 +1,104 @@
+import logging
+
+import requests
+import lz4.frame
+import threading
+import time
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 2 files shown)


src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self, downloadable_result_settings, t_spark_arrow_result_link):
+        super().__init__()
+        self.settings = downloadable_result_settings
+        self.result_link = t_spark_arrow_result_link
+        self.is_download_scheduled = False
+        self.is_download_finished = threading.Event()
+        self.is_file_downloaded_successfully = False
+        self.is_link_expired = False
+        self.is_download_timedout = False
+        self.http_code = None
```
susodapop, line 22: When we call the .run() method, these properties always set the same way. Do they need to be initialised here? Should they instead be initialised only in the run method? Or only in this method?

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -0,0 +1,103 @@
...
+                        uncompressed_data = bytearray()
+                        while start < len(compressed_data):
+                            data, num_bytes, e = lz4.frame.decompress_chunk(d_context, compressed_data[start:])
+                            uncompressed_data += data
+                            start += num_bytes
+                    self.result_file = uncompressed_data
+                else:
+                    self.result_file = response.content
+                self.is_file_downloaded_successfully = len(self.result_file) == self.result_link.bytesNum
+        except:
```
susodapop, line 97: Blind `except` blocks like this can silently hide unrelated errors. Can you catch the exceptions you expect, such as a read error or a decode error? 

At the very least, log the exception like this:

```suggestion
        except Exception as e:
            logger.error(e)
```

susodapop (commented):
Do we have integration tests for this part? Or are those in a subsequent PR?

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -0,0 +1,103 @@
...
+            self.is_file_downloaded_successfully = False
+
+        finally:
+            session.close()
```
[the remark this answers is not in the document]
susodapop: Yep, or you can one-liner this with `session and session.close()`

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -0,0 +1,103 @@
...
+import lz4.frame
+import threading
+import time
+
+logger = logging.getLogger(__name__)
+
+
+class ResultSetDownloadHandler(threading.Thread):
+
+    def __init__(self, downloadable_result_settings, t_spark_arrow_result_link):
```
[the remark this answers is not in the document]
mattdeekay: I annotated the types except for `downloadable_result_settings` since the type isn't available until the next PR in line for the Download Manager

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -0,0 +1,103 @@
...
+            proxy = {
+                "http": f"http://{self.settings.proxy_host}:{self.settings.proxy_port}",
+                "https": f"https://{self.settings.proxy_host}:{self.settings.proxy_port}",
+            }
+            session.proxies.update(proxy)
+
+            # ProxyAuthentication -> static enum BASIC and NONE
+            if self.settings.proxy_auth == "BASIC":
+                session.auth = requests.auth.HTTPBasicAuth(self.settings.proxy_uid, self.settings.proxy_pwd)
+
```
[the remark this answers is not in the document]
mattdeekay: Removing for now. I'll add support for proxies in the future

src/databricks/sql/cloudfetch/downloader.py
mattdeekay: I think they all need to be initialized because upstream, `ResultFileDownloadManager` may set some of these properties, and because of multithreading, there can be scenarios where `is_file_download_successful()` looks at properties that `run()` may not have set yet. Also, `run()` on a handler may be called multiple times depending on retries. I'll keep the initialization here but will create a `_reset()` function that resets some of these properties

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -0,0 +1,103 @@
...
+            current_time = int(time.time() * 1000)
+            if (self.result_link.expiryTime < current_time) or (
+                    self.result_link.expiryTime - current_time < (
+                    self.settings.result_file_link_expiry_buffer * 1000)
+            ):
+                self.is_link_expired = True
+                return
+
+        session = requests.Session()
+        session.timeout = self.settings.download_timeout
```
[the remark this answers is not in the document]
mattdeekay: Available downloader settings will be a part of the DownloadManager (w/ doc strings in that file). Should they be here for reference or is the former sufficient?

mattdeekay:
> Do we have integration tests for this part? Or are those in a subsequent PR?

Nope, will be part of the full e2e PR

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -19,85 +24,125 @@ def __init__(self, downloadable_result_settings, t_spark_arrow_result_link):
...
-                        return False
-                else:
-                    self.download_completion_semaphore.acquire()
-        except:
+            if not self.is_download_finished.wait(timeout=timeout):
+                self.is_download_timedout = True
+                logger.debug(
+                    "Cloud fetch download timed out after {} seconds for url: {}".format(
+                        self.settings.download_timeout,
+                        self.result_link.fileLink,
```
rcypher-databricks, line 43: I don't think we should be logging the actual URLs

mattdeekay: Oh I guess even with `debug` that may be a hazard. Will change

src/databricks/sql/cloudfetch/downloader.py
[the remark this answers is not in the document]
susodapop: Got it. We'll just need it to be annotated before it lands in `main` else it will fail the linter.

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -19,85 +24,126 @@ def __init__(self, downloadable_result_settings, t_spark_arrow_result_link):
...
 
         finally:
-            session.close()
+            session and session.close()
+            # Awaken threads waiting for this to be true which signals the run is complete
             self.is_download_finished.set()
-            self.download_completion_semaphore.release()
+
+    def _reset(self):
+        # Reset download-related flags for every retry of run()
```
susodapop, line 103: Use a docstring so that IDEs can show this info on hover.

```suggestion
        """
        Reset download-related flags for every retry of run()
        """
```

susodapop (approved):

PR #128, opened 2023-05-17
Remove unused import for SQLAlchemy 2 compatibility

WilliamGentry:
This PR removes an unused import that was preventing the `databricks-sql-python` module from being compatible with SQLAlchemy 2.

In our company's project, we have fully migrated to SQLAlchemy 2.0.4. However, the current `databricks-sql-python` module obviously requires SQLAlchemy >=1.3.0. Despite this, I found that by simply removing a single unused import, the module could be effectively used in our SQLAlchemy 2.0.4 project.

I ran the unit tests before and after removing the import. In both cases, the results were identical with 92 tests passed and 2 skipped. This makes sense since the imported `processors` object was never used anywhere. 

While the ideal solution would be to update the entire project to be compatible with SQLAlchemy 2.0, this smaller change allows users to integrate `databricks-sql-python` into their SQLAlchemy 2 projects without any immediate breaking issues, if they're willing to use the connector out-of-spec. This serves as a temporary workaround for those users until a full migration to SQLAlchemy 2 can be achieved (I'm happy to help with that)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix: Revised SQLAlchemy dialect and examples for compatibility with SQLAlchemy==1.3.x (#173)
 - Fix: oauth would fail if expired credentials appeared in ~/.netrc (#122)
 - Fix: Python HTTP proxies were broken after switch to urllib3 (#158)
+- Other: remove unused import in SQLAlchemy dialect
 - Other: Relax pandas dependency constraint to allow ^2.0.0 (#164)
 - Other: Connector now logs operation handle guids as hexadecimal instead of bytes (#170)
 - Other: test_socket_timeout_user_defined e2e test was broken (#144)
--- a/src/databricks/sqlalchemy/dialect/__init__.py
+++ b/src/databricks/sqlalchemy/dialect/__init__.py
@@ -5,7 +5,7 @@ import decimal, re, datetime
 from dateutil.parser import parse
 
 import sqlalchemy
-from sqlalchemy import types, processors, event
+from sqlalchemy import types, event
 from sqlalchemy.engine import default, Engine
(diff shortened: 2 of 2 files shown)


susodapop:
Thanks for doing the research on this! Of course this is an easy merge. Did you have any issues installing with SQLAlchemy==2.x given that we're pinned to ^1.x right now? 

I thought that if we merge this it will still fail on install because of the dependency specification in `pyproject.toml`.

susodapop (approved):
Change LGTM just wondering if we should update pyproject.toml for this as well. Even experimentally as a `.dev` release?

WilliamGentry:
Like you inferred, installing does fail for us. We're largely experimenting right now so we're working off my fork where I bumped SQLAlchemy to ^2.0 and Pandas to ^2.0.  An official or `.dev` release with loosened requirements would be great.

For the record all unit tests pass when run with the upgraded deps (as long as `processors` import removed) and so far connect library works fine with upgraded deps for our workload (mostly basic SELECT query construction).

susodapop:
Sorry to leave this hanging so long. I confirmed the tests pass and am preparing to merge. I'll cut you a dev release with the SQLAlchemy version relaxed to include sqlalchemy>=2.0 -- I suspect we have more changes needed to properly support sqla v2 but this will be a start.

susodapop:
I've forked `main` as of a few moments ago into a branch called `sqla2`. There is one commit on that branch right now which relaxes the SQLAlchemy requirement to allow version 2 and sets the project version to 2.7.1.dev3, which you can now install from Pypi with `pip install databricks-sql-connector==2.7.1.dev3`.

I note that most of our e2e tests fail because `Engine.execute()` is deprecated in sqla2. But this should give you something to experiment with 😃

PR #134, opened 2023-05-31
arrow table to pandas util api

BrainLogic:
While working with databricks-sql-python API, a question appears - how to get pandas dataframe using databricks connection and execute sql API.
At least a handy util api is necessary in order to provide conversion from arrow table to pandas dataframe, moreover it has already been implemented, I just separated the logic from fetchall API.


BrainLogic:
Databricks, what do you think, is it useful? Or do you see other ways to achieve the same?
@susodapop please share your thoughts on this

PR #137, opened 2023-06-06
Default socket timeout to 15 min

mattdeekay:
Currently, if socket timeout is not specified in the kwargs, there will be no socket timeout. As per PECO-742, a single request should not block indefinitely. If the server side does not respond within one minute, we should timeout. We'll set the timeout conservatively to 15 min

### DIFF
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -152,7 +152,11 @@ class ThriftBackend:
             ssl_context=ssl_context,
         )
 
-        timeout = 60 if kwargs.get("_socket_timeout") is None else kwargs.get("_socket_timeout")
+        timeout = (
+            60
+            if kwargs.get("_socket_timeout") is None
+            else kwargs.get("_socket_timeout")
+        )
         # setTimeout defaults to 60 seconds and is expected in ms
         self._transport.setTimeout(timeout and (float(timeout) * 1000.0))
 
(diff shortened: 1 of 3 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -152,8 +152,12 @@ def __init__(
...
         )
 
-        timeout = kwargs.get("_socket_timeout")
-        # setTimeout defaults to None (i.e. no timeout), and is expected in ms
+        timeout = (
+            60
+            if kwargs.get("_socket_timeout") is None
+            else kwargs.get("_socket_timeout")
+        )
+        # setTimeout defaults to 60 seconds and is expected in ms
```
susodapop, line 160: The `dict.get()` method in Python returns `None` if the key isn't present and accepts a default value as its second argument. So this can be simplified like so:

```suggestion
        DEFAULT_SOCKET_TIMEOUT = float(60)
        timeout = kwargs.get("_socket_timeout",  DEFAULT_SOCKET_TIMEOUT)

        # setTimeout defaults to 60 seconds and is expected in ms
```

src/databricks/sql/thrift_backend.py
```diff
@@ -152,8 +153,8 @@ def __init__(
             ssl_context=ssl_context,
         )
 
-        timeout = kwargs.get("_socket_timeout")
-        # setTimeout defaults to None (i.e. no timeout), and is expected in ms
+        timeout = kwargs.get("_socket_timeout", DEFAULT_SOCKET_TIMEOUT)
+        # setTimeout defaults to 60 seconds and is expected in ms
```
andrefurlan-db, line 157: old

tests/e2e/driver_tests.py
```diff
@@ -459,14 +459,25 @@ def test_temp_view_fetch(self):
...
         #  in non-blocking mode, since no poll is done before the read
         with self.assertRaises(OperationalError) as cm:
             with self.cursor({"_socket_timeout": 0}):
                 pass
 
         self.assertIsInstance(cm.exception.args[1], io.BlockingIOError)
 
+    @skipIf(pysql_has_version('<', '2'), 'requires pysql v2')
+    def test_socket_timeout_user_defined(self):
+        #  We expect to see a BlockingIO error when the socket is opened
```
andrefurlan-db, line 472: comment does not apply to this test

susodapop (approved):

PR #141, opened 2023-06-08
Fix: some thrift RPCs failed with BadStatusLine

### LINKED ISSUE #140: dbt errors since latest update
We're getting weird errors in DBT:

image

The errors occur in random models, different every run. 

Downgrading to 2.5.2 resolves this issue. 

Logs aren't very clear:  Databricks adapter: <class 'databricks.sql.exc.RequestError'>: Error during request to server: Remote end closed connection without response

susodapop:
## Description

Some dbt customers began seeing this error message after upgrading to connector version 2.6.0

```
INFO     databricks.sql.thrift_backend:thrift_backend.py:260 Error during request to server: {"method": "ExecuteStatement", "session-id": "b'\\x01\\xee\\x06 \\xa8\\xa5\\x14@\\x8c\\xe8\\xf3\\xfe\\xc7\\xff\\xa4\\xa4'", "query-id": null, "http-code": 200, "error-message": "", "original-exception": "('Connection aborted.', BadStatusLine('0\\r\\n'))", "no-retry-reason": "non-retryable error", "bounded-retry-delay": null, "attempt": "1/30", "elapsed-seconds": "0.00203704833984375/900.0"}
```

After some research, I found that the `BadStatusLine` error is raised by Python's `http.client` if it reads an unexpected status code from an HTTP response. The specific bytes it found `0\r\n` are the trailing bytes of a response sent with `transfer-encoding: chunked`. The `0\r\n` trailing bytes are meant to indicate to the client that "this is the last chunk".

It seems that in some cases these trailing bytes are not consumed by thrift in pysql. And since we release our HTTP connection back to the connection pool after each thrift RPC, it could happen that the next RPC would actually read-in the trailing bytes of the previous RPC while expecting to read the status code and headers of the current RPC. This raised an error.

The solution is to specifically drain the connection buffer before releasing the connection back to the pool, which is the fix implemented in this PR.

For testing purposes, I deployed this PR to pypi as version `2.6.1.dev1` so we could verify its behaviour with dbt.

Closes #140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 2.6.x (Unreleased)
 
+- Fix: http.client would raise a BadStatusLine exception in some cases
+
 ## 2.6.0 (2023-06-07)
 
 - Add support for HTTP 1.1 connections (connection pools)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "databricks-sql-connector"
-version = "2.6.1.dev1"
+version = "2.6.0"
 description = "Databricks SQL Connector for Python"
 authors = ["Databricks <databricks-sql-connector-maintainers@databricks.com>"]
 license = "Apache-2.0"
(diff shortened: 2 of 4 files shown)


andrefurlan-db (approved):
👍

PR #145, opened 2023-06-13
[ES-706907] Retry GetOperationStatus for http errors

susodapop:
Now that we use `urllib3` for our HTTP connections, `urrlib3` wraps socket timeouts in an `HTTPError` class. To maintain the existing behaviour we need to catch this directly now.


andrefurlan-db (approved):
👍

PR #146, opened 2023-06-14
Cloud Fetch download manager

mattdeekay:
Download manager for Cloud Fetch. Maintains a thread pool for `ResultFileDownloadHandler` download threads. `add_file_links()` method is called to prepare handlers, one for each cloud fetch link. `get_next_downloaded_file()` will schedule relevant handlers for download, check/wait for the download status of the next file the client needs, and returns the downloaded file.

### DIFF
--- a/src/databricks/sql/cloudfetch/download_manager.py
+++ b/src/databricks/sql/cloudfetch/download_manager.py
@@ -46,7 +46,7 @@ class DownloadedFile:
 
 class ResultFileDownloadManager:
     def __init__(self, max_download_threads: int, lz4_compressed: bool):
-        self.download_handlers = []
+        self.download_handlers: List[ResultSetDownloadHandler] = []
         self.thread_pool = ThreadPoolExecutor(max_workers=max_download_threads + 1)
         self.downloadable_result_settings = DownloadableResultSettings(lz4_compressed)
         self.fetch_need_retry = False
--- a/src/databricks/sql/cloudfetch/download_manager.py
+++ b/src/databricks/sql/cloudfetch/download_manager.py
@@ -17,13 +17,13 @@ class DownloadableResultSettings:
 
     Attributes:
         is_lz4_compressed (bool): Whether file is expected to be lz4 compressed.
-        link_expiry_buffer (int): Time in seconds to prevent download of a link before it expires. Default 0 secs.
+        link_expiry_buffer_secs (int): Time in seconds to prevent download of a link before it expires. Default 0 secs.
         download_timeout (int): Timeout for download requests. Default 60 secs.
         max_consecutive_file_download_retries (int): Number of consecutive download retries before shutting down.
     """
 
     is_lz4_compressed: bool
-    link_expiry_buffer: int = 0
+    link_expiry_buffer_secs: int = 0
     download_timeout: int = 60
     max_consecutive_file_download_retries: int = 0
 
--- a/src/databricks/sql/cloudfetch/download_manager.py
+++ b/src/databricks/sql/cloudfetch/download_manager.py
@@ -73,7 +73,7 @@ class ResultFileDownloadManager:
             if handler.is_download_scheduled:
                 continue
             try:
-                self.thread_pool.submit(handler)
+                self.thread_pool.submit(handler.run)
             except:
                 break
             handler.is_download_scheduled = True
--- a/src/databricks/sql/cloudfetch/download_manager.py
+++ b/src/databricks/sql/cloudfetch/download_manager.py
@@ -4,28 +4,10 @@ from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from typing import List, Union
 
-from databricks.sql.cloudfetch.downloader import ResultSetDownloadHandler
+from databricks.sql.cloudfetch.downloader import ResultSetDownloadHandler, DownloadableResultSettings
 from databricks.sql.thrift_api.TCLIService.ttypes import TSparkArrowResultLink
 
 logger = logging.getLogger(__name__)
-
-
-@dataclass
-class DownloadableResultSettings:
-    """
-    Class for settings common to each download handler.
-
-    Attributes:
-        is_lz4_compressed (bool): Whether file is expected to be lz4 compressed.
-        link_expiry_buffer_secs (int): Time in seconds to prevent download of a link before it expires. Default 0 secs.
-        download_timeout (int): Timeout for download requests. Default 60 secs.
-        max_consecutive_file_download_retries (int): Number of consecutive download retries before shutting down.
-    """
-
-    is_lz4_compressed: bool
-    link_expiry_buffer_secs: int = 0
-    download_timeout: int = 60
-    max_consecutive_file_download_retries: int = 0
 
 
 @dataclass
--- a/src/databricks/sql/cloudfetch/download_manager.py
+++ b/src/databricks/sql/cloudfetch/download_manager.py
@@ -165,7 +165,7 @@ class ResultFileDownloadManager:
                 self.num_consecutive_result_file_download_retries += 1
 
                 # Re-submit handler run to thread pool and recursively check download status
-                self.thread_pool.submit(handler)
+                self.thread_pool.submit(handler.run)
                 return self._check_if_download_successful(handler)
             else:
                 self.fetch_need_retry = True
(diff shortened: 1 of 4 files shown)


src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+
+        self.num_consecutive_result_file_download_retries = 0
+        self.fetch_need_retry = False
+        return True
+
+    def _stop_all_downloads_and_clear_handlers(self):
+        self.download_handlers = []
+
+
+DownloadableResultSettings = namedtuple(
```
susodapop, line 114: Not sure how much effort we need to put in here. But for understandability you can use a Python dataclass instead, that way we get type annotations for each of these members and the ability to describe them all in a docstring. Thoughts?

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+    "DownloadableResultSettings",
+    "is_lz4_compressed result_file_link_expiry_buffer download_timeout use_proxy disable_proxy_for_cloud_fetch "
+    "proxy_host proxy_port proxy_uid proxy_pwd max_consecutive_file_download_retries download_retry_wait_time"
+)
+
+DownloadedFile = namedtuple(
```
susodapop, line 120: Another candidate for a data class.

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+    def _schedule_downloads(self):
+        """
+        Schedule downloads for all download handlers if not already scheduled
+        """
+        for handler in self.download_handlers:
+            if handler.is_download_scheduled:
+                continue
+            try:
+                self.thread_pool.submit(handler.run)
+            except:
```
susodapop, line 77: Naked `except` clauses can silently hide errors. Please catch any exceptions you know about and allow the others to raise naturally.


```suggestion
            except Exception as e:
                logger.error(e)
```

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self, max_download_threads, lz4_compressed):
+        self.download_handlers = []
+        self.thread_pool = ThreadPoolExecutor(max_workers=max_download_threads + 1)
+        self.downloadable_result_settings = _get_downloadable_result_settings(lz4_compressed)
+        self.fetch_need_retry = False
+        self.num_consecutive_result_file_download_retries = 0
+        self.cloud_fetch_index = 0
+
+    def add_file_links(self, t_spark_arrow_result_links, next_row_index):
```
susodapop, line 16: Type annotations can help a lot here.

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+        for t_spark_arrow_result_link in t_spark_arrow_result_links:
```
susodapop, line 17: For clarity, I recommend just calling the loop variable `link`

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+            if t_spark_arrow_result_link.rowCount <= 0:
+                continue
+            self.download_handlers.append(ResultSetDownloadHandler(
+                self.downloadable_result_settings, t_spark_arrow_result_link))
+        self.cloud_fetch_index = next_row_index
+
+    def get_next_downloaded_file(self, next_row_index):
```
susodapop, line 24: It's probably in a separate PR, but where is this method called from?

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+            return result
+        # Download was not successful for next download item, force a retry
+        return None
+
+    def _remove_past_handlers(self, next_row_index):
+        """
+        Remove any download handlers whose start to end range doesn't include the next row to be fetched
+        i.e. no need to download
+        """
+        i = 0
```
susodapop, line 60: Isn't the point of cloudfetch to allow us to fetch results "out of order" in chunks and then re-assemble them once all downloads are complete? If I understand correctly, why do we need to expire handlers on the basis of the current row-count? Isn't that arbitrarily slowing the download process?

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+    def _check_if_download_successful(self, handler):
+        if not handler.is_file_download_successful():
+            if handler.is_link_expired:
+                self._stop_all_downloads_and_clear_handlers()
+                self.fetch_need_retry = True
+                return False
+            elif handler.is_download_timedout:
+                if self.num_consecutive_result_file_download_retries >= \
+                        self.downloadable_result_settings.max_consecutive_file_download_retries:
+                    # raise Exception("File download exceeded max retry limit")
```
susodapop, line 96: Let's not commit comments like this.

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+        return next_indices[0] if len(next_indices) > 0 else None
+
+    def _check_if_download_successful(self, handler):
+        if not handler.is_file_download_successful():
+            if handler.is_link_expired:
+                self._stop_all_downloads_and_clear_handlers()
+                self.fetch_need_retry = True
+                return False
+            elif handler.is_download_timedout:
+                if self.num_consecutive_result_file_download_retries >= \
```
susodapop, line 94: I believe either `requests` or `urllib3` (which `requests` uses under the hood) can automatically handle retries for basic HTTP requests like this. We don't need to implement it ourselves.

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -0,0 +1,139 @@
...
+            else:
+                self.fetch_need_retry = True
+                return False
+
+        self.num_consecutive_result_file_download_retries = 0
+        self.fetch_need_retry = False
+        return True
+
+    def _stop_all_downloads_and_clear_handlers(self):
+        self.download_handlers = []
```
susodapop, line 111: Is this all? It seems like this will leave the ongoing threads running until a garbage collector runs. I expect to see a call to the ThreadPoolExecutor that actually kills those threads.

susodapop (commented):
Note: to review this in my local environment I had to cherry pick these commits on top of the commits for #127.

src/databricks/sql/cloudfetch/download_manager.py
mattdeekay: It's in a separate PR, will be called from the CloudFetch queue (_create_next_table)

src/databricks/sql/cloudfetch/download_manager.py
mattdeekay: Yes, we initiate the downloads of multiple cloud fetch files in parallel, but we read these files one by one in order to fill our resulting pyarrow table. So yes, we don't really need to expire handlers since we would be going in order and won't have any "past handlers". But, it's nice to keep in case we have actually moved past the row range of a handler

src/databricks/sql/cloudfetch/download_manager.py
mattdeekay: Gotcha. For now, I'll keep this but will explore your suggestion for a future performance improvement

src/databricks/sql/cloudfetch/download_manager.py
mattdeekay: I added `thread_pool.shutdown()`

src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -12,50 +12,90 @@
...
+        is_lz4_compressed (bool): Whether file is expected to be lz4 compressed.
+        link_expiry_buffer (int): Time in seconds to prevent download of a link before it expires. Default 0 secs.
+        download_timeout (int): Timeout for download requests. Default 60 secs.
+        max_consecutive_file_download_retries (int): Number of consecutive download retries before shutting down.
+    """
+
     is_lz4_compressed: bool
-    link_expiry_buffer: float = 0.0
-    download_timeout: float = 0.0
+    link_expiry_buffer: int = 0
```
rcypher-databricks, line 26: are the expiry buffer and timeout in minutes, seconds?

mattdeekay: Seconds - I'll make this clearer

rcypher-databricks (approved):

susodapop (approved):

PR #151, opened 2023-06-21
Cloud fetch queue and integration

mattdeekay:
Changes:
1. Some static methods in ThriftBackend that we also need for `ResultSetQueue` are moved to `utils.py` to avoid circular dependency (`thrift_backend.py` <> `utils.py`)
2. `ResultSetQueue` abstract base class, with `ResultSetQueueFactory` producing `ArrowQueue` if result is arrow-based, `ArrowQueue` again if column-based, and `CloudFetchQueue` if url-based.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -17,7 +17,7 @@ from databricks.sql.experimental.oauth_persistence import OAuthPersistence
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_RESULT_BUFFER_SIZE_BYTES = 10485760
+DEFAULT_RESULT_BUFFER_SIZE_BYTES = 104857600
 DEFAULT_ARRAY_SIZE = 100000
 
 
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -839,6 +839,7 @@ class ResultSet:
             lz4_compressed=self.lz4_compressed,
             arrow_schema_bytes=self._arrow_schema_bytes,
             description=self.description,
+            max_download_threads=self.connection.max_download_threads,
         )
         self.results = results
         self.has_more_rows = has_more_rows
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -155,8 +155,6 @@ class Connection:
         # (True by default)
         # use_cloud_fetch
         # Enable use of cloud fetch to extract large query results in parallel via cloud storage
-        # max_download_threads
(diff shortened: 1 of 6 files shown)


rcypher-databricks (approved):

rcypher-databricks (approved):

src/databricks/sql/client.py
```diff
@@ -189,6 +193,8 @@ def read(self) -> Optional[OAuthToken]:
         self._session_handle = self.thrift_backend.open_session(
             session_configuration, catalog, schema
         )
+        self.use_cloud_fetch = kwargs.get("use_cloud_fetch", False)
+        self.max_download_threads = kwargs.get("max_download_threads", 10)
```
mattdeekay, line 197: With some performance testing, I did not find too much of a performance increase by allowing more threads than 10. This may vary by individual machine or cluster

src/databricks/sql/client.py
```diff
@@ -17,7 +17,7 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_RESULT_BUFFER_SIZE_BYTES = 10485760
+DEFAULT_RESULT_BUFFER_SIZE_BYTES = 104857600
```
mattdeekay, line 20: 10485760 bytes was too small for cloud fetch to work as one result link may represent more than that many bytes.

src/databricks/sql/thrift_backend.py
```diff
@@ -136,6 +139,9 @@ def __init__(
             "_use_arrow_native_timestamps", True
         )
 
+        # Cloud fetch
+        self.max_download_threads = kwargs.get("max_download_threads", 10)
```
mattdeekay, line 143: I did not see a performance improvement from setting max_download_threads to above 10. However, this may differ depending on machine/cluster.

src/databricks/sql/client.py
susodapop: Is this number arbitrary? Should we make it configurable?

susodapop (approved):
LGTM

src/databricks/sql/client.py
mattdeekay: The increase to 104857600 is somewhat arbitrary. I'll provide a follow-up PR to make this configurable

mattdeekay: The increase to 104857600 is somewhat arbitrary - x10 from the original simply has worked for cloud fetch. I'll provide a follow-up PR to make this configurable

PR #154, opened 2023-06-21
Cloud Fetch e2e tests

mattdeekay:
E2e tests comparing cloud fetch and non-cloud fetch outputs for the same query with an assortment of query row limits, number of cloud fetch downloader threads, and lz4 compression.

### DIFF
--- a/tests/e2e/test_driver.py
+++ b/tests/e2e/test_driver.py
@@ -638,11 +638,11 @@ class PySQLCoreTestSuite(SmokeTestMixin, CoreTestMixin, DecimalTestsMixin, Times
     @skipUnless(pysql_supports_arrow(), 'needs arrow support')
     def test_cloud_fetch(self):
         # This test can take several minutes to run
-        limits = [100000, 600000]
+        limits = [100000, 300000]
         threads = [10, 25]
         self.buffer_size_bytes = 104857600
         self.arraysize = 100000
-        base_query = "SELECT * FROM store_sales "
+        base_query = "SELECT * FROM store_sales WHERE ss_sold_date_sk = 2452234 "
         for num_limit, num_threads, lz4_compression in itertools.product(limits, threads, [True, False]):
             with self.subTest(num_limit=num_limit, num_threads=num_threads, lz4_compression=lz4_compression):
                 cf_result, noop_result = None, None
--- a/tests/e2e/test_driver.py
+++ b/tests/e2e/test_driver.py
@@ -1,3 +1,4 @@
+import itertools
 from contextlib import contextmanager
 from collections import OrderedDict
 import datetime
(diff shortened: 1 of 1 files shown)


rcypher-databricks (approved):

tests/e2e/test_driver.py
```diff
@@ -52,6 +53,7 @@ def __init__(self, method_name):
         # If running in local mode, just use environment variables for params.
         self.arguments = os.environ if get_args_from_env else {}
         self.arraysize = 1000
+        self.buffer_size_bytes = 10485760
```
susodapop, line 56: Does this need to be updated per this comment on your other open pull request?

https://github.com/databricks/databricks-sql-python/pull/151#discussion_r1251176373

tests/e2e/test_driver.py
```diff
@@ -633,6 +635,27 @@ def test_closing_a_closed_connection_doesnt_fail(self):
             
             self.assertTrue(expected_message_was_found, "Did not find expected log messages")
 
+    @skipUnless(pysql_supports_arrow(), 'needs arrow support')
+    def test_cloud_fetch(self):
```
susodapop, line 639: Since this is a long running test, should we put it into the large query suite?

tests/e2e/test_driver.py
```diff
@@ -633,6 +635,27 @@ def test_closing_a_closed_connection_doesnt_fail(self):
...
+        # This test can take several minutes to run
+        limits = [100000, 300000]
+        threads = [10, 25]
+        self.buffer_size_bytes = 104857600
+        self.arraysize = 100000
+        base_query = "SELECT * FROM store_sales WHERE ss_sold_date_sk = 2452234 "
```
susodapop, line 645: This appears to require the `samples` catalog. Where are you setting that in this code?

Normally we run our tests against a catalog called `main` and the schema varies from one run to the next?

susodapop (commented):
The test looks good. Just a few questions about specifics.

tests/e2e/test_driver.py
mattdeekay: That's a great suggestion, I moved it there. I was also worried about how this test may take a long time compared to the other core tests in the suite

tests/e2e/test_driver.py
mattdeekay: I think it's actually catalog `hive_metastore` that has the store_sales table. Tbh, any large table with lots of rows would do and this is the one I found. Is it okay to specify a different catalog from `main` and only test against the `default` schema (as I'm doing in the latest commit)?

tests/e2e/test_driver.py
mattdeekay: I actually modified it in L114 in this file for `test_cloud_fetch` and was keeping `self.buffer_size_bytes = 10485760` to keep it as before for the existing e2e tests. I'll update this value to the new default

tests/e2e/test_driver.py
susodapop: That should be fine, but I would document in the comments the reason for this.

susodapop (approved):
Please add a comment about what's needed for the the e2e test to run successfully. That will be important in the coming weeks when we enable automatic e2e tests with github actions.

PR #158, opened 2023-06-23
Fix proxy connection pool creation

sebbegg:
Stumbled across this while trying to use the connection from within a proxy environment.

As far as I can tell the proxy cool creation is passed the wrong host/port and was missing the `https` scheme.
Without the proper schema urllib won't do a proper `CONNECT` call and the requests later on will fail.
It appears `realhost` and `realport` in the `THttpClient` were not used anywhere, but were probably intended for the proxy use-case?

Tested this locally against our databricks subscription.

<sub>Sebastian Eckweiler (sebastian.eckweiler@mercedes-benz.com), Mercedes-Benz AG on behalf of MBition GmbH ([Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md))</sub>


danigian:
@susodapop please give the proper love to this PR.
We were having issues with latest 2.7.0 and I can confirm that when using a proxy, the library requires the changes in this PR 👍 

Thanks @sebbegg for coming up with this

susodapop:
Thanks for the feedback. Fear not, this hasn’t escaped our notice. It
should merge and be released in the next week or so.

marcgs:
Any update when this fix will be released @susodapop ? We are facing the same issue and are working with a patched version with @sebbegg 's fix ATM.

susodapop:
It will be part of the next release this week.

susodapop (approved):
LGTM thanks for this!

PR #159, opened 2023-06-24
Don't raise exception when closing a stale Thrift session

susodapop:
Prior to this change, if a context manager or a user explicitly called `Connection.close()` twice on the same connection, the second call would raise an exception: 

```
databricks.sql.exc.DatabaseError: Invalid SessionHandle: SessionHandle [<GUID>]
```

This sometimes affects dbt, airflow and others if they use connections as a context manager.

This pull request explicitly catches this error, logs it cleanly, and then continues execution. The pull request is broken into a few commits that do this:

1. Make it easy to see log messages when running pytest
2. Implement a method that converts the `bytes` session_handle we receive from Thrift server into a human readable GUID
3. **Explicitly catch the InvalidSession handle error** (this is the most important part of this PR)
4. Log when this happens so we can observe it (this shouldn't happen in the normal course of usage)
5. Update the changelog
6. Unrelated: update `test.env.example` to include the environment variables we need for running sqlalchemy e2e tests. This is just convenient. Has no impact on the connector as installed for users.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -247,7 +247,14 @@ class Connection:
         if close_cursors:
             for cursor in self._cursors:
                 cursor.close()
-        self.thrift_backend.close_session(self._session_handle)
+        try:
+            self.thrift_backend.close_session(self._session_handle)
+        except DatabaseError as e:
+            if "Invalid SessionHandle" in str(e):
+                pass
+            else:
+                raise e
+
         self.open = False
 
     def commit(self):
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -255,9 +255,12 @@ class Connection:
                 logger.warning(
                     f"Attempted to close session that was already closed: {e}"
                 )
-                pass
             else:
-                raise e
+                logger.warning(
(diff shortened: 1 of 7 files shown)


src/databricks/sql/client.py
```diff
@@ -244,7 +247,18 @@ def _close(self, close_cursors=True) -> None:
...
+            logger.info(f"Closing session {self.get_session_id_hex()}")
+            self.thrift_backend.close_session(self._session_handle)
+        except DatabaseError as e:
+            if "Invalid SessionHandle" in str(e):
+                logger.warning(
+                    f"Attempted to close session that was already closed: {e}"
+                )
+                pass
+            else:
+                raise e
```
jadewang-db, line 260: wondering if we should swallow all exceptions for close session calls.

jadewang-db (approved):

src/databricks/sql/client.py
susodapop: Good thought. I've updated the code to log all exceptions but not raise them.

susodapop:
Okay I researched why it failed. It's because the Databricks thrift server no longer raises an exception upon multiple attempts to emit `closeSession` for the same handler. I've revised my test to account for this and updated the logging.

Issue #102, opened 2023-04-13
Pandas 2.0 support


nicholasarquette:
Any idea when Pandas 2.0 will be supported? I just tried to install it in my virtual environment and there was a conflict with the databricks-sql-python package?

![image](https://user-images.githubusercontent.com/89653636/231811297-28547c41-4c14-416d-b892-f46eec4fa0b1.png)

susodapop:
This is a duplicate of #91. We'll discuss timelines for this and update that ticket within the next week.

PedroBrochado:
Hello @susodapop 👋 
Sorry to be that person....
Any idea when this will be on the roadmap?

susodapop:
Thanks for the ping. We'll have the answer today about the timeline.

susodapop:
After discussion with the team we've actually deprioritised this. We won't consider a timeline for it again for at least a few more weeks. I know that's not a great answer ☹️  It's a side-effect of a lot of things in the pipeline for both this connector and  dbt-databricks (its biggest user).

Perhaps you can help with this? Long-term we'd like to remove the pandas dependency from the connector entirely. Short-term, do you think it would be acceptable to simply update `pyproject.toml` to allow `pandas==^2.0.0`? If so, we would absolutely review a pull request implementing such a change. If it passes our test suite then there's little risk.

This wouldn't work if pandas 2.0 support involves significant syntax changes to the few places where we `import pandas as pd` and do something with it.

PedroBrochado:
@susodapop I had a quick attempt on it, all unit tests pass with `pandas@^2.0.0` (it works on my machine 😅). 
But I had to update the python version to `>=3.8.0`. Is this acceptable?

PedroBrochado:
@susodapop Any thoughts on my comment above about `python >=3.8.0`

susodapop:
Dropping support for Python 3.7 is a significant decision that we need to adequately consider how it will affect other users. I wonder if Poetry will allow us to specify an older version of pandas for Python 3.7 and a newer version of Pandas for >3.8?

cdcadman:
@PedroBrochado @susodapop, I think either of these approaches would work:

- pandas = ">=1.2.5": https://github.com/cdcadman/databricks-sql-python/commit/5d37969fe5f1adbeb6c925286630ae4b4312173a
- pandas = [
    {version = "^2.0.0", python = ">=3.8"},
    {version = "^1.2.5", python = "<3.8"}
[...]

PR #164, opened 2023-06-29
Relax pandas dependency constraint to allow ^2.0.0 (#102)

itsdani:
## What type of PR is this? 

- [x] Other

## Description

This keeps all previous versions supported, while allowing the use of `pandas >=2.0.0`.
The `<1.4.0` constraint for `python 3.7` makes package resolution much faster since `pandas >=1.4.0` doesn't support python 3.7 anyways.

## How is this tested?

- [x] Unit tests (on python `3.7.17`, `3.10.7` and `3.11.1`)

#102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 ## 2.7.x (Unreleased)
 
-- Add support for Cloud Fetch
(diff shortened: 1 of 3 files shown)


susodapop:
Review coming soon 👀

susodapop (approved):
I checked out and ran the test suite locally with `pandas==2.0.3`. e2e and unit tests pass with Python 3.11.

PR #170, opened 2023-07-08
Use hex string version of operation ID instead of bytes

### LINKED ISSUE #163: Write operation ids as hex instead of bytes
Thrift responses include operationId and sessionId guids as byte arrays. We print these in our logs and include them in ServerOperationError responses. But they aren't human readable. We can use Python's built-in UUID library to convert them into human-readable hex strings. In the case of query ids, these can be directly tied back to Databricks Query History API responses.

susodapop:
_This PR is easiest to review commit-by-commit_

This pull request modifies how we log failures for operations. Rather than printing the operationId as `bytes` we can convert it to hex, which we can cross-reference with the Query History API.

Closes #163

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add support for Cloud Fetch
 - Fix: Revised SQLAlchemy dialect and examples for compatibility with SQLAlchemy==1.3.x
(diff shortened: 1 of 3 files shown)


rcypher-databricks (approved):

susodapop:
Fixed merge conflict on the changelog.

Issue #171, opened 2023-07-10
Engine introspection is not working with SQLAlchemy==1.3.24


bkyryliuk:
Repro:
```
from sqlalchemy import create_engine
from sqlalchemy.engine.reflection import Inspector
engine = create_engine(f'databricks://token:{TOCKEN}@dropbox-prod.cloud.databricks.com:443/default?catalog=hive_metastore', connect_args={"http_path":"/sql/1.0/warehouses/{ID}"})
inspector = Inspector.from_engine(engine)
columns = inspector.get_columns(f"{catalog}.{schema}.{table}")
```

Error mgs:
```
AttributeError: 'DatabricksDialect' object has no attribute 'get_driver_connection'
```

susodapop:
Thanks for this report. I've opened a pull request that should fix this for you. You can even test it out using a dev build I just published to pypi: `pip install databricks-sql-connector==2.7.1.dev1`

You need to modify your repro to use the connection string that databricks-sql-python expects (this is addressed in the PR description), but otherwise it runs like it should.

```python
from sqlalchemy import create_engine
from sqlalchemy.engine.reflection import Inspector


host="****.cloud.databricks.com"
http_path="/sql/1.0/warehouses/****"
access_token="dapi****"
catalog="samples"
schema="nyctaxi"

engine = create_engine(f'databricks://token:{access_token}@{host}', connect_args={"http_path": http_path, "catalog": catalog, "schema": schema})
inspector = Inspector.from_engine(engine)
columns = inspector.get_columns("trips", schema="nyctaxi")
```

Note that you should include the `schema` in your calls to `Inspector.get_columns()`. I even added a smoke test to our e2e suite so we can tell if this breaks in the future.

PR #173, opened 2023-07-10
[Issue 171] Improve sqlalchemy dialect backward compatibility with 1.3.24

susodapop:
## Description

In #113 we relaxed the version requirement for SQLAlchemy. But some changes are needed to the client code to actually _work_ with version 1.3. This wasn't caught in our e2e tests initially because our test infra installs the newest supported version of SQLAlchemy -- so the tests never actually ran under `1.3.24`.

This pull request is split into easily understood commits that make sense to review one-by-one. Here are the changes needed to backport 1.3 support:

- [ ] sqla 1.4 changed how connection strings are parsed ([changelog](https://docs.sqlalchemy.org/en/20/changelog/migration_14.html#the-url-object-is-now-immutable)). To use databricks-sql-connector with version 1.3 you must include `http_path`, `catalog`, and `schema` in your extra `connect_args` _not in the connection string_. This pull request reflects that change in `examples/sqlalchemy.py` and our test suite.
- [ ] sqla 1.4 no longer automatically adds CHECK constraints for Boolean and Enum column types ([changelog](https://docs.sqlalchemy.org/en/20/changelog/migration_14.html#enum-and-boolean-datatypes-no-longer-default-to-create-constraint)). To support version 1.3 in our code we have to explicitly disable this.
- [ ] sqla 1.4 changed the dialect properties used to access the underlying dbapi connection (this is not mentioned in the changelog afaik), which we use when calling `databricks-sql-connector`'s `get_columns` and `get_tables` methods.
- [ ] sqla 1.4 changed the semantics when using the `select()` construct ([changelog](https://docs.sqlalchemy.org/en/20/changelog/migration_14.html#orm-query-is-internally-unified-with-select-update-delete-2-0-style-execution-available)). That change is not backwards compatible. So I added conditional logic in our tests and example script.
- [ ] Pandas requires sqlalchemy 1.4 so I add a skip condition for our pandas ingestion e2e test. It simply won't work for version 1.3
- [ ] Issue #171 tries to directly access the sqlalchemy `Inspector` object for reflection. I added a smoke test to ensure that this works in a basic use-case. **This pull request _does not_ implement a comprehensive suite of tests. More changes may be needed. We will expect further feedback from users on older versions of SQLAlchemy.**

Closes https://github.com/databricks/databricks-sql-python/issues/171
Follow-up for https://github.com/databricks/databricks-sql-python/pull/113

## Tests

I ran the e2e tests with SQLAlchemy 1.4.48 and 1.3.24. All tests now pass. I also ran `examples/sqlalchemy.py` under both installations to confirm it works.

To test this yourself, check out this branch and run `poetry install`. Then run `poetry run pip install sqlalchemy==1.3.24` to force the SQLAlchemy version back to 1.3. Then just run the sqlalchemy e2e tests.

You can also run `pip install databricks-sql-connector==2.7.1.dev1` and run the `examples/sqlalchemy.py` script from a fresh virtualenv to ensure that it works for you.

#### 1.3.24

CleanShot 2023-07-10 at 17 03 28@2x

#### 1.4.48

CleanShot 2023-07-10 at 16 58 58@2x

Prior to this change, version 1.3.24 would error-out immediately.

### DIFF
--- a/tests/e2e/sqlalchemy/test_basic.py
+++ b/tests/e2e/sqlalchemy/test_basic.py
@@ -13,6 +13,23 @@ except ImportError:
 
 USER_AGENT_TOKEN = "PySQL e2e Tests"
 
+def sqlalchemy_1_3():
+    import sqlalchemy
+    return sqlalchemy.__version__.startswith("1.3")
+
+def version_agnostic_select(object_to_select, *args, **kwargs):
+    """
+    SQLAlchemy==1.3.x requires arguments to select() to be a Python list
+    
+    https://docs.sqlalchemy.org/en/20/changelog/migration_14.html#orm-query-is-internally-unified-with-select-update-delete-2-0-style-execution-available
+    """
+
+    if sqlalchemy_1_3():
+        return select([object_to_select], *args, **kwargs)
+    else:
+        return select(object_to_select, *args, **kwargs)
+    
+
 
 @pytest.fixture
 def db_engine():
@@ -154,7 +171,7 @@ def test_bulk_insert_with_core(db_engine, metadata_obj, session):
     metadata_obj.create_all()
     db_engine.execute(insert(SampleTable).values(rows))
 
-    rows = db_engine.execute(select(SampleTable)).fetchall()
+    rows = db_engine.execute(version_agnostic_select(SampleTable)).fetchall()
 
     assert len(rows) == num_to_insert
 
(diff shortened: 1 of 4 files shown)


tests/e2e/sqlalchemy/test_basic.py
```diff
@@ -237,7 +238,11 @@ class SampleObject(base):
         SampleObject.name.in_(["Bim Adewunmi", "Miki Meek"])
     )
 
-    output = [i for i in session.scalars(stmt)]
+    if sqlalchemy_1_3():
```
susodapop, line 257: Whoops. This change belongs in a separate commit. Basically: version 1.3 doesn't have the `.scalars` method whereas 1.4 does.

rcypher-databricks (approved):

mattdeekay (approved):
Appreciate the detailed walkthrough with the commit messages!

PR #174, opened 2023-07-11
[Issue 115] sqlalchemy: fix has_table so it honours schema= argument

### LINKED ISSUE #115: SQLAlchemy dialect `has_table` doesn't use the schema name.
[has_table](https://github.com/databricks/databricks-sql-python/blob/398ab445efc7ad73c4c666817a0e584547e5b2fb/src/databricks/sqlalchemy/dialect/__init__.py#L275-L293) gets the schema namebut [doesn't pass it](https://github.com/databricks/databricks-sql-python/blob/398ab445efc7ad73c4c666817a0e584547e5b2fb/src/databricks/sqlalchemy/dialect/__init__.py#L285) to `DESCRIBE`.

This causes SQLAlchemy to fail to drop a given table (in a given schema).

susodapop:
## Description

Our `has_table` method was badly implemented because it didn't honour the provided `schema=` argument. I've updated the method and added an integration test. I also added a `catalog=` argument so users can run `has_table` across catalogs in UC-land.

Closes #115

~~This change is branched off of #173 so there appears to be more happening than is necessary. You only need to review the last two commits.~~ I rebased off of `main` so now there is a sensible number of commits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.7.x (Unreleased)
 
 - Add support for Cloud Fetch
+- SQLAlchemy has_table function now honours schema= argument and adds catalog= argument
 - Fix: Revised SQLAlchemy dialect and examples for compatibility with SQLAlchemy==1.3.x
 - Fix: oauth would fail if expired credentials appeared in ~/.netrc
 - Fix: Python HTTP proxies were broken after switch to urllib3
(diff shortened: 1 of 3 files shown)


susodapop:
I've confirmed this passes all sqlalchemy e2e tests on version 1.3.24 and 1.4.48

CleanShot 2023-07-10 at 20 31 45@2x

rcypher-databricks (approved):

mattdeekay (approved):

susodapop:
Synced with `main` to fix merge conflict.

PR #182, opened 2023-07-26
Use urllib3 for retries

susodapop:
## Description

This pull request implements a sub-class of `urllib3`'s `Retry` class. The goal is reliability, testability, and extensibility of our retry behaviour which prior to this change was entangled with the command execution flow in `thrift_backend.py`. This new behavior is gated behind a new config called `enable_v3_retries`.

The code is extensively commented and includes accurate docstrings. All tests pass.

### DIFF
--- a/src/databricks/sql/auth/retry.py
+++ b/src/databricks/sql/auth/retry.py
@@ -30,7 +30,7 @@ class CommandType(Enum):
         value_name_map = {i.value: i.name for i in cls}
         valid_command = value_name_map.get(value, False)
         if valid_command:
-            return getattr(cls, valid_command)
+            return getattr(cls, str(valid_command))
         else:
             return cls.OTHER
 
--- a/src/databricks/sql/auth/retry.py
+++ b/src/databricks/sql/auth/retry.py
@@ -70,18 +70,6 @@ class DatabricksRetryPolicy(Retry):
         List of integer HTTP status codes that the connector will retry, even for dangerous
         commands like ExecuteStatement. This is passed to urllib3 by extending its status_forcelist
 
-    :param _retry_start_time:
-        Float unix timestamp. Used to monitor the overall request duration across successive
-        retries. Never set this value directly. Use self.start_retry_timer() instead. Users
-        never set this value. It is set by ThriftBackend immediately before issuing a network
-        request.
-
-    :param _command_type:
-        CommandType of the current request being retried. Used to modify retry behaviour based
-        on the type of Thrift command being issued. See self.should_retry() for details. Users
-        never set this value directly. It is set by ThriftBackend immediately before issuing
-        a network request.
-
     :param urllib3_kwargs:
         Dictionary of arguments that are passed to Retry.__init__. Any setting of Retry() that
         Databricks does not override or extend may be modified here.
@@ -95,8 +83,6 @@ class DatabricksRetryPolicy(Retry):
         stop_after_attempts_duration: float,
         delay_default: float,
         force_dangerous_codes: List[int],
-        _retry_start_time: Optional[float] = None,
-        _command_type: Optional[CommandType] = None,
         urllib3_kwargs: dict = {},
     ):
         # These values do not change from one command to the next
@@ -106,10 +92,6 @@ class DatabricksRetryPolicy(Retry):
         self.stop_after_attempts_duration = stop_after_attempts_duration
         self.delay_default = delay_default
         self.force_dangerous_codes = force_dangerous_codes
-
-        # These values do change from one command to the next
-        self._retry_start_time = _retry_start_time
-        self.command_type = _command_type
 
         # the urllib3 kwargs are a mix of configuration (some of which we override)
         # and counters like `total` or `connect` which may change between successive retries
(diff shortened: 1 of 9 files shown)


jadewang-db (approved):

src/databricks/sql/auth/retry.py
```diff
@@ -0,0 +1,378 @@
...
+
+class CommandType(Enum):
+    EXECUTE_STATEMENT = "ExecuteStatement"
+    CLOSE_SESSION = "CloseSession"
+    CLOSE_OPERATION = "CloseOperation"
+    OTHER = "Other"
+
+    @classmethod
+    def get(cls, value: str):
+        value_name_map = {i.value: i.name for i in cls}
```
benc-db, line 30: This is kind of a weird way to use an enum, since you're not actually enumerating.  I'm assuming you have some place where you're taking a string value from external I/O?  If not, worth thinking about how we can use this as a true enum.  If so, we shouldn't need to build the value map on each call, so move this out and up.

src/databricks/sql/auth/retry.py
```diff
@@ -0,0 +1,378 @@
...
+        requests. This parameter is not used to retry failed network requests. We include
+        it in this class to keep all retry behaviour encapsulated in this file.
+
+    :param force_dangerous_codes:
+        List of integer HTTP status codes that the connector will retry, even for dangerous
+        commands like ExecuteStatement. This is passed to urllib3 by extending its status_forcelist
+
+    :param _retry_start_time:
+        Float unix timestamp. Used to monitor the overall request duration across successive
+        retries. Never set this value directly. Use self.start_retry_time() instead. Users
```
benc-db, line 71: if these are not set by users, best not to expose them in a public constructor.

src/databricks/sql/auth/retry.py
```diff
@@ -0,0 +1,378 @@
...
+        # Include urllib3's current state in our __init__ params
+        databricks_init_params["urllib3_kwargs"].update(**urllib3_init_params)  # type: ignore
+
+        return type(self)(
+            **databricks_init_params,  # type: ignore[arg-type]
+        )
+
+    @property
+    def command_type(self) -> Optional[CommandType]:
+        return self._command_type or None
```
benc-db, line 212: Why is this necessary?  It seems like this only evaluates the right hand of the or if its already None.

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -28,6 +30,7 @@ def __init__(
         key_file=None,
         ssl_context=None,
         max_connections: int = 1,
+        retry_policy: Union[DatabricksRetryPolicy, int] = 0,
```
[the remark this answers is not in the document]
benc-db: Could we not accomplish the same thing with an Option?

src/databricks/sql/thrift_backend.py
```diff
@@ -335,15 +368,20 @@ def attempt_request(attempt):
 
             error, error_message, retry_delay = None, None, None
             try:
+
                 # The MagicMocks in our unit tests have a `name` property instead of `__name__`.
```
benc-db, line 372: I know this isn't part of this review, but I feel like something is off when I see mention to test frameworks in production code.  What is the line below doing, and how/why does it matter what MagicMocks does?

src/databricks/sqlalchemy/dialect/__init__.py
```diff
@@ -74,7 +74,7 @@ class DatabricksDialect(default.DefaultDialect):
     driver: str = "databricks-sql-python"
     default_schema_name: str = "default"
 
-    preparer = DatabricksIdentifierPreparer
+    preparer = DatabricksIdentifierPreparer  # type: ignore
```
benc-db, line 77: why do we need an ignore directive here?

tests/e2e/common/retry_test_mixins.py
```diff
@@ -15,8 +32,10 @@ def test_client_should_not_retry_429_if_RateLimitRetry_is_0(self):
                     rows = cursor.fetchall()
                     self.assertEqual(len(rows), 1)
                     self.assertEqual(rows[0][0], 1)
-        expected = "Maximum rate of 1 requests per SECOND has been exceeded. " \
-                   "Please reduce the rate of requests and try again after 1 seconds."
+        expected = (
```
benc-db, line 35: is there anyway we can test with behavior rather than log verification?  Log verification tends to be a very brittle testing approach.

benc-db (changes requested):
Will not block if this release is very time sensitive, as most of the comments are minor python style things.

src/databricks/sql/auth/retry.py
susodapop: Fixed in 88e8a111184571f95f21b6e7f39bc5427303e541

src/databricks/sql/auth/retry.py
```diff
@@ -0,0 +1,378 @@
...
+
+    @command_type.setter
+    def command_type(self, value: CommandType):
+        self._command_type = value
+
+    @property
+    def get_operation_status_delay(self) -> float:
```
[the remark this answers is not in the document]
susodapop: Fixed in c80a5eef655f083a76605580ea710f414f7fc1c7

for context, the way this is meant to be read as "GetOperationStatus_delay", but I used snake case to be more pythonic. However to fix it I just renamed it to `delay_default` which is the name of the argument that sets it in `__init__` anyway 👍

src/databricks/sql/auth/thrift_http_client.py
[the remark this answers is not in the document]
susodapop: No. `Optional` means it can be either the specified type or `None` ([docs](https://docs.python.org/3/library/typing.html#typing.Optional)). In this case it must be either `0` (urllib3 retries nothing) or `DatabricksRetryPolicy`, so `Union` is the correct route.

src/databricks/sqlalchemy/dialect/__init__.py
susodapop: This is unrelated to retries. Just something caught by mypy that i needed to go away. Our next project is to completely overhaul the SQLAlchemy dialect. I'll address this then.

tests/e2e/common/retry_test_mixins.py
susodapop: The test you commented on will be removed in v3. We never run it today. It only appears changed here because `black` caught it in the format step.

That said, in the new tests I do perform a log verification because what we're testing is that (1) the connector stops retrying when it receives a 404 (2) The connector logs that this happened. I agree that asserting logs is brittle when the log message is coming from an upstream dependency. But since the log message we're asserting in my test down below is generated _by this connector_ it's not brittle in that way.

src/databricks/sql/auth/retry.py
susodapop: Yeah this one's weird. But I think how it sits is the right approach. 

The point here is to avoid passing around strings (fragile) and assert types.

`CommandType.get()` is only called by `thrift_backend.py` as a way to normalise the name of the command it is preparing to execute, which it communicates to the retry policy.

src/databricks/sql/auth/retry.py
susodapop: Good catch. This was a holdover from during development.

Fixed a38a2859789cf73f513aa1af70c6bf5aa676e0a6

src/databricks/sql/thrift_backend.py
susodapop: This is excellent feedback and has been on my mind. I just opened https://github.com/databricks/databricks-sql-python/pull/188 to fix this by adding `autospec` to our Mocks and cleaning up the application logic.

The code comment you found was a result of rushing a change out the door. This is good tech debt to tidy up now.

rcypher-databricks (approved):

src/databricks/sql/thrift_backend.py
susodapop: #188 merged and I rebased that change into this branch 👍

PR #186, opened 2023-08-09
Update README.md to have consistent DATABRICKS_ACCESS_TOKEN

dbarrundia-tiger:


susodapop:
I'm declining this PR because the name `DATABRICKS_TOKEN` is already consistent with our other documentation ([example](https://docs.databricks.com/en/dev-tools/auth.html)). Ultimately, the name of the environment variable doesn't matter as long as you pass its value to `client()`.

dbarrundia-tiger:
@susodapop sure feel free to decline this PR, but then your [Quickstart](https://github.com/databricks/databricks-sql-python#quickstart) is not really a Quickstart right because it wouldn't work?

You have `access_token = os.getenv("DATABRICKS_ACCESS_TOKEN")`, but that's not really consistent with:
`export DATABRICKS_TOKEN=dapi********************************`

Screenshot 2023-08-09 at 11 47 52 AM

susodapop (approved):

PR #195, opened 2023-08-11
Bump to 2.9.1

susodapop:
Explicitly pinning the urllib3 dependency is important for non-isolated execution environments (like DBR) where `urllib3` is shared across Python environments. This connector expects version ^2.0.0 and raises exceptions without it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 2.9.x (Unreleased)
 
+## 2.9.1 (2023-08-11)
+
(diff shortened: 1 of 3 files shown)


susodapop:
I've pushed a dev release of this build as `2.9.1dev1` so we can manually verify it works when installing `databricks-sql-connector` in a notebook.

susodapop:
I have manually verified that the dev release of this branch successfully executes queries on older DBR versions with a sub-2.0 version of `urllib3` installed.

benc-db (approved):

PR #197, opened 2023-08-15
Make backwards compatible with urllib3~=1.0

susodapop:
This is required to support dbt-databricks because dbt-core is pinned to `urllib3~=1.0`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ numpy = [
 sqlalchemy = "^1.3.24"
 openpyxl = "^3.0.10"
 alembic = "^1.0.11"
-urllib3 = "^2.0.0"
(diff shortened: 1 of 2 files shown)


susodapop:
I tested this manually in an environment with urllib==1.26.16 installed. All e2e and unit tests pass. I am also pushing `2.9.2dev1` to pypi so we can make sure this will work with dbt-databricks as well.

susodapop:
I've manually run the integration tests for dbt-databricks using the dev version I deployed yesterday. Everything works. This PR should be good to merge. Then we'll release version 2.9.2 of pysql and cut a new release of dbt-databricks that bumps the pysql dependency to 2.9.2.

benc-db (approved):

PR #198, opened 2023-08-16
Update thrift definitions

mattdeekay:
Simply updating thrift definitions to current as of Aug 2, 2023.
This will unlock the `TSparkParameter` field that will be used for query parameterization.


susodapop:
This was implemented in #217

PR #199, opened 2023-08-16
Convenience improvements to v3 retry logic

susodapop:
This PR does three things:
- Adds an example of how to use v3 retries
- Doesn't log out every time we run a command without v3 retries
- Updates the changelog


susodapop:
I ran our e2e tests and checked that the example code works as expected for version 2.9.1.

benc-db (approved):

PR #200, opened 2023-08-16
Named parameters support

mattdeekay:
Supports server-side named query parameterization. Client will need to call execute like this:
```
cursor.execute('SELECT :param_a FROM samples', {"param_a": "some_value"})
```
This removes client-side query injection that was a facade for query parameterization while the server-side support was still under development.


andrefurlan-db (commented):
please make sure that parameterized query support is consistent across all drivers/connectors. @rcypher-databricks needs to review and make sure it is. Specially data type handling stuff

src/databricks/sql/client.py
```diff
@@ -482,21 +482,17 @@ def _handle_staging_remove(self, presigned_url: str, headers: dict = None):
             )
 
     def execute(
-        self, operation: str, parameters: Optional[Dict[str, str]] = None
```
rcypher-databricks, line 485: If we allowed users to do this I think we should continue to support it, but instead of injecting the parameter values in the driver we should convert to the format expected by the backend and pass it on.

src/databricks/sql/utils.py
```diff
@@ -470,3 +468,33 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+        spark_params = []
+        for k, v in parameters.items():
+            sql_type, spark_value = ParamConverter.get_type_and_value(v)
+            spark_params.append(
+                TSparkParameter(name=k, type=sql_type, value=spark_value)
+            )
+        return spark_params
+
+    @staticmethod
+    def get_type_and_value(value: Any) -> (str, TSparkParameterValue):
```
rcypher-databricks, line 486: We need to have more comprehensive handling of types.  There needs to be a way for a user to specify the desired sql type of the parameter value.

susodapop:
Closing as this was implemented in #217

Issue #205, opened 2023-08-22
Unexpected argument `backoff_max` with databricks-sql-connector 2.9.2


jueonpark-dev:
Hello,

Release note says "databricks-sql-connector 2.9.2 is compatible with urllib3>=1.0.0", but it still pass `backoff_max` argument, which was introduced from urllib3 2.0.0 (ref: https://github.com/urllib3/urllib3/blob/6446fef0cf432ca035169602a1447a0d8ef53e80/src/urllib3/util/retry.py#L209)

```
    return sql.connect(
  File "/usr/local/lib/python3.10/site-packages/databricks/sql/__init__.py", line 50, in connect
    return Connection(server_hostname, http_path, access_token, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/databricks/sql/client.py", line 186, in __init__
    self.thrift_backend = ThriftBackend(
  File "/usr/local/lib/python3.10/site-packages/databricks/sql/thrift_backend.py", line 189, in __init__
    self.retry_policy = databricks.sql.auth.thrift_http_client.DatabricksRetryPolicy(
  File "/usr/local/lib/python3.10/site-packages/databricks/sql/auth/retry.py", line 132, in __init__
    super().__init__(
TypeError: Retry.__init__() got an unexpected keyword argument 'backoff_max'
```

susodapop:
Thanks for this report. Fix will be released tomorrow.

susodapop:
Fix is incorporated into this release: https://github.com/databricks/databricks-sql-python/releases/tag/v2.9.3

PR #206, opened 2023-08-23
Make backwards compatible with urllib3~=1.0 [Follow up #197] 

susodapop:
## Description

Not sure how we missed this when testing #197, but `urllib3~=1.x` does not implement `backoff_max` or `backoff_jitter` configurations. So executions would fail for users who installed `urllib31=1.x`. This PR completes the work of #197 and updates our retry test mocks to be compatible with the internal workings of the older urllib3 version.

urllib3 version 1 expects the response to include a `.msg` attribute with `.keys()` and `.items()` methods. While in real-life execution this would be an `HTTPMessage` object, the `dict()` that our mocks use for headers is already sufficient and implements those methods. `.msg` is ignored in urllib3 version 2 so the same mock definition works regardless of which version of urllib3 is installed. 

Going forward, we should have some kind of smoke test that runs our tests with both versions of urllib3 installed. A bare install of this connector always pulls urllib3 version 2, but many customers and partners will have already installed an older urllib3 -- this case isn't captured in our current testing pattern.

Resolves https://github.com/databricks/databricks-sql-python/issues/205
Resolves ES-836977

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.9.4 (Unreleased)
 
-## 2.9.3 (2023-08-23)
+## 2.9.3 (2023-08-24)
 
 - Fix: Connections failed when urllib3~=1.0.0 is installed (#206)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,22 +1,26 @@
 # Release History
 
-## 2.9.x (Unreleased)
+## 2.9.4 (Unreleased)
+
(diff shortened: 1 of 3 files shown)


benc-db (approved):

susodapop:
The python 3.7 checks are failing because our poetry github action fails with no details. Investigating.

susodapop:
Fixed the actions failure in #208. Rebased onto `main` to incorporate that fix and updated the changelog release date.

susodapop:
Checks all pass now 🚀

Issue #207, opened 2023-08-24
Drop support for Python 3.7


susodapop:
We support all non-EOL versions of Python. Python 3.7 reached EOL in June 2023.

PR #208, opened 2023-08-24
Github Actions Fix: poetry install fails for python 3.7 tests

susodapop:
The latest version of poetry released on 20 August 2023 (four days ago as of this commit) which drops support for Python 3.7, causing our github action to fail.

Until we complete #207 we need to conditionally install the last version of poetry that supports Python 3.7 (poetry==1.5.1)

### DIFF
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -29,6 +29,7 @@ jobs:
       - name: Install Poetry
         uses: snok/install-poetry@v1
         with:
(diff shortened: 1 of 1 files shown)


susodapop:
Failed publish-test is due to pypi being down for maintenance right now. Not a blocker for merging.

susodapop:
No changelog entry required as this is purely a fix to Github Actions.

susodapop:
For reference, the conditional syntax used in this PR was copped from here: https://github.com/actions/runner/issues/409#issuecomment-752775072

PR #213, opened 2023-09-01
document un-optimized behaviour of executemany

### LINKED ISSUE #196: Connections `executemany` implementation is not a prepared statement.
While [PEP249](https://peps.python.org/pep-0249/#executemany) says

> Modules are free to implement this method using multiple calls to the [.execute()](https://peps.python.org/pep-0249/#id21) method or by using array operations to have the database process the sequence as a whole in one call.

Just reading the docstring from the connection class is maximally misleading:

https://github.com/databricks/databricks-sql-python/blob/7aaa014172cb1c1e32a7776f81f6b75d66d4717a/src/databricks/sql/client.py#L527-L538

As a user, I (of course) assumed that databricks-sql-connector does something smart like e.g. batched inserts for a insert query. Instead I find that I have many thousand sequentially executed individual insert requests, one for each row.

I suggest to change the docstring to point out that behaviour!

martinitus:
Resolves #196


martinitus:
@susodapop here you go!
Thanks for considering a merge of this! :-)
Regards, Martin

susodapop:
For anyone watching this, it will merge soon. I wrote about a workaround for this inefficiency (until we can actually implement a more efficient one) here: https://github.com/databricks/databricks-sql-python/issues/239#issuecomment-1745520552

susodapop (approved):
LGTM. I ran `black` on the source files. Will follow-up with an addition to `examples/` with a workaround.

PR #216, opened 2023-09-05
[PECO-1029] Updated thrift compiler version

nithinkdb:
Due to our need to incorporate new changes into thrift, and the difficulty of recompiling an older version of thrift, we are creating this pr to be a simple version bump that switches our thrift compiler to a newer version (the default version we can install from thrift).


rcypher-databricks (approved):

yunbodeng-db:
Lack of description and no details in PECO-1029 either. Why do we need to bump the version and the compiler used to generate the definition is an older version?

src/databricks/sql/thrift_api/TCLIService/constants.py
```diff
@@ -13,57 +13,57 @@
...
-        12: "STRUCT",
-        8: "TIMESTAMP",
-        1: "TINYINT",
-        13: "UNIONTYPE",
-        18: "VARCHAR",
+    10: "ARRAY",
+    4: "BIGINT",
+    9: "BINARY",
+    0: "BOOLEAN",
+    19: "CHAR",
```
susodapop, line 50: Changes like this are fine. The significant number of changed lines are a result of different formatted output from the compiler.

src/databricks/sql/thrift_api/TCLIService/ttypes.py
```diff
@@ -2841,6 +2907,270 @@ def __ne__(self, other):
...
+     - compressionCodec
+
+    """
+
+
+    def __init__(self, compressionCodec=None,):
+        self.compressionCodec = compressionCodec
+
+    def read(self, iprot):
+        if iprot._fast_decode is not None and isinstance(iprot.trans, TTransport.CReadableTransport) and self.thrift_spec is not None:
```
susodapop, line 2922: Is there any chance that we should not be exposing these Thrift commands in this public repository? Is this new command used anywhere in the connector?

susodapop (commented):
I agree with Yunbo's feedback that we should stick with the same thrift compiler version for this change (0.16 rather than 0.17).

Please describe the nature of the change you're trying to implement by this updated definition. What specific thrift commands are you trying to implement? Can we disregard the others?

This may feel unnecessary but it's exactly the kind of documentation we'll want to see if we need to reconsider this change in the future.

src/databricks/sql/thrift_api/TCLIService/ttypes.py
nithinkdb: I don't think this one is, but this is kind of on the people who edited the original thrift right? I assume this is a feature that is hypothetically useful to a thrift server at some point.

nithinkdb:
> I agree with Yunbo's feedback that we should stick with the same thrift compiler version for this change (0.16 rather than 0.17).
> 
> Please describe the nature of the change you're trying to implement by this updated definition. What specific thrift commands are you trying to implement? Can we disregard the others?
> 
> This may feel unnecessary but it's exactly the kind of documentation we'll want to see if we need to reconsider this change in the future.

Yep, updated. We just need to add support for parameters so that we can use them.

src/databricks/sql/thrift_api/TCLIService/ttypes.py
nithinkdb: Also, what is the reason for keeping the version of our thrift compiler? I've downloaded the latest stable release; I'm not sure how to even download an older version.

susodapop (approved):
Beautiful!

PR #217, opened 2023-09-05
[PECO-1026] Add Parameterized Query support to Python

nithinkdb:


src/databricks/sql/thrift_api/TCLIService/TCLIService-remote
susodapop, line 1: Ignoring changes to this file as they are generated by thrift compiler

src/databricks/sql/thrift_api/TCLIService/TCLIService.py
susodapop, line 1: Ignoring changes to this file as they are generated by thrift compiler

src/databricks/sql/thrift_api/TCLIService/constants.py
susodapop, line 1: Ignoring changes to this file as they are generated by thrift compiler

src/databricks/sql/thrift_api/TCLIService/ttypes.py
susodapop, line 1: Ignoring changes to this file as they are generated by thrift compiler

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+        self.name = name
+        self.value = value
+        self.type = type
+        
+        
+def named_parameters_to_dbsqlparams(parameters: Any):
+    dbsqlparams = []
+    if(type(parameters) is dict):
+        # Support legacy impl
+        for name,parameter in parameters.items():
```
susodapop, line 506: Nit: can you run `black` over this code?

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+    type: DbSqlType
+    def __init__(self,name="",value=None,type=None):
+        self.name = name
+        self.value = value
+        self.type = type
+        
+        
+def named_parameters_to_dbsqlparams(parameters: Any):
+    dbsqlparams = []
+    if(type(parameters) is dict):
```
susodapop, line 504: The Pythonic way to write this:
```suggestion
    if isinstance(parameters, dict):
```

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+        # Support legacy impl
+        for name,parameter in parameters.items():
+            dbsqlparams.append(DbSqlParameter(name=name,value=parameter))
+    else:
+        for parameter in parameters:
+            if(type(parameter) is DbSqlParameter):
```
susodapop, line 510: ditto
```suggestion
            if isinstance(parameter, DbSqlParameter):
```

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+                dbsqlparams.append(parameter)
+            else:
+                dbsqlparams.append(DbSqlParameter(value=parameter))
```
susodapop, line 513: I presume this branch of the else condition for line 504 is the case where parameters are passed as a `list`. In which case it seems like the parameter isn't named. In keeping with the idea that a function should do one thing, does this belong in a separate function?

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+    name: str
+    value: Any
+    type: DbSqlType
+    def __init__(self,name="",value=None,type=None):
+        self.name = name
+        self.value = value
+        self.type = type
+        
+        
+def named_parameters_to_dbsqlparams(parameters: Any):
```
susodapop, line 502: This function seems to handle both the case of named parameters in a dictionary and position parameters in a list. We should 

a) split this logic into separate functions (preferred)
b) rename the function

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+            if(type(param.value) is str):
+                param.type = DbSqlType.STRING
+            elif(type(param.value) is int):
+                param.type = DbSqlType.INTEGER
+            elif(type(param.value) is float):
+                param.type = DbSqlType.FLOAT
+            elif(type(param.value) is datetime.datetime):
+                param.type = DbSqlType.TIMESTAMP
+            elif(type(param.value) is bool):
+                param.type = DbSqlType.BOOLEAN
```
susodapop, line 528: If we keep these, all of these should be calls to `isinstance`.

Can we replace this long chain of conditions with a dictionary of `Type: DbsqlType`? Then all these conditions would be encapsulated in the following:

```suggestion
                param.type = type_lookup(type(param.value), $some_default_value)
```

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+        if(param.value):
```
susodapop, line 529: `if` is not a function in python, just a keyword. This code _works_ because the parentheses create a tuple which is still truthy or falsey. But the syntax isn't pythonic.

```suggestion
        if param.value:
```

It's also not clear what the failure modes are for this code. What if `param.value` is `None`? What if `param.value` evaluates falsey such as the integer `0` or the singleton `False`?

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+            dbsqlparams.append(DbSqlParameter(name=name,value=parameter))
+    else:
+        for parameter in parameters:
+            if(type(parameter) is DbSqlParameter):
+                dbsqlparams.append(parameter)
+            else:
+                dbsqlparams.append(DbSqlParameter(value=parameter))
+    return dbsqlparams
+
+def infer_types(params: list[DbSqlParameter]):
```
susodapop, line 516: We should prefer [pure functions](https://pythoninformer.com/programming-techniques/functional-programming/pure-functions/) wherever possible. `infer_types` isn't pure right now which makes it harder to test and therefore more likely to produce bugs.

It should be rewritten to keep its inputs immutable (don't modify their values in place) and return a new list.

src/databricks/sql/utils.py
```diff
@@ -470,3 +474,66 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+                param.type = DbSqlType.TIMESTAMP
+            elif(type(param.value) is bool):
+                param.type = DbSqlType.BOOLEAN
+        if(param.value):
+            param.value = str(param.value)
+
+def named_parameters_to_tsparkparams(parameters: Any):
+    tspark_params = []
+    dbsql_params = named_parameters_to_dbsqlparams(parameters)
+    infer_types(dbsql_params)
```
susodapop, line 535: Rewrite `infer_types` to return a new list.

```suggestion
    inferred_type_parameters = infer_types(dbsql_params)
```

tests/unit/test_parameters.py
```diff
@@ -0,0 +1,16 @@
+from databricks.sql.utils import named_parameters_to_tsparkparams
+from databricks.sql.thrift_api.TCLIService.ttypes import (
+    TSparkParameter,
+    TSparkParameterValue
+)
+from databricks.sql.utils import DbSqlParameter, DbSqlType
+
+class TestTSparkParameterConversion(object):
```
susodapop, line 8: We need to split this into some smaller tests that isolate the different behaviours. The test is currently hard to read and is also missing a test check for type inference of a `None` value.

Ideally we'd like to test `infer_types` directly rather than the higher-up function `named_parameters_to_tsparkparams` so we can more directly verify each step of the way.

susodapop (changes requested):

tests/unit/test_parameters.py
nithinkdb: I think that the sql void type is unsupported.

susodapop: Unsupported by what? VOID is a valid data type on [Databricks](https://docs.databricks.com/en/sql/language-manual/sql-ref-datatypes.html).

Even if it's not supported for parameterized execution, we need to capture the connector behaviour when a user passes a Python `None` as a parameter value. At minimum an exception should be raised and we should have a test to prove that this happens.

nithinkdb: Ok, we can definitely raise an exception. Just wanted to point out that across all connectors we are not supporting the void type for parameterized exec.

src/databricks/sql/utils.py
nithinkdb: ya, this is not a useful check, I've changed it to just stringify

tests/unit/test_parameters.py
nithinkdb: This now should raise a value exception if none is passed in.

src/databricks/sql/utils.py
nithinkdb: Well, the parameter might still be named. What we want the users to do is to use DBSQLParameters to pass in named parameters. That way they can encapsulate type as well.

nithinkdb: but ya, not really a way to decouple these: we need to support both the legacy and the new style.

src/databricks/sql/thrift_backend.py
```diff
@@ -614,7 +613,10 @@ def _create_arrow_table(self, t_row_set, lz4_compressed, schema_bytes, descripti
                 num_rows,
             ) = convert_column_based_set_to_arrow_table(t_row_set.columns, description)
         elif t_row_set.arrowBatches is not None:
-            (arrow_table, num_rows,) = convert_arrow_based_set_to_arrow_table(
+            (
+                arrow_table,
+                num_rows,
```
susodapop, line 618: Python doesn't require these parentheses and they affect readability. Why not?

```python
arrow_table, num_rows = convert_arrow_based_set_to_arrow_table( ... )
```

src/databricks/sql/utils.py
susodapop: Will need to speak with you privately about this. My original feedback still stands. Nothing about this function resists splitting it into separate, testable methods.

src/databricks/sql/utils.py
```diff
@@ -470,3 +475,81 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+    def __init__(self, name="", value=None, type=None):
+        self.name = name
+        self.value = value
+        self.type = type
+
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and self.__dict__ == other.__dict__
+
+
+def named_parameters_to_dbsqlparams(parameters: Any):
```
susodapop, line 509: At the very least we should update the type annotation to `Union[Dict, List]`. But ideally we'll split these properly with good type annotations throughout.

susodapop (commented):
Looking much better. Let's discuss splitting out `named_parameters_to_dbsqlparams`.

src/databricks/sql/thrift_backend.py
nithinkdb: this was just the result of running black on the whole file. Not sure, but if we standardize to a linter, we should use it on everything

rcypher-databricks (approved):

susodapop:
I'm seeing multiple of the sqlalchemy e2e tests failing following this merge.

susodapop:
An example of the failure:

```
self = <databricks.sql.thrift_backend.ThriftBackend object at 0x105aed110>
op_handle = TOperationHandle(operationId=THandleIdentifier(guid=b'\x01\xeeY\x80]\xea\x1bf\x98\x97\xf2oI\xc9Ee', secret=b'3\x8dR\x9d\x82rF\xeb\x84\x82\xcbA\x94f\x83\x9d', executionVersion=0), operationType=0, hasResultSet=True, modifiedRowCount=None)
get_operations_resp = TGetOperationStatusResp(status=TStatus(statusCode=0, infoMessages=None, sqlState=None, errorCode=None, errorMessage=No...118)\n\t... 3 more\n", responseValidation=None, idempotencyType=0, statementTimeout=172800, statementTimeoutLevel=None)

    def _check_command_not_in_error_or_closed_state(
        self, op_handle, get_operations_resp
    ):
        if get_operations_resp.operationState == ttypes.TOperationState.ERROR_STATE:
            if get_operations_resp.displayMessage:
>               raise ServerOperationError(
                    get_operations_resp.displayMessage,
                    {
                        "operation-id": op_handle
                        and self.guid_to_hex_id(op_handle.operationId.guid),
                        "diagnostic-info": get_operations_resp.diagnosticInfo,
                    },
                )
E               sqlalchemy.exc.DatabaseError: (databricks.sql.exc.ServerOperationError) 
E               [PARSE_SYNTAX_ERROR] Syntax error at or near '%'.(line 1, pos 71)
E               
E               == SQL ==
E               INSERT INTO `PySQLTest_1695426231` (name, episodes, some_bool) VALUES (%(name)s, %(episodes)s, %(some_bool)s)
E               -----------------------------------------------------------------------^^^
E               
E               [SQL: INSERT INTO `PySQLTest_1695426231` (name, episodes, some_bool) VALUES (%(name)s, %(episodes)s, %(some_bool)s)]
E               [parameters: ({'name': 'Bim Adewunmi', 'episodes': 6, 'some_bool': 1}, {'name': 'Miki Meek', 'episodes': 12, 'some_bool': 0})]
E               (Background on this error at: https://sqlalche.me/e/14/4xp6)
```

I'm wondering if this PR supports named parameters or not?  We may need to update our sqlalchemy dialect to not send named parameters.

PR #220, opened 2023-09-11
[PECO-1055] Updated thrift defs to allow Tsparkparameters

nithinkdb:
https://github.com/databricks/runtime/commit/7aac05c4faf39bf66ab159ec2e7942721990fecf
Updates the thrift file to reflect the changes from this commit, which enables parameters. Necessary for the parameterized query PR on python.


susodapop (approved):
Stamp. This is all auto-generated so nothing we need to modify by hand or review.

PR #225, opened 2023-09-18
Fix inaccurate changelog entry: _enable_v3_retries

### LINKED ISSUE #219: Changelog shows incorrect parameter name for specifying retry policy
A small but annoying issue:

I was facing a slew of warnings of the form `DatabricksRetryPolicy is currently bypassed. The CommandType cannot be set.`. In my search for a way to resolve this warning, I found the following helpful line in the [changelong](https://github.com/databricks/databricks-sql-python/blob/main/CHANGELOG.md?plain=1#L21C68-L21C68) (line 21):

> Replace retry handling with DatabricksRetryPolicy. This is disabled by default. To enable, set `enable_v3_retries=True` when creating `databricks.sql.client` (#182)

It claims that setting `enable_v3_retries=True` when creating an sql connection will enable the new retry mechanism. I hoped that this would silence the warning. I added the argument, but the warnings persisted. By digging through the code, I found that the correct form of the argument to set is **`_enable_v3_retries=True` with an underscore at the beginning, not `enable_v3_retries=True`**.

After making this change, the warnings have been silenced.

i.e., creating the sql connection this way properly sets the retry policy and silences the warnings:
```
sql.connect(
    server_hostname=hostname,
    http_path=http_path,
    access_token=token,
    _enable_v3_retries=True,
)
```

susodapop:
Closes #219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ __Note: this release was yanked from Pypi on 13 September 2023 due to compatibil
 
 ## 2.9.0 (2023-08-10)
 
-- Replace retry handling with DatabricksRetryPolicy. This is disabled by default. To enable, set `enable_v3_retries=True` when creating `databricks.sql.client` (#182)
+- Replace retry handling with DatabricksRetryPolicy. This is disabled by default. To enable, set `_enable_v3_retries=True` when creating `databricks.sql.client` (#182)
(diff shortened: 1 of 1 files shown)


susodapop:
Skipping tests here because this is merely a fix to the CHANGELOG

PR #227, opened 2023-09-22
Parameterized queries: Add e2e tests for parameterized queries

susodapop:
## Description

This is a follow-up to #217 adding three e2e tests. This loosely follows the design of a similar test in our nodejs connector: https://github.com/databricks/databricks-sql-nodejs/blob/main/tests/e2e/query_parameters.test.js

The idea of these tests is to demonstrate the query syntax and how to actually pass parameter values through the Python API.

Next I'll open a PR adding a file to the `/examples` directory.

### DIFF
--- a/tests/e2e/common/parameterized_query_tests.py
+++ b/tests/e2e/common/parameterized_query_tests.py
@@ -1,0 +1,185 @@
+import pytest
+import databricks.sql as sql
+from databricks.sql import Error
+import pytz
+
+from decimal import Decimal
+
+from databricks.sql.utils import DbSqlParameter, DbSqlType
+from datetime import datetime
+
+from databricks.sql.client import Connection
+
+
+class PySQLParameterizedQueryTestSuiteMixin:
+    """Namespace for tests of server-side parameterized queries"""
+
+    def test_parameterized_query_named_and_inferred_e2e(self):
+        """Verify that named parameters passed to the database as a dict are returned of the correct type
+        All types are inferred.
+        """
+
+        conn: Connection
+
+        query = """
+        SELECT
+            :p_bool AS col_bool,
+            :p_int AS col_int,
+            :p_double AS col_double,
+            :p_date as col_date,
+            :p_timestamp as col_timestamp,
+            :p_str AS col_str
+        """
+
+        named_parameters = {
+            "p_bool": True,
+            "p_int": 1234,
+            "p_double": 3.14,
+            "p_date": datetime(2023, 9, 6, 3, 14, 27, 843, tzinfo=pytz.UTC),
+            "p_timestamp": datetime(2023, 9, 6, 3, 14, 27, 843, tzinfo=pytz.UTC),
+            "p_str": "Hello",
+        }
+        with self.connection() as conn:
+            cursor = conn.cursor()
+            cursor.execute(query, parameters=named_parameters)
+            result = cursor.fetchone()
+
+            assert result.col_bool == True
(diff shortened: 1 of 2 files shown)


nithinkdb (approved):
LGTM!

src/databricks/sql/thrift_backend.py
```diff
@@ -613,10 +613,7 @@ def _create_arrow_table(self, t_row_set, lz4_compressed, schema_bytes, descripti
                 num_rows,
             ) = convert_column_based_set_to_arrow_table(t_row_set.columns, description)
         elif t_row_set.arrowBatches is not None:
-            (
-                arrow_table,
-                num_rows,
-            ) = convert_arrow_based_set_to_arrow_table(
+            (arrow_table, num_rows,) = convert_arrow_based_set_to_arrow_table(
```
susodapop, line 616: This is just a formatting change unrelated to the bulk of the PR.

src/databricks/sqlalchemy/dialect/requirements.py
```diff
@@ -24,9 +24,11 @@ def __some_example_requirement(self):
 import sqlalchemy.testing.exclusions
 
 import logging
+
```
susodapop, line 27: This is a formatting change unrelated to the rest of this PR.

tests/e2e/common/parameterized_query_tests.py
```diff
@@ -0,0 +1,181 @@
...
+import pytz
+
+from databricks.sql.client import Connection
+from databricks.sql.utils import DbSqlParameter, DbSqlType
+
+
+class PySQLParameterizedQueryTestSuiteMixin:
+    """Namespace for tests of server-side parameterized queries"""
+
+    def test_parameterized_query_named_and_inferred_e2e(self):
```
benc-db, line 13: A minor point of testing style that you can choose to ignore (as here is not really the worst violation I've seen on this point): unless the setup is very expensive (in which case consider a shared setup), a test should validate one thing, so that if the test fails, you know immediately where it failed (if not why).  Here we are bundling all of the column types together; this is fine if we're pretty confident that the asserts will either all fail or all pass.  If this is not the case, might be better to test each type individually.

susodapop: Agree that will make this easier to evaluate and extend in the future. Will ping you for a re-review once it's done.

tests/e2e/common/parameterized_query_tests.py
```diff
@@ -0,0 +1,181 @@
...
+
+        named_parameters = {
+            "p_bool": True,
+            "p_int": 1234,
+            "p_double": 3.14,
+            "p_date": datetime(2023, 9, 6, 3, 14, 27, 843, tzinfo=pytz.UTC).date(),
+            "p_timestamp": datetime(2023, 9, 6, 3, 14, 27, 843, tzinfo=pytz.UTC),
+            "p_str": "Hello",
+        }
+        with self.connection() as conn:
```
benc-db, line 38: I could be mistaken, but this block looks duplicated below.  Good idea to DRY it up, so if you need to make changes in future, you can make them in one place.

benc-db (approved):
Approved with suggestions.

tests/e2e/common/parameterized_query_tests.py
susodapop: Done in a8841f884b8576561961cf69ec356ecf15ca6752

benc-db:
Nicely done.

PR #228, opened 2023-09-23
[PECO-1109] Parameterized Query: add suport for inferring decimal types

susodapop:
## Description

This connector has strong support for Python Decimal types. But the parameterized query implementation merged last week is missing the ability to infer them. This pull request implements type inference for decimals with e2e tests for four different scenarios:

- A parameter is set equal to a primitive `Decimal()` object (type must be inferred)
- A `Decimal()` parameter value is wrapped in `DbsqlParameter` with the `type` set to `None` (type must be inferred)
- A `Decimal()` parameter value is wrapped in `DbsqlParameter` and the type is set to `DbsqlType.DECIMAL` (type is not inferred but cast expression must be calculated)
- A `Decimal()` parameter value is wrapped in `DbsqlParameter` and the type is set to a custom implementation (type is not inferred and cast expression is accepted from the user as-is)

In the first three scenarios, pysql will inspect the passed Decimal and provide the appropriate Databricks SQL cast expression to contain that decimal (such as `DECIMAL(18,2)` or `DECIMAL(38,5)`)

I also split out the `infer_types` tests so we can make special assertions about decimals and ran `isort` over a couple of files.

I'll open a subsequent PR with a full example of how users can work with these APIs.

### DIFF
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -478,11 +478,15 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
 
 
 class DbSqlType(Enum):
+    """The values of this enumeration are passed as literals to be used in a CAST
+    evaluation by the thrift server.
+    """
+
     STRING = "STRING"
     DATE = "DATE"
     TIMESTAMP = "TIMESTAMP"
     FLOAT = "FLOAT"
-    DECIMAL = "DECIMAL"
+    DECIMAL = "DECIMAL(6,2)"
     INTEGER = "INTEGER"
     BIGINT = "BIGINT"
     SMALLINT = "SMALLINT"
@@ -495,7 +499,7 @@ class DbSqlType(Enum):
 class DbSqlParameter:
     name: str
     value: Any
-    type: DbSqlType
+    type: Union[DbSqlType, Enum]
 
     def __init__(self, name="", value=None, type=None):
         self.name = name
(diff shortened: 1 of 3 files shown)


src/databricks/sql/utils.py
```diff
@@ -478,11 +478,15 @@ def _create_arrow_array(t_col_value_wrapper, arrow_type):
...
+    """The values of this enumeration are passed as literals to be used in a CAST
+    evaluation by the thrift server.
+    """
+
     STRING = "STRING"
     DATE = "DATE"
     TIMESTAMP = "TIMESTAMP"
     FLOAT = "FLOAT"
-    DECIMAL = "DECIMAL"
+    DECIMAL = "DECIMAL(6,2)"
```
susodapop, line 489: If we keep this as just `DECIMAL` then DBR will create a decimal with zero precision after the decimal point because in Databricks SQL

```sql
SELECT CAST("1234.56" as DECIMAL)
```

returns `1235`

susodapop: As of 1565700 I've implemented dynamic casting and removed the `DECIMAL(6,2)` default

benc-db (approved):
LGTM

nithinkdb (approved):
LGTM!

susodapop:
The python 3.8 lint check is hanging. Merging without this check passing as it passes for numerous other versions and appears to just be flaky.

susodapop:
Also, we need to update the changelog to discuss the new parameterisation features. But we can punt this to a subsequent PR as there is more work to do before this releases.

PR #229, opened 2023-09-26
[PECO-1083] Updated thrift files and added check for protocol version

nithinkdb:
Since we've migrated to a new version of runtime that supports parameterized queries, we need to add a protocol check to ensure that the server side is up to date in supporting such queries. Hence, we're updating the thrift files to access the new protocol, and adding a version check when doing parameterized queries.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -227,7 +227,9 @@ class Connection:
         return self.thrift_backend.handle_to_id(self._session_handle)
 
     def get_session_protocol_version(self):
-        return self.thrift_backend.extract_protocol_version_from_handle(self._session_handle)
+        return self.thrift_backend.extract_protocol_version_from_handle(
+            self._session_handle
+        )
 
     def get_session_id_hex(self):
         return self.thrift_backend.handle_to_hex_id(self._session_handle)
@@ -509,7 +511,9 @@ class Cursor:
             self.connection.get_session_protocol_version()
             < ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V8
         ):
-            raise Error("Parameterized operations are not supported by this server. Support will begin with server version DBR 14.1")
+            raise Error(
+                "Parameterized operations are not supported by this server. Support will begin with server version DBR 14.1"
+            )
         else:
             parameters = named_parameters_to_tsparkparams(parameters)
 
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -227,7 +227,7 @@ class Connection:
         return self.thrift_backend.handle_to_id(self._session_handle)
 
     def get_session_protocol_version(self):
-        return self.thrift_backend.handle_to_protocol_version(self._session_handle)
+        return self.thrift_backend.extract_protocol_version_from_handle(self._session_handle)
 
     def get_session_id_hex(self):
         return self.thrift_backend.handle_to_hex_id(self._session_handle)
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -509,7 +509,7 @@ class Cursor:
             self.connection.get_session_protocol_version()
             < ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V8
         ):
-            raise Error("Parameterized operations are not supported by this server")
+            raise Error("Parameterized operations are not supported by this server. Support will begin with server version DBR 14.1")
         else:
             parameters = named_parameters_to_tsparkparams(parameters)
 
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -5,6 +5,7 @@ import pyarrow
 import requests
 import json
 import os
+from databricks.sql.thrift_api.TCLIService import ttypes
 
 from databricks.sql import __version__
 from databricks.sql import *
@@ -224,6 +225,9 @@ class Connection:
 
     def get_session_id(self):
         return self.thrift_backend.handle_to_id(self._session_handle)
+
+    def get_session_protocol_version(self):
+        return self.thrift_backend.handle_to_protocol_version(self._session_handle)
 
     def get_session_id_hex(self):
         return self.thrift_backend.handle_to_hex_id(self._session_handle)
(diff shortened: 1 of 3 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -1003,6 +1003,10 @@ def cancel_command(self, active_op_handle):
     def handle_to_id(session_handle):
         return session_handle.sessionId.guid
 
+    @staticmethod
+    def handle_to_protocol_version(session_handle):
```
susodapop, line 1007: Why is this method named this way? The other `handle_to...` methods are converting a session handle from bytes to hex. This method is actually extracting the protocol version, right?

src/databricks/sql/client.py
```diff
@@ -501,6 +505,11 @@ def execute(
         """
         if parameters is None:
             parameters = []
+        elif (
+            self.connection.get_session_protocol_version()
+            < ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V8
+        ):
+            raise Error("Parameterized operations are not supported by this server")
```
susodapop, line 512: Can we include the protocol version in the exception? Potentially with a reference to the earliest version of DBR that should support parameters? Just something to help users discern what they can do to resolve this without digging into the source code.

susodapop (commented):
Can you please fill out the PR description?

I don't understand how `ttypes` changed without also changing TCLIService.py. Can you confirm that this is expected?

src/databricks/sql/client.py
susodapop: Case in point: a user of the node driver ran into this and opened an issue https://github.com/databricks/databricks-sql-nodejs/issues/184

It would be nice to communicate the more about the nature of this across all our drivers.

nithinkdb: fair enough. I can edit them to note that DBR version 14.1 is the minimum version.

src/databricks/sql/thrift_backend.py
yunbodeng-db: @nithinkdb please provide PR description. It's going to help your reviewers to review your PR effectively and quickly.
The PECO ticket itself does not have any description either so this is not good practice.

src/databricks/sql/client.py
kravets-levko: @nithinkdb can we do the same for Nodejs driver?

nithinkdb: Yep.

src/databricks/sql/thrift_backend.py
nithinkdb: Updated the PR description. Not sure I understand, Jesse. For example handle_to_id just accesses the member variable of sessionId.guid. This is an analogous case no? This is mainly just a good place to put this utility function.

susodapop: I agree it's a good place for it. I think the name is unclear. Elsewhere you call this method `get_session_protocol_version` which is clear ("use the session handle to get the protocol version"). That would be a good / clear name for this method too. Another alternative would be `extract_protocol_version_from_handle`.

nithinkdb:
> Can you please fill out the PR description?
> 
> I don't understand how `ttypes` changed without also changing TCLIService.py. Can you confirm that this is expected?

I am not sure how to verify this, but I've double checked my thrift version, and I used this version of the thrift file https://github.com/databricks/runtime/blob/5606e6cf37038e9fa52aa49c590c5c746688514f/sql/hive-thriftserver/if/TCLIService.thrift
It does make sense that the thrift definitions that have been updated in between what we did before and what we have done now mostly consist of adding new fields, which would only be recorded in ttypes right?

src/databricks/sql/thrift_backend.py
nithinkdb: I disagree.

def get_session_id(self):
        return self.thrift_backend.handle_to_id(self._session_handle)

   def get_session_protocol_version(self):
        return self.thrift_backend.handle_to_protocol_version(self._session_handle)

    def get_session_id_hex(self):
        return self.thrift_backend.handle_to_hex_id(self._session_handle)
      
We currently follow this convention for all of these different functions, why break this convention now? Each of these takes a handle and returns a sub value of it. I'd be fine if we decide to refactor all of these, but i don't think it makes sense to have one outlier.

susodapop: We're not disagreeing here, my friend. Look at the name of the method I'm commenting on: `handle_to_protocol_version`. I'm suggesting changing this name to `get_session_protocol_version` to match what you call it elsewhere (and what similar methods are named) because that indicates what it actually does.

susodapop: A session handle has a name, which is a GUID. That GUID can be expressed in Python `bytes` or as a hexadecimal string. It makes sense to name a method `handle_to_hex_str` because this method clearly indicates that the handle is just being converted to hex.

But `handle_to_protocol_version` doesn't make sense because the handle name doesn't include a protocol version. Rather the session handle object includes a property from which we can fetch the protocol version. So the name makes sense to reflect this. Speaking semantically: the hex string == the byte string. But the hex string != the protocol version.

nithinkdb: I see.

src/databricks/sql/client.py
```diff
@@ -509,7 +511,9 @@ def execute(
             self.connection.get_session_protocol_version()
             < ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V8
         ):
-            raise Error("Parameterized operations are not supported by this server. Support will begin with server version DBR 14.1")
+            raise Error(
+                "Parameterized operations are not supported by this server. Support will begin with server version DBR 14.1"
```
susodapop, line 515:
```suggestion
                "Parameterized operations are not supported by this server. DBR 14.1 is required."
```

susodapop (approved):

susodapop:
Hey there, did you run the e2e tests for this before merging? They all fail in the case where DBR doesn't return `serverProtocolVersion`, which is the case on the internal test instances we're using today. I suggest we revert and implement with a solution that doesn't cause _every single connection to fail_, since not all connections will be used for passing parameters.

PR #231, opened 2023-09-26
SQLAlchemy 2: reorganise dialect files into a single directory

susodapop:
The SQLAlchemy dialect authors documentation suggests this file layout. This PR does the following:

- Move all SQLAlchemy implementation, including their test suite into `src/databricks/sqlalchemy`
- Therefore moves the setup.cfg file out of our root directory as it's _only_ required by SQLAlchemy's custom pytest plugin
- Removes a pytest directive in `pyproject.toml` that was only relevant for testing SQLAlchemy

Subsequent PR's will resolve failing tests and further normalise our implementation consistent with SQLAlchemy's guidance for version 2.


susodapop:
More work is needed on parameterised queries as we're now seeing more failures in the SQLAlchemy test suite that look like this:

```
ValueError: Parameter type cannot be inferred
```

susodapop:
I've verified that our e2e tests (for both SQLAlchemy and the main driver) all pass for this change.

tests/e2e/sqlalchemy/test_basic.py
```diff
@@ -181,7 +183,7 @@ def test_bulk_insert_with_core(db_engine, metadata_obj, session):
     )
 
     rows = [
-        {"name": names[i % 3], "number": random.choice(range(10000))}
+        {"name": names[i % 3], "number": random.choice(range(64))}
```
susodapop, line 186: This is necessary because DBSQL can only accept 255 parameters per execution, right now.

benc-db (approved):

susodapop:
mypy check was failing because of this bug https://github.com/python/mypy/issues/13627

So I'm bumping the mypy dependency to 0.981 which incorporates the fix.

PR #233, opened 2023-09-28
Query parameters: implement support for binding NoneType parameters

susodapop:
## Description

This PR adds support and test for inferring `NoneType` parameter values passed to DBR as parameters. Prior to this, binding a `None` as a parameter was impossible. This includes both unit and e2e tests. Should be an easy review.

<details><summary>The original PR / issue description appears below</summary>

### Background

Trying to run the SQLAlchemy test suite after query parameterisation merged, we can’t bind a parameter with a value of NoneType because the connector can’t infer what column type it should be.

Of course, any nullable column type can accept a NULL. I don’t see in in the specification how to handle this but my guess is that I can just use a STRING column and pass a null value.

After some experimentation in the DBSQL web UI I found that CASTing a NULL into any other type still returns a NULL of that type. For example:

```sql
SELECT CAST(NULL as STRING)
SELECT CAST(NULL as BIGINT)
SELECT CAST(NULL as DATE)
SELECT CAST(NULL as TIMESTAMP)
SELECT CAST(NULL as BINARY)
SELECT CAST(NULL as VOID)
```

all return the same value: `null`.

So what this PR ought to do is construct a TSparkParameter such that it writes out a CAST expression which returns a `null`. But I've run into trouble.

### Description

This pull request is supposed to add support for binding a Python `None` to a parameter marker in a query. **It does not work**. I'm opening this pull request and branch for visibility about the specific test I'm running.

When I tried setting TSparkParameter.value to `"NULL"` I get this exception:

```
[undefined]databricks.sql.exc.DatabaseError: [INVALID_PARAMETER_MARKER_VALUE.INVALID_VALUE_FOR_DATA_TYPE] An invalid parameter mapping was provided: the value 'NULL' for parameter 'p' cannot be cast to VOID because it is malformed.
self = <tests.e2e.test_driver.PySQLCoreTestSuite testMethod=test_dbsqlparam_inferred_none>
```

I assume this is because the resulting CAST expression looks like this:

```sql
SELECT CAST("NULL" as VOID)
```

Which doesn't work because `"NULL"` is a string and I need to actually cast the literal value `NULL` (no quote marks).

However, when I try setting the TSparkParameter value to `None` I get a blind transport error:

```
./tests/e2e/test_driver.py::PySQLCoreTestSuite::test_dbsqlparam_inferred_none Failed: [undefined]databricks.sql.exc.RequestError: Error during request to server: org.apache.thrift.transport.TTransportException
self = <tests.e2e.test_driver.PySQLCoreTestSuite testMethod=test_dbsqlparam_inferred_none>

    def test_dbsqlparam_inferred_none(self):
    
        params = [DbSqlParameter(name="p", value=None, type=None)]
>       result = self._get_one_result(self.QUERY, params)

tests/e2e/common/parameterized_query_tests.py:85: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/e2e/common/parameterized_query_tests.py:30: in _get_one_result
    cursor.execute(query, parameters=parameters)
src/databricks/sql/client.py:509: in execute
    execute_response = self.thrift_backend.execute_command(
src/databricks/sql/thrift_backend.py:844: in execute_command
    resp = self.make_request(self._client.ExecuteStatement, req)
src/databricks/sql/thrift_backend.py:483: in make_request
    self._handle_request_error(error_info, attempt, elapsed)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <databricks.sql.thrift_backend.ThriftBackend object at 0x10279b950>
error_info = RequestErrorInfo(error=TApplicationException(None), error_message='org.apache.thrift.transport.TTransportException', r...nifest=None, resultRetentionSeconds=None, resultByteLimit=None, resultDataFormat=None, originatingClientIdentity=None))
attempt = 1, elapsed = 0.10815000534057617

    def _handle_request_error(self, error_info, attempt, elapsed):
        max_attempts = self._retry_stop_after_attempts_count
        max_duration_s = self._retry_stop_after_attempts_duration
    
        if (
            error_info.retry_delay is not None
            and elapsed + error_info.retry_delay > max_duration_s
        ):
            no_retry_reason = NoRetryReason.OUT_OF_TIME
        elif error_info.retry_delay is not None and attempt >= max_attempts:
            no_retry_reason = NoRetryReason.OUT_OF_ATTEMPTS
        elif error_info.retry_delay is None:
            no_retry_reason = NoRetryReason.NOT_RETRYABLE
        else:
            no_retry_reason = None
    
        full_error_info_context = error_info.full_info_logging_context(
            no_retry_reason, attempt, max_attempts, elapsed, max_duration_s
        )
    
        if no_retry_reason is not None:
            user_friendly_error_message = error_info.user_friendly_error_message(
                no_retry_reason, attempt, elapsed
            )
            network_request_error = RequestError(
                user_friendly_error_message, full_error_info_context, error_info.error
            )
            logger.info(network_request_error.message_with_context())
    
>           raise network_request_error
E           databricks.sql.exc.RequestError: Error during request to server: org.apache.thrift.transport.TTransportException

src/databricks/sql/thrift_backend.py:314: RequestError
```

</detail>

### DIFF
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import copy
 import datetime
 import decimal
 from abc import ABC, abstractmethod
@@ -495,6 +494,7 @@ class DbSqlType(Enum):
     BOOLEAN = "BOOLEAN"
     INTERVAL_MONTH = "INTERVAL MONTH"
     INTERVAL_DAY = "INTERVAL DAY"
+    VOID = "VOID"
 
 
 class DbSqlParameter:
@@ -542,20 +542,40 @@ def infer_types(params: list[DbSqlParameter]):
         datetime.date: DbSqlType.DATE,
         bool: DbSqlType.BOOLEAN,
         Decimal: DbSqlType.DECIMAL,
+        type(None): DbSqlType.VOID,
     }
-    new_params = copy.deepcopy(params)
-    for param in new_params:
+
+    new_params = []
+    for param in params:
+        _name: str = param.name
+        _value: Any
+        _type: DbSqlType
+
         if not param.type:
             if type(param.value) in type_lookup_table:
-                param.type = type_lookup_table[type(param.value)]
+                _type = type_lookup_table[type(param.value)]
             else:
-                raise ValueError("Parameter type cannot be inferred")
+                raise ValueError(
+                    "Parameter type cannot be inferred from {} - {}".format(
+                        type(param.value), str(param.value)
+                    )
+                )
 
-        if param.type == DbSqlType.DECIMAL:
+        else:
+            _type = param.type
+            
+        if _type == DbSqlType.DECIMAL:
             cast_exp = calculate_decimal_cast_string(param.value)
-            param.type = DbsqlDynamicDecimalType(cast_exp)
+            _type = DbsqlDynamicDecimalType(cast_exp)
 
-        param.value = str(param.value)
+        if _type == DbSqlType.VOID:
+            new_params.append(DbSqlParameter(name=_name, type=DbSqlType.VOID))
+            continue
+        else:
+            _value = str(param.value)
+
+        new_params.append(DbSqlParameter(name=_name, value=_value, type=_type))
+
     return new_params
 
 
@@ -594,11 +614,15 @@ def named_parameters_to_tsparkparams(parameters: Union[List[Any], Dict[str, str]
         dbsql_params = named_parameters_to_dbsqlparams_v2(parameters)
     inferred_type_parameters = infer_types(dbsql_params)
     for param in inferred_type_parameters:
-        tspark_params.append(
-            TSparkParameter(
-                type=param.type.value,
-                name=param.name,
-                value=TSparkParameterValue(stringValue=param.value),
+        # The only way to pass a VOID/NULL to DBR is to declare TSparkParameter without declaring
+        # its value or type arguments. If we set these to NoneType, the request will fail with a
+        # thrift transport error
+        if param.type == DbSqlType.VOID:
+            this_tspark_param = TSparkParameter(name=param.name)
+        else:
+            this_tspark_param_value = TSparkParameterValue(stringValue=param.value)
+            this_tspark_param = TSparkParameter(
+                type=param.type.value, name=param.name, value=this_tspark_param_value
             )
-        )
+        tspark_params.append(this_tspark_param)
     return tspark_params
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -546,28 +546,27 @@ def infer_types(params: list[DbSqlParameter]):
     }
 
     new_params = []
+
+    # cycle through each parameter we've been passed
     for param in params:
         _name: str = param.name
-        _value: Any
+        _value: Any = param.value
         _type: DbSqlType
 
-        if not param.type:
-            if type(param.value) in type_lookup_table:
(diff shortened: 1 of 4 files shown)


rcypher-databricks (approved):

src/databricks/sql/utils.py
```diff
@@ -542,20 +542,41 @@ def infer_types(params: list[DbSqlParameter]):
         datetime.date: DbSqlType.DATE,
         bool: DbSqlType.BOOLEAN,
         Decimal: DbSqlType.DECIMAL,
+        type(None): DbSqlType.VOID,
```
benc-db, line 545: This is mostly something that should have been mentioned earlier, but this lookup stable is static, so pull it to an outer layer (possibly the outermost layer) so that it's not created on every call.

src/databricks/sql/utils.py
```diff
@@ -542,20 +542,41 @@ def infer_types(params: list[DbSqlParameter]):
...
-            else:
-                raise ValueError("Parameter type cannot be inferred")
-
-        if param.type == DbSqlType.DECIMAL:
+
+    new_params = []
+
+    # cycle through each parameter we've been passed
+    for param in params:
+        _name: str = param.name
```
benc-db, line 552: Strive for each unit of code having a single reason to change (Single Responsibility Principle).  E.g., here I would, at a minimum, suggest that the logic inside the loop be a separate function, but might go so far as to have the special logic for each type in their own methods, so that you end up with

def infer_types(params: list[DbSqlParameter]) -> list[DbSqlParameter]
def infer_type(param: DbSqlParameter) -> DbSqlParameter
def _convert_to_decimal_type(decimal_parameter: DbSqlParameter) -> DbSqlDynamicDecimalType
def _conver_to_void_type(decimal_parameter: DbSqlParameter) -> DbSqlVoidType

where infer_type has a switch statement (or if/elif I guess, stupid old pythons).

benc-db: Ideally, most functions in a dynamic language like python should be 10 lines or less, to improve comprehension and make testing easier (less likely to have to test combinations of logical branches together).

src/databricks/sql/utils.py
```diff
@@ -594,11 +615,15 @@ def named_parameters_to_tsparkparams(parameters: Union[List[Any], Dict[str, str]
...
     for param in inferred_type_parameters:
-        tspark_params.append(
-            TSparkParameter(
-                type=param.type.value,
-                name=param.name,
-                value=TSparkParameterValue(stringValue=param.value),
+        # The only way to pass a VOID/NULL to DBR is to declare TSparkParameter without declaring
+        # its value or type arguments. If we set these to NoneType, the request will fail with a
+        # thrift transport error
+        if param.type == DbSqlType.VOID:
```
benc-db, line 621: Recommend making a real type like DbsqlDynamicDecimalType, and then having each type implement a protocol for producing their TSparkParameter.  Then this part of the method becomes

...
[inferred_param.to_tspark_parameter() for inferred_param in inferred_type_parameters]

benc-db (approved):
Consider my feedback before merging, but it is not blocking.

nithinkdb (approved):

src/databricks/sql/utils.py
susodapop: This is really good feedback. Will push to a second pull request shortly.

PR #167, opened 2023-07-01
[PECO-840] WIP

susodapop:

PR #235, opened 2023-09-29
[PECO-840] Port staging ingestion behaviour to new UC Volumes

susodapop:
⚠️ **Note to reviewers: this will be easiest to review commit-by-commit as each change is small and logical. All these tests are copied from the original staging locations feature and only lightly augmented to support UC volumes.**

## Description

Last year we introduced staging locations, a feature which is being phased out with [UC volumes](https://learn.microsoft.com/en-us/azure/databricks/data-governance/unity-catalog/create-volumes) instead. No change is needed to the connector except to update our e2e tests to use the new UC volumes syntax. Which is what this pull request implements. I originally wrote this change in July 2023, as the feature was first rolling out and we didn't have a server to test against. We now have that server and I have verified that all our tests pass.

This is a reopen of #234 and #167


susodapop:
I synced with `main` to fix the merge conflict warning.

nithinkdb (approved):
LGTM!

benc-db (approved):

PR #238, opened 2023-10-01
SQLAlchemy 2: add type compilation for all CamelCase types

susodapop:
## Background

As we move toward SQLAlchemy 2 compatibility, this pull request starts tackling the type compiler. The [sqla documentation](https://docs.sqlalchemy.org/en/20/core/type_basics.html#module-sqlalchemy.types) is not easy to work through since it's written toward _users_ not dialect authors. For reviewers (and posterity) the key thing to understand about this PR is that we're currently only addressing type **compilation**, which is the way an instance of SQLAlchemy's `TypeEngine` is rendered as a string for inclusion in sql expressions (`CREATE TABLE`, `CAST` etc.)

_This PR does not affect the way type mapping (also called binding) occurs. Mapping/binding is the process of converting the data received from Databricks to- and from- SQLAlchemy Python types. That will be addressed in a subsequent PR._

## Description
SQLAlchemy's documentation tells users that they can expect every dialect to implement their generic [CamelCase types](https://docs.sqlalchemy.org/en/20/core/type_basics.html#generic-camelcase-types). This PR makes sure that if a user creates a table with one of those types, the correct [Databricks data type](https://docs.databricks.com/en/sql/language-manual/sql-ref-datatypes.html) is rendered. For example, SQLAlchemy wants to write a `DateTime()` type as a `DATETIME` field (which is supported on other backends). But Databricks uses a `TIMESTAMP` type for this. So this PR confirms that when a user declares a table with a `DateTime()` field, the `CREATE TABLE` statement will print the correct column type.

We do this using the `@compiles` decorator which is documented [here](https://docs.sqlalchemy.org/en/20/core/custom_types.html#overriding-type-compilation). This seems to be the modern, preferred way to override default compilation behaviour in SQLAlchemy.

There are two commits here which you may wish to examine separately:
- 5f0a650b8f1347301d97ee0a8253854c06da499f Extract the old type compiler (this is a pattern we inherited from pyhive's dialect when we first wrote ours). The recommended way in SQLA documentation is to use `@compiles` instead
- 2acd66672e52c65d9e6248aaddfcb3d53351d075 Add compiles directives and tests that prove that the compiler renders proper data type names when a generic (CamelCase) SQLAlchemy type definition is used.


susodapop:
I have confirmed that all existing e2e tests pass with this change. I have not run the SQLAlchemy dialect compliance tests yet as those will mostly be affected by updates to the type mapping rather than type compilation.

benc-db (approved):

PR #240, opened 2023-10-02
SQLAlchemy 2: add type compilation for all supported types

susodapop:
## Background

This PR finished the work started in https://github.com/databricks/databricks-sql-python/pull/238 by adding tests for compilation of Databricks-supported [UPPERCASE types](https://docs.sqlalchemy.org/en/20/core/type_basics.html#the-uppercase-datatypes). As of this PR, we can guarantee that all Databricks supported types are handled by the SQLAlchemy compiler.

Up next: implement value binding for these types!

## Description

I added a new test class called `TestUppercaseTypesCompilation` that uses the same fixtures as the last PR. It takes advantage of the pytest `parametrize` marker to keep the tests readable and avoid unecessary duplication.

The `ARRAY` type required special handling which I've documented in the code.


benc-db (approved):

susodapop:
I have verified that all e2e tests pass with this change.

PR #241, opened 2023-10-02
[PECO-1117] Allow legacy codepath for parameterized queries to be used.

nithinkdb:
We need to allow for parameterized queries to be used in a legacy matter.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -18,6 +18,7 @@ from databricks.sql.utils import (
     ExecuteResponse,
     ParamEscaper,
     named_parameters_to_tsparkparams,
+    inject_parameters
 )
 from databricks.sql.types import Row
 from databricks.sql.auth.auth import get_python_sql_connector_auth_provider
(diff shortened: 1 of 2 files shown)


src/databricks/sql/client.py
```diff
@@ -501,6 +502,11 @@ def execute(
         """
         if parameters is None:
             parameters = []
+        elif self.thrift_backend._use_legacy_parameters:
```
yunbodeng-db, line 505: I think we need to add unit tests to cover the behavior. Any new code should have unit tests.

What about an app which can connect to different dbrs (thinking of dbt). This new property can serve as a forcing function but the default behaviors should be allow the client code to continue to function. Would you check the server version and decide to use the server support or the old client support of the server does not support it?
If it is going to be addressed in a separate PR, it's fine.

yunbodeng-db:
Please sign your DCO. It's easy to do it now before you have more commits later.

src/databricks/sql/client.py
susodapop: We'll circulate a document internally about this decision. I don't think it should be handled / discussed in the Github code review comments as this isn't great visibility across the team.

Agree that this code should be tested and should not merge without tests.

susodapop:
```suggestion
        elif self.thrift_backend._use_inline_params:
```

I'd propose naming this as obviously unsafe / inline since `legacy` isn't indicative of what the change actually _does_.

susodapop (changes requested):
Don't merge this. We'll circulate a document internally so we can have multiple eyes on the design.

susodapop:
Closing as we implemented this in #267 with tests.

PR #244, opened 2023-10-10
[PECO-1134] v3 Retries: allow users to bound the number of redirects to follow

susodapop:
## Background

Some dbt users have found a spike in the number of redirect responses sent from sql gateway. In pysql, following redirect response consumes one attempt against the configured `_stop_after_attempts_count`. If the number of redirects for a given thrift command exceeds this limit, a `MaxRetryError` is raised.

Users can order the connector to follow more redirects by increasing the `_stop_after_attempts_count`.

## Description

This PR introduces the `_retry_max_redirects` configuration for users of the new v3 retry behaviour (which will become the default in the `3.0.0` release). The behaviour of this setting is well-tested in this PR. The critical piece of info is that on its own, `_retry_max_redirects` can only ever *decrease* the number of attempts pysql makes. It would require a significant change to our retry logic to trick the HTTP transport (`urrlib3` at the moment) to not decrement its internal `_attempts_remaining` in the case of redirects.

Outside of debugging, most users will not need to set `_retry_max_redirects`.

For Databricks users reading this PR: if you observe an unusually high number of redirect responses, please report this to Databricks support. This isn't an exact science, but the default `_retry_stop_after_attempts_count` of 30 should be more than enough to receive an actionable response from Databricks runtime. If the number of redirects leads you to increase this count above, say, `100` we should investigate the root cause of those redirects.

### DIFF
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -628,10 +628,7 @@ class ThriftBackend:
                 num_rows,
             ) = convert_column_based_set_to_arrow_table(t_row_set.columns, description)
         elif t_row_set.arrowBatches is not None:
-            (
-                arrow_table,
-                num_rows,
-            ) = convert_arrow_based_set_to_arrow_table(
+            (arrow_table, num_rows,) = convert_arrow_based_set_to_arrow_table(
                 t_row_set.arrowBatches, lz4_compressed, schema_bytes
             )
         else:
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -132,6 +132,8 @@ class ThriftBackend:
         # (defaults to False)
         # _retry_max_redirects
         #  An integer representing the maximum number of redirects to follow for a request.
+        #  This number must be <= _retry_stop_after_attempts_count.
+        #  (defaults to None)
         # max_download_threads
         #  Number of threads for handling cloud fetch downloads. Defaults to 10
 
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -189,7 +189,7 @@ class ThriftBackend:
         self.force_dangerous_codes = kwargs.get("_retry_dangerous_codes", [])
 
         additional_transport_args = {}
-        _max_redirects: int = kwargs.get("_retry_max_redirects", {})
+        _max_redirects: int = kwargs.get("_retry_max_redirects")
         if _max_redirects:
             urllib3_kwargs = {"redirect": _max_redirects}
         else:
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -130,6 +130,8 @@ class ThriftBackend:
         # _enable_v3_retries
         # Whether to use the DatabricksRetryPolicy implemented in urllib3
         # (defaults to False)
+        # _retry_max_redirects
+        #  An integer representing the maximum number of redirects to follow for a request.
         # max_download_threads
         #  Number of threads for handling cloud fetch downloads. Defaults to 10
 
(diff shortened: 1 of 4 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -130,6 +130,10 @@ def __init__(
         # _enable_v3_retries
         # Whether to use the DatabricksRetryPolicy implemented in urllib3
         # (defaults to False)
+        # _retry_max_redirects
+        #  An integer representing the maximum number of redirects to follow for a request.
```
benc-db, line 134: Is this validated anywhere?  Specifically, what is the user experience if retry_max_redirects > retry_stop_after_attempts_count?

src/databricks/sql/thrift_backend.py
```diff
@@ -185,6 +189,11 @@ def __init__(
         self.force_dangerous_codes = kwargs.get("_retry_dangerous_codes", [])
 
         additional_transport_args = {}
+        _max_redirects: int = kwargs.get("_retry_max_redirects", {})
```
benc-db, line 192: Why is there this multilayer structure?  I would expect _retry_max_redirects to be an int.

tests/e2e/common/retry_test_mixins.py
```diff
@@ -323,3 +330,90 @@ def test_retry_abort_close_operation_on_404(self):
...
+                        **self._retry_policy,
+                        "_retry_max_redirects": max_redirects,
+                    }
+                ) as conn:
+                    pass
+            assert "too many redirects" == str(cm.exception.reason)
+            # Total call count should be 3 (original + 2 retries)
+            assert mock_obj.return_value.getresponse.call_count == expected_call_count
+
+    def test_retry_max_redirects_raises_1_2(self):
```
benc-db, line 360: Are all of these actually testing different behavior modes?  If not, recommend removing the ones that are functional duplicates.

benc-db:
Big question though: what problem does this really solve?  I suspect the solution the users are looking for is that Databricks doesn't get stuck in a redirect loop.  This just fails faster, right?

src/databricks/sql/thrift_backend.py
susodapop: I'm not validating it as far as the user is concerned. It won't raise an exception or a log message. The behaviour is validated by tests, however.

src/databricks/sql/thrift_backend.py
susodapop: Good catch! This was an artefact from development. Fixed in 645a0304e86421ea1141d5f61be23b69c615f84c

src/databricks/sql/thrift_backend.py
benc-db: I don't love that the user gets no message about why we didn't respect their config (and that's the way it seems to me as a user).  I think its worth logging a warning, since setting it that way implies the user doesn't understand how the configs interact.

susodapop: It's not that we don't respect the config. We do, it's just that the configs disagree. I'll add a warning for this situation.

benc-db: Thank you, I meant that it *appeared* that you didn't respect the config :).  And adding the warning can clear up that appearance.

tests/e2e/common/retry_test_mixins.py
susodapop: It was mainly due diligence to confirm that the exception raised as expected for multiple different configs. Later, I added the assertion that the exception text is "too many redirects" rather than "too many $statusCode responses" which accomplishes the same thing. Will remove these faux-paramterised tests.

susodapop: Fixed in 21d418d34dc74fcbc6ffe19fecb97215c7dea07a

src/databricks/sql/thrift_backend.py
susodapop: Added and tested in 3c0c73e28d94b81a821ac4ee5237d2828a53b525

susodapop:
> Big question though: what problem does this really solve? I suspect the solution the users are looking for is that Databricks doesn't get stuck in a redirect loop. This just fails faster, right?

Correct. I mention in the PR description that this will typically only affect debugging. Initially this was envisioned as a way to independently control redirects without decrementing the overall `_attempts_remaining`. That plan was built on my misunderstanding of how urllib3 handles redirects. Because until this PR our handling of redirects was to let urllib3 handle them automatically. This is even called out in the docstring for `retry.py::should_retry`:

```
        Q: What about OSErrors and Redirects?
        A: urllib3 automatically retries in both scenarios
```

It wasn't til I began e2e testing the fix that the true behaviour became clearer. But at that point I'd already coded this extra config and figured the tests alone are worth the PR. And we may as well grant ourselves some extra debugging capability if this becomes an issue in the future.

benc-db (approved):

PR #248, opened 2023-10-12
[PECO-1137] Reintroduce protocol checking to Python test fw

nithinkdb:
We previously had protocol checks enabled, but this caused a regression, because it looked for protocol version inside the sessionhandle object, which is sometimes undefined. Now, we have changed the behavior such that we extract the protocol version from the sessionhandle if it exists, but if not, we fallback to getting it from the opensessionresponse, which is guaranteed to have it.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -228,11 +228,13 @@ class Connection:
     def get_session_id(self):
         return self.thrift_backend.handle_to_id(self._session_handle)
 
-    # Since the sessionHandle will sometimes have a serverProtocolVersion, it takes
-    # precedence over the serverProtocolVersion defined in the OpenSessionResponse.
     @staticmethod
     def get_protocol_version(openSessionResp):
-        if (
+        """
+        Since the sessionHandle will sometimes have a serverProtocolVersion, it takes
+        precedence over the serverProtocolVersion defined in the OpenSessionResponse.
+        """
+                if (
             openSessionResp.sessionHandle
             and hasattr(openSessionResp.sessionHandle, "serverProtocolVersion")
             and openSessionResp.sessionHandle.serverProtocolVersion
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -528,7 +528,7 @@ class Cursor:
         elif not Connection.server_parameterized_queries_enabled(
             self.connection.protocol_version
         ):
-            raise Error(
+            raise NotSupportedError(
                 "Parameterized operations are not supported by this server. DBR 14.1 is required."
             )
         else:
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -13,6 +13,7 @@ from databricks.sql.exc import (
     SessionAlreadyClosedError,
     CursorAlreadyClosedError,
 )
+from databricks.sql.thrift_api.TCLIService import ttypes
 from databricks.sql.thrift_backend import ThriftBackend
 from databricks.sql.utils import (
     ExecuteResponse,
@@ -196,9 +197,11 @@ class Connection:
             **kwargs,
         )
 
-        self._session_handle = self.thrift_backend.open_session(
+        self._open_session_resp = self.thrift_backend.open_session(
             session_configuration, catalog, schema
         )
+        self._session_handle = self._open_session_resp.sessionHandle
+        self.protocol_version = self.get_protocol_version(self._open_session_resp)
         self.use_cloud_fetch = kwargs.get("use_cloud_fetch", True)
         self.open = True
         logger.info("Successfully opened session " + str(self.get_session_id_hex()))
@@ -225,10 +228,15 @@ class Connection:
     def get_session_id(self):
         return self.thrift_backend.handle_to_id(self._session_handle)
 
-    def get_session_protocol_version(self):
-        return self.thrift_backend.extract_protocol_version_from_handle(
-            self._session_handle
-        )
+    @staticmethod
+    def get_protocol_version(openSessionResp):
+        if (
+            openSessionResp.sessionHandle
+            and hasattr(openSessionResp.sessionHandle, "serverProtocolVersion")
+            and openSessionResp.sessionHandle.serverProtocolVersion
+        ):
+            return openSessionResp.sessionHandle.serverProtocolVersion
+        return openSessionResp.serverProtocolVersion
 
     @staticmethod
     def server_parameterized_queries_enabled(protocolVersion):
@@ -518,7 +526,7 @@ class Cursor:
             parameters = []
 
         elif not Connection.server_parameterized_queries_enabled(
-            self.connection._session_handle
+            self.connection.protocol_version
         ):
             raise Error(
                 "Parameterized operations are not supported by this server. DBR 14.1 is required."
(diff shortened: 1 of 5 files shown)


src/databricks/sql/client.py
```diff
@@ -501,6 +524,13 @@ def execute(
         """
         if parameters is None:
             parameters = []
+
+        elif not Connection.server_parameterized_queries_enabled(
+            self.connection.protocol_version
+        ):
+            raise Error(
+                "Parameterized operations are not supported by this server. DBR 14.1 is required."
```
susodapop, line 532: Our plan is to re-enable the code path for inlining parameters in a subsequent PR, yes?

src/databricks/sql/client.py
```diff
@@ -501,6 +524,13 @@ def execute(
         """
         if parameters is None:
             parameters = []
+
+        elif not Connection.server_parameterized_queries_enabled(
+            self.connection.protocol_version
+        ):
+            raise Error(
```
susodapop, line 531: We should raise a `NotSupportedError` rather than the generic `Error` here.

tests/unit/test_parameters.py
```diff
@@ -9,15 +9,68 @@
...
+from databricks.sql.thrift_api.TCLIService import ttypes
 
+from decimal import Decimal
+from databricks.sql.client import Connection
 from typing import List
 
 
+class TestSessionHandleChecks(object):
+    def test_get_session_handle(self):
+        assert (
```
susodapop, line 28: For new tests we should have one assertion per test. Please refactor this either into separate tests or a single test using `pytest.mark.parametrize`.

src/databricks/sql/client.py
```diff
@@ -225,6 +228,26 @@ def __del__(self):
     def get_session_id(self):
         return self.thrift_backend.handle_to_id(self._session_handle)
 
+    @staticmethod
+    def get_protocol_version(openSessionResp):
+        if (
```
susodapop, line 233: Since this is a new behaviour to fetch first from the `sessionHandle` versus falling back to the `openSessionResp`, please add a docstring or comment explaining why we do this.

tests/unit/test_parameters.py
```diff
@@ -9,15 +9,68 @@
...
+            Connection.get_protocol_version(
+                TOpenSessionResp(
+                    serverProtocolVersion=ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V7,
+                    sessionHandle=TSessionHandle(1, None),
+                )
+            )
+            == ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V7
```
susodapop, line 35: Per my comment on the definition of `get_protocol_version`, the test names or comments should explain what we're testing for here. These assertions verify that `get_protocol_version` falls back to the `openSessionResp.serverProtocolVersion` if it's not present in `sessionHandle`. We should have some indication as to _why_.

tests/unit/test_parameters.py
```diff
@@ -9,15 +9,68 @@
...
+            == ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V8
+        )
+        assert (
+            Connection.get_protocol_version(
+                TOpenSessionResp(
+                    serverProtocolVersion=ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V8,
+                    sessionHandle=TSessionHandle(1, None),
+                )
+            )
+            == ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V8
```
susodapop, line 55: This test assertion is effectively a duplicate of the first assertion currently on line 35. Consider removing or separating into a separate test with a justification for why this test is required in addition to the first one.

tests/unit/test_parameters.py
```diff
@@ -9,15 +9,68 @@
...
+        )
+
+    def test_parameters_enabled(self):
+        assert Connection.server_parameterized_queries_enabled(None) == False
```
susodapop, line 59: Similar to feedback above: use one assertion per test case. This is a natural use-case for `pytest.mark.parametrize`.

src/databricks/sql/client.py
```diff
@@ -196,9 +197,11 @@ def read(self) -> Optional[OAuthToken]:
             **kwargs,
         )
 
-        self._session_handle = self.thrift_backend.open_session(
+        self._open_session_resp = self.thrift_backend.open_session(
             session_configuration, catalog, schema
         )
+        self._session_handle = self._open_session_resp.sessionHandle
```
susodapop, line 203: The change on line 199 and 203 effectively changes the type of `self._session_handle` from an `openSessionResponse` to an actual `sessionHandle` object. Have you confirmed that this works in the other places where `self._session_handle` is used?

I suspect that it _does_ work because both the old type (`openSessionResponse`) and the new type (`sessionHandle`) have a property called `serverProtocolVersion`. But if anywhere else in the code we check for a property other than `self._session_handle.serverProtocolVersion` and this property is not present on the `sessionHandle` type, it will fail.

My guess is that the original name for this property (`self._session_handle`) was imprecise and your change is making the name more accurately match its contents 👍

src/databricks/sql/thrift_api/TCLIService/ttypes.py
```diff
@@ -36,6 +36,7 @@ class TProtocolVersion(object):
     SPARK_CLI_SERVICE_PROTOCOL_V5 = 42245
     SPARK_CLI_SERVICE_PROTOCOL_V6 = 42246
     SPARK_CLI_SERVICE_PROTOCOL_V7 = 42247
+    SPARK_CLI_SERVICE_PROTOCOL_V8 = 42248
```
susodapop, line 39: Please confirm: this regeneration of `ttypes.py` is identical to the original implementation that we reverted, yes? It was necessary so we could capture the new `..._V8` value.

susodapop (changes requested):
Change looks good. Please update the pr description to describe the actual change and its motivation. The core change here is that you have consolidated the way the connector checks the SPARK_CLI_SERVICE_PROTOCOL version so that it now looks in two places instead of one place. Why was this change necessary?

src/databricks/sql/client.py
nithinkdb: Yep!

src/databricks/sql/thrift_api/TCLIService/ttypes.py
nithinkdb: Yep

src/databricks/sql/client.py
nithinkdb: No, the original behavior was to return the same type of handle.

susodapop: Thanks for clarifying. I think somehow I missed the final clause of line 203 🤦

tests/e2e/common/parameterized_query_tests.py
```diff
@@ -34,8 +37,19 @@ def _quantize(self, input: Union[float, int], place_value=2) -> Decimal:
...
 
-    def test_primitive_inferred_none(self):
+    @patch(
+        "databricks.sql.client.Connection.server_parameterized_queries_enabled",
+        return_value=False,
+    )
+    def test_protocol_too_low(self, mock_parameterized_queries_enabled):
+        params = {"p": None}
+        with pytest.raises(
+            Error,
```
susodapop, line 47: Might as well make this test more specific. A `NotSupportedError` should be raised. Else this test will pass for _any_ known exception (connection failed, syntax error, network stack failure etc)

susodapop (approved):
Okay to merge once these two small changes are applied 🙏

PR #249, opened 2023-10-13
SQLAlchemy 2: implement and refactor schema reflection methods

susodapop:
## Background

The biggest piece of work required for this PR was figuring how to get `ComponentReflectionTest` to even _run_ as it's setUp and tearDown methods were failing for various reasons. 

## Description

This PR changes the has_table, get_pk_constraint, and get_foreign_keys methods into functional shape. They now pass the related tests in ComponentReflectionTest. `get_foreign_keys` is **finished**, following the guidance from @benc-db with regard to encapsulating strange parsing logic into more, smaller methods.

In a follow-up PR I will make the same refactor to `get_pk_constraint`, but I wanted to get eyes on the approach for `get_foreign_keys` before I proceed.

I implemented a new `_describe_table_extended` method that is re-used by severaral methods.


src/databricks/sqlalchemy/__init__.py
```diff
@@ -31,6 +33,141 @@ class DatabricksImpl(DefaultImpl):
         __dialect__ = "databricks"
 
 
+import logging
+
+logger = logging.getLogger(__name__)
+
+DBR_LTE_12_NOT_FOUND_STRING = "Table or view not found"
+DBR_GT_12_NOT_FOUND_STRING = "TABLE_OR_VIEW_NOT_FOUND"
```
susodapop, line 41: These duplicate lines will be removed in a subsequent PR.

src/databricks/sqlalchemy/__init__.py
```diff
@@ -31,6 +33,141 @@ class DatabricksImpl(DefaultImpl):
...
+
+
```
benc-db, line 43: Why is all of this stuff in __init__.py?

susodapop: It's a biproduct of earlier decisions. I'll move it out of __init__.py in a follow-up PR.

susodapop: The guidance from sqlalchemy is that this should be in a file called `base.py`.

benc-db (approved):
Approved assuming followup refactor from \__init\__,py to base.py.  Documentation and style looks good.

PR #251, opened 2023-10-16
SQLAlchemy 2: Finish implementing all of ComponentReflectionTest

susodapop:
This is the final pull request implementing our support for component reflection of SQLAlchemy. As with other PR's towards this feature, the specifics of the implementation are esoteric and what matters most is that the tests pass.

For context, the `ComponentReflection` test is the largest block of tests in SQLAlchemy's dialect compliance test suite. It comprises roughly 50% of the test cases. As of this PR, every test in this suite is either passing or skipped with good reason (feature not supported by Databricks or simply not prioritised for implementation).

After this PR will come one more which clears out the remaining test cases that need to be called out in `test_suite.py`. After that I'll update the documentation for users and the new-and-improved dialect will be ready for release.

**note for reviewers**: This is based on the changes in #250 which will be reviewed separately. Once that PR merges I will merge `main` into this PR. But if you feel like reviewing before that happens, just look at the commits beginning with [66497ca](https://github.com/databricks/databricks-sql-python/pull/251/commits/66497cad4ae794a5d4480f277b21034ab76aa3c2)

### DIFF
--- a/src/databricks/sqlalchemy/_ddl.py
+++ b/src/databricks/sqlalchemy/_ddl.py
@@ -23,7 +23,7 @@ class DatabricksDDLCompiler(compiler.DDLCompiler):
         pass
 
     def visit_check_constraint(self, constraint, **kw):
-        logger.warn("Databricks does not support check constraints")
+        logger.warning("This dialect does not support check constraints")
         pass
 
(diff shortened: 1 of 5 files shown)


susodapop:
I rebased off `main` now that #250 has merged.

PR #254, opened 2023-10-19
Fix syntax error in http handler

susodapop:
This code path never runs from DBR because we never send a `Set-Cookie` header. But some internal users during testing have found this codepath raises a syntax error.

### DIFF
--- a/src/databricks/sql/auth/thrift_http_client.py
+++ b/src/databricks/sql/auth/thrift_http_client.py
@@ -107,7 +107,6 @@ class THttpClient(thrift.transport.THttpClient.THttpClient):
         self.retry_policy and self.retry_policy.start_retry_timer()
 
     def open(self):
-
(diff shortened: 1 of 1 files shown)


src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -191,7 +189,7 @@ def flush(self):
 
         # Saves the cookie sent by the server response
         if "Set-Cookie" in self.headers:
-            self.setCustomHeaders(dict("Cookie", self.headers["Set-Cookie"]))
+            self.setCustomHeaders({"Cookie", self.headers["Set-Cookie"]})
```
andreid-db, line 192: I don't think setCustomHeaders is correct here, this will replace all existing headers . Shouldn't it be something like headers.update to append ?

benc-db:
Close in favor of newer PR

Issue #176, opened 2023-07-11
Please add async/await support


thetadweller:
I am writing a web app that needs to run multiple concurrent queries against Databricks SQL Warehouse. Due to existing library is synchronous my processes tend to get locked for a duration of SQL query so that subsequent calls from other clients end up being queued. As such,  I am forced to run multiple Python processes to handle multiple concurrent calls even though all of them are I/O bound and would have been handled by a handful of processes had I been able to write queries using async / await.

I tried to find a workaround using SQLAlchemy and async I/O wrappers but returned a message that connection is not asynchronous:
`InvalidRequestError: The asyncio extension requires an async driver to be used. The loaded 'databricks-sql-python' is not async.`

susodapop:
Thanks for raising this issue. I agree that `async` support would be a huge asset in the growing world of asyncio Python.

This is a nontrivial project though, namely because we use urllib3 for our HTTP handling and urllib3 doesn't support async. In the short term you can work around this with multiple `stdlib/threading` rather than multiple processes but obviously the ergonomics there aren't good. We have an example of it in `examples/query_cancel.py`

I'm self-assigning this issue for now.

I'd like to see more signal from other users so we can gauge the demand for this feature. Single-threaded usage comprises the majority of customer feature requests with regard to this connector (the connector is 18 months old and this request has only come up once before [here](https://github.com/databricks/databricks-sql-python/issues/82)).

The more people ask for it the easier it will be to prioritise

krishnamenon22:
Similar Request here. Key concern is to handle timeout when cluster boots up .

vhrichfernandez:
I need to run 30 async queries in parallel to see if they will parse in UC with 1 minute timeout to consider it a parseable query in Unity Catalog. I'm using `concurrent.futures.ProcessPoolExecutor` but getting the following error:

```
>               _, error_list = execute_sql_queries_async(list(tables.values()))

tests/infrastructure/database/test_dbx_connect.py:25:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
sql_utils.py:240: in execute_sql_queries_async
    _, error = future.result()
/root/.pyenv/versions/3.10.12/lib/python3.10/concurrent/futures/_base.py:451: in result
    return self.__get_result()
[...]

PR #260, opened 2023-10-25
Add support for async/await syntax

susodapop:
This is a staging PR for #176. As PR's merge into this branch we can rerun the code checks (since these don't run for PR's against the non-primary branch).

Further description TBD


susodapop:
We're dramatically modifying the result fetching behaviour which will heavily conflict with this PR for the time-being. Closing for now.

PR #264, opened 2023-10-27
sqla2 clean-up: make sqlalchemy optional and don't mangle the user-agent

### LINKED ISSUE #192: User Agent string is getting mangled
We are passing in a string for our user agent (Unsupervied) to the connector via the DSN and using SQLAlchemy to access Databricks. "sqlalchemy + " is getting added over and over. The screenshot has 5 examples of it being added, but it is variable - we see some with hundreds.

I think this matters to Databricks as an org more than end-developers as this is breaking your analytics about tools using Databricks.

image

susodapop:
## Description

This PR fixes #192 with an e2e test and updates pyproject.toml to make sqlalchemy + alembic optional. After this, if users wish to install the dialect's dependencies they can

```shell
pip install databricks-sql-connector[sqlalchemy]
```

to get SQLA's dependencies and

```shell
pip install databricks-sql-connector[alembic]
```

for alembic's dependencies.

Closes #192

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,13 +2,13 @@
 
 [[package]]
 name = "alembic"
-version = "1.12.0"
+version = "1.12.1"
 description = "A database migration tool for SQLAlchemy."
-optional = false
+optional = true
 python-versions = ">=3.7"
 files = [
-    {file = "alembic-1.12.0-py3-none-any.whl", hash = "sha256:03226222f1cf943deee6c85d9464261a6c710cd19b4fe867a3ad1f25afda610f"},
-    {file = "alembic-1.12.0.tar.gz", hash = "sha256:8e7645c32e4f200675e69f0745415335eb59a3663f5feb487abfa0b30c45888b"},
+    {file = "alembic-1.12.1-py3-none-any.whl", hash = "sha256:47d52e3dfb03666ed945becb723d6482e52190917fdb47071440cfdba05d92cb"},
+    {file = "alembic-1.12.1.tar.gz", hash = "sha256:bca5877e9678b454706347bc10b97cb7d67f300320fa5c3a94423e8266e2823f"},
 ]
 
 [package.dependencies]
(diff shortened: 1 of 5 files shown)


poetry.lock
```diff
@@ -237,75 +252,72 @@ test = ["pytest (>=6)"]
...
-    {file = "greenlet-2.0.2-cp39-cp39-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl", hash = "sha256:be4ed120b52ae4d974aa40215fcdfde9194d63541c7ded40ee12eb4dda57b76b"},
-    {file = "greenlet-2.0.2-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:94c817e84245513926588caf1152e3b559ff794d505555211ca041f032abbb6b"},
-    {file = "greenlet-2.0.2-cp39-cp39-musllinux_1_1_aarch64.whl", hash = "sha256:1a819eef4b0e0b96bb0d98d797bef17dc1b4a10e8d7446be32d1da33e095dbb8"},
-    {file = "greenlet-2.0.2-cp39-cp39-musllinux_1_1_x86_64.whl", hash = "sha256:7efde645ca1cc441d6dc4b48c0f7101e8d86b54c8530141b09fd31cef5149ec9"},
-    {file = "greenlet-2.0.2-cp39-cp39-win32.whl", hash = "sha256:ea9872c80c132f4663822dd2a08d404073a5a9b5ba6155bea72fb2a79d1093b5"},
-    {file = "greenlet-2.0.2-cp39-cp39-win_amd64.whl", hash = "sha256:db1a39669102a1d8d12b57de2bb7e2ec9066a6f2b3da35ae511ff93b01b5d564"},
-    {file = "greenlet-2.0.2.tar.gz", hash = "sha256:e7c8dc13af7db097bed64a051d2dd49e9f0af495c26995c00a9ee842690d34c0"},
+    {file = "greenlet-3.0.1-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:f89e21afe925fcfa655965ca8ea10f24773a1791400989ff32f467badfe4a064"},
```
benc-db, line 260: Did you mean to check in a change with greenlet in it?

benc-db: nm, I realize this is probably from a dependency's dependency

benc-db (approved):

poetry.lock
susodapop: Correct. SQLAlchemy installs `greenlet` to handle their async implementation, which is described [here](https://gist.github.com/zzzeek/2a8d94b03e46b8676a063a32f78140f1) (in case you're curious 😉 )

PR #267, opened 2023-10-30
Native Parameters: reintroduce INLINE approach with tests

susodapop:
## Background

In #217, we replaced the previous inline parameterisation behaviour with native, server-side bound parameters. Prior to the release of version 3.0.0 of this connector, we are re-introducing the legacy behaviour (and keeping it as the default behaviour) controlled by a configuration called `use_inline_params`. This PR reintroduces the inline approach.

## Description

This PR does four things:

- It reintroduces the `INLINE` parameterisation approach and makes that the default.
- It forces our SQLAlchemy dialect to use the `NATIVE` approach
- It reorganises and updates our e2e tests to cope with this change, including a short-term patch that will be removed once our internal testing compute instance properly advertises its ability to handle native parameters.
- It introduces a placeholder page where, in a subsequent PR, we'll document the behaviour of inline vs native parameters and how users can go about moving from one to the other. 

With this change the connector is now capable of running parameterised queries with one of two approaches:

- **INLINE** uses the legacy parameter approach which has existed in this connector since its first version. With this approach, passed parameter values are rendered as escaped strings within the query body. This mode is used by default and by setting `use_inline_params` when calling `sql.connect()`.
- **NATIVE** uses server-side bound parameters, support for which was added in DBR 14.1. This is a safer approach to parameterised queries because parameter values are understood to be user input by DBR and are appropriately escaped. 

Before executing a query, this connector performs a protocol version check to ensure that the server can accept native parameters. But note that this protocol check is _broken_ because some DBR instances which _do_ support native parameters nevertheless return a protocol version that suggests otherwise. I've added a patch to our connector test suite and the SQLAlchemy test suite to override this.

As for tests, prior to this PR we had end-to-end tests for the native parameter approach but incomplete tests of the inline approach. I therefore reorganised the native parameter tests so they can be parameterised based on a `ParameterApproach`. This involved removing `unittest.TestCase` as the base-class for the parameterised query tests and relocating those tests from the `common` directory and into the top-level `tests/e2e/` directory. This improves convenience of running these tests because they can now be invoked directly and as a group. Before this change, they were mixed-in to `PySQLCoreTestSuite` and could not easily be invoked in isolation.


src/databricks/sql/__init__.py
```diff
@@ -5,7 +5,7 @@
 # PEP 249 module globals
 apilevel = "2.0"
 threadsafety = 1  # Threads may share the module, but not connections.
-paramstyle = "named"  # Python extended format codes, e.g. ...WHERE name=%(name)s
+paramstyle = "pyformat"  # Python extended format codes, e.g. ...WHERE name=%(name)s
```
susodapop, line 8: This change reverts an erroneous change I made when developing SQLAlchemy.

src/databricks/sql/client.py
```diff
@@ -519,36 +608,60 @@ def execute(
...
-            parameters = []
 
-        elif not Connection.server_parameterized_queries_enabled(
-            self.connection.protocol_version
-        ):
-            raise NotSupportedError(
-                "Parameterized operations are not supported by this server. DBR 14.1 is required."
-            )
+        if parameters:
+            param_approach = self._determine_parameter_approach()
```
susodapop, line 639: I'm open to feedback from reviewers. Does it make sense to remove the conditional on line 638 and then update `_determine_parameter_approach` so that it can sense that no parameters were provided and return `ParameterApproach.NONE`?

tests/e2e/test_driver.py
```diff
@@ -101,6 +101,60 @@ def assertEqualRowValues(self, actual, expected):
         for act, exp in zip(actual, expected):
             self.assertSequenceEqual(act, exp)
 
+class PySQLPytestTestCase():
```
susodapop, line 104: This is a duplicate of `PySQLTestCase` that doesn't subclass `unittest.TestCase`. This is necessary if we want to use `pytest.mark.parametrize` since that decorator fails when applied to tests contained in a subclass of `unittest.TestCase`.

tests/e2e/test_driver.py
```diff
@@ -144,7 +198,7 @@ def test_cloud_fetch(self):
 # Exclude Retry tests because they require specific setups, and LargeQueries too slow for core
 # tests
 class PySQLCoreTestSuite(SmokeTestMixin, CoreTestMixin, DecimalTestsMixin, TimestampTestsMixin,
-                         PySQLTestCase, PySQLStagingIngestionTestSuiteMixin, PySQLRetryTestsMixin, PySQLParameterizedQueryTestSuiteMixin, PySQLUCVolumeTestSuiteMixin):
```
susodapop, line 147: I moved the parameterized cases into their own logical block of tests so they don't need to be imported here.

tests/e2e/test_driver.py
```diff
@@ -341,7 +395,7 @@ def test_escape_single_quotes(self):
             assert rows[0]["col_1"] == "you're"
 
             # Test escape syntax in parameter
-            cursor.execute("SELECT * FROM {} WHERE {}.col_1 LIKE :var".format(table_name, table_name), parameters={"var": "you're"})
+            cursor.execute("SELECT * FROM {} WHERE {}.col_1 LIKE %(var)s".format(table_name, table_name), parameters={"var": "you're"})
```
susodapop, line 398: I erroneously changed this test case while developing the SQLAlchemy dialect. This merely reverts it to the correct syntax.

tests/e2e/test_parameterized_queries.py
```diff
@@ -0,0 +1,382 @@
...
+            "databricks.sql.client.Connection.server_parameterized_queries_enabled",
+            return_value=True,
+        ) as mock_parameterized_queries_enabled:
+            try:
+                yield mock_parameterized_queries_enabled
+            finally:
+                pass
+
+    def _inline_roundtrip(self, params: dict):
+        target_column = self.inline_type_map[type(params.get("p"))]
```
susodapop, line 133: This INSERT, SELECT, DELETE dance is necessary because simply selecting "SELECT %(param)s" in INLINE mode would always return a `str` and the nature of the test is to confirm that types are maintained.

susodapop: I'll add the above context to the docstring for this method.

tests/e2e/test_parameterized_queries.py
```diff
@@ -0,0 +1,382 @@
...
+    def _native_roundtrip(
+        self, parameters: Union[Dict, List[Dict]], bypass_patch=False
+    ):
+        with self.connection(extra_params={"use_inline_params": False}) as conn:
+            with conn.cursor() as cursor:
+                with self.conditional_protocol_patch(bypass_patch):
+                    cursor.execute(self.NATIVE_QUERY, parameters=parameters)
+                    return cursor.fetchone()
+
+    def _get_one_result(self, approach: ParameterApproach, params, bypass_patch=False):
```
susodapop, line 157: `bypass_patch` is only needed when another test is already patching the protocol version.

tests/e2e/test_parameterized_queries.py
```diff
@@ -0,0 +1,382 @@
+import datetime
```
susodapop, line 1: Most of the contents of this file were moved from `parameterized_query_tests.py`. I renamed the file because pytest won't discover it unless the filename begins with `test_`.

tests/e2e/test_parameterized_queries.py
```diff
@@ -0,0 +1,382 @@
...
+        params = {"p": None}
+        with pytest.raises(
+            NotSupportedError,
+            match="Parameterized operations are not supported by this server. DBR 14.1 is required.",
+        ):
+            result = self._get_one_result(
+                ParameterApproach.NATIVE, params, bypass_patch=True
+            )
+
+    @both_approaches
```
susodapop, line 201: Any test that is not decorated with `@both_approaches` is effectively unchanged in this PR, apart from updating the call to `_get_one_result` to include a `ParameterApproach`.

tests/e2e/test_parameterized_queries.py
```diff
@@ -0,0 +1,382 @@
...
+    def test_primitive_inferred_decimal(
+        self, approach: ParameterApproach, inline_table
+    ):
+        params = {"p": Decimal("1234.56")}
+        result = self._get_one_result(approach, params)
+        assert result.col == Decimal("1234.56")
+
+    def test_dbsqlparam_inferred_none(self):
```
susodapop, line 258: `@both_approaches` is not used for any of the non-primitive test cases because INLINE mode is not compatible `DbsqlParameter` objects.

nithinkdb (approved):
LGTM!

tests/e2e/test_parameterized_queries.py
susodapop: Done.

src/databricks/sql/client.py
susodapop: I have applied this change.

tests/e2e/test_driver.py
yunbodeng-db: If a server supports native param query and the app sets use_inline_params False, it'd use `SELECT * FROM {} WHERE {}.col_1 LIKE :var` format?

src/databricks/sql/__init__.py
```diff
@@ -5,7 +5,10 @@
 # PEP 249 module globals
 apilevel = "2.0"
 threadsafety = 1  # Threads may share the module, but not connections.
-paramstyle = "named"  # Python extended format codes, e.g. ...WHERE name=%(name)s
+
+# Python extended format codes, e.g. ...WHERE name=%(name)s
+# Note that when we switch to ParameterApproach.NATIVE, paramstyle will be `named`
+paramstyle = "pyformat"
```
yunbodeng-db, line 11: What is the use of `paramstyle` here?
Should we say: inline = pyformat while native = named?
I think I am not very clear on this.

yunbodeng-db (commented):
Mostly LGTM. I did leave two comments that I basically wonder which style is going to be used. My simple mind says: pyformat style if `use_inline_params` is True otherwise named format. Whether the server supports native or not is independent?

tests/e2e/test_driver.py
susodapop: This is an old test. Prior to this PR, it was the _only_ e2e test of inline parameter rendering. We could consolidate it into the same test file with the other tests and update it to dynamically shift its syntax. For now, if we ran this test with `use_inline_params=False`, it would fail with a syntax error. 

However, since we're running the e2e test suite with `use_inline_params=True` by default the test is technically correct. And the new tests added in this PR already add coverage of this behaviour.

Also, we have unit tests for the parameter escaper which I wrote last year.

susodapop:
@yunbodeng-db wrote:
> Mostly LGTM. I did leave two comments that I basically wonder which style is going to be used. My simple mind says: pyformat style if use_inline_params is True otherwise named format. Whether the server supports native or not is independent?

This is correct. In a PR tomorrow I'll add a doc page that discusses this in detail for users to consume. tl;dr use `pyformat` for inline params, `named` for native params.

PR #275, opened 2023-11-09
Remove block on running queries if protocol seems to low

susodapop:
## Description

This PR accomplishes two things:
1. It stops pysql from abandoning a parameterised query run with `ParameterApproach.NATIVE`  if it thinks that DBR doesn't support it.
2. It removes the patch from our test fixtures that was required prior to implementing part 1.

This PR hits a staging branch so I can group all the forthcoming parameter approach changes and documentation into a single commit on `main`.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -413,12 +413,8 @@ class Cursor:
                 )
             return ParameterApproach.INLINE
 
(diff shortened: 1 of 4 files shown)


susodapop:
All e2e tests pass for this change.

benc-db (approved):

PR #276, opened 2023-11-09
Add paramstyle transformer

susodapop:
## Background

The goal of this PR is to make it so that queries that include parameter markers in the older `pyformat` paramstyle as defined in PEP-249 can still run with connector >= 3.0.0 when `use_inline_params=False`. Without this change, any existing parameterised queries would not execute at all because they use a parameter marker syntax that DBR doesn't understand.

Thefore it's desirable that query like this:

```sql
SELECT * FROM table WHERE field = %(param)s
```
can be executed with the native parameter approach. We do this by automatically converting `pyformat` parameter markers into `named` format markers. So the above query would be executed as:

```sql
SELECT * FROM table WHERE field = :param
```

## Description

There are a few significant changes in this PR:
1. It implements a method called `transform_paramstyle` which performs a regex looking for named `pyformat` parameter markers in a query string and replaces them with `named` markers. If there are no `pyformat` markers, the query text is returned as-is. There are unit tests for this behaviour.
2. It makes it so that `transform_paramstyle` is called on the query text whenever parameters are passed to `cursor.execute()` and `use_inline_params=False`.
3. It updates all of our parameterised query tests to cope with this new dimension.

_Most_ of the changes are to the e2e tests. Those changes are in short:

### Define a 3-dimension test matrix for the roundtrip parameterisation tests
The components of the matrix are: 
  - The **type** of Python value being parameterised
  - The **approach** to parameterisation being used
  - The **paramstyle** to be used in the query.

When running with the inline approach, only `pyformat` paramstyle is used.

### Refactor roundtrip tests to use `pytest.mark.parametrize`

This makes the test-code easier to read. And also protects against unintended differences between assertions for different types.

This means that these tests:
- `test_dbsqlparam_inferred_none`
- `test_dbsqlparam_inferred_bool`
- `test_dbsqlparam_inferred_integer`
- `test_dbsqlparam_inferred_double`
- `test_dbsqlparam_inferred_date`
- `test_dbsqlparam_inferred_timestamp`
- `test_dbsqlparam_inferred_string`
- `test_dbsqlparam_inferred_decimal`

Are replaced with a single, parameterised test:
- `test_dbsql_param_with_inferrence`

Ditto for the primitive and custom decimal type tests.

## Up next

We'll switch the default parameter approach from inline to native and document all of this in README.parameters.md.

### DIFF
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -390,10 +390,7 @@ def inject_parameters(operation: str, parameters: Dict[str, str]):
 
 def transform_paramstyle(operation: str) -> str:
     """
-    Transform a paramstyle string to a format string.
-
     Performs a regex substitution such that any occurence of `%(param)s` will be replaced with `:param`
-    as long as `param` is a key in the parameters dictionary.
 
     This utility function is built to assist users in the transition between the default paramstyle in
     this connector prior to version 3.0.0 (`pyformat`) and the new default paramstyle (`named`).
@@ -578,7 +575,8 @@ def resolve_databricks_sql_integer_type(integer):
 
     Note: TINYINT is never inferred here because it is a rarely used type and clauses like LIMIT and OFFSET
     cannot accept TINYINT bound parameter values. If you need to bind a TINYINT value, you can explicitly
-    declare its type in a DbsqlParameter object, which will bypass this inference logic."""
+    declare its type in a DbsqlParameter object, which will bypass this inference logic.
+    """
     if -128 <= integer <= 127:
         # If DBR is ever updated to permit TINYINT values passed to LIMIT and OFFSET
         # then we can change this line to return DbSqlType.TINYINT
@@ -599,8 +597,9 @@ TYPE_INFERRENCE_LOOKUP_TABLE = {
     Decimal: DbSqlType.DECIMAL,
     type(None): DbSqlType.VOID,
 }
-def infer_types(params: list[DbSqlParameter]):
 
+
+def infer_types(params: list[DbSqlParameter]):
     new_params = []
 
     # cycle through each parameter we've been passed
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import datetime
 import decimal
+import re
 from abc import ABC, abstractmethod
 from collections import OrderedDict, namedtuple
 from collections.abc import Iterable
@@ -385,6 +386,39 @@ class ParamEscaper:
 
 def inject_parameters(operation: str, parameters: Dict[str, str]):
     return operation % parameters
+
+
+def transform_paramstyle(operation: str) -> str:
+    """
+    Transform a paramstyle string to a format string.
+
+    Performs a regex substitution such that any occurence of `%(param)s` will be replaced with `:param`
+    as long as `param` is a key in the parameters dictionary.
+
+    This utility function is built to assist users in the transition between the default paramstyle in
+    this connector prior to version 3.0.0 (`pyformat`) and the new default paramstyle (`named`).
+
+    Args:
+        operation (str): The operation or SQL text to transform.
+
+    Returns:
+        str
+    """
+
+    _output_operation = operation
+
+    PYFORMAT_PARAMSTYLE_REGEX = r"%\((\w+)\)s"
+    pat = re.compile(PYFORMAT_PARAMSTYLE_REGEX)
+    NAMED_PARAMSTYLE_FMT = ":{}"
+    PYFORMAT_PARAMSTYLE_FMT = "%({})s"
+
+    pyformat_markers = pat.findall(operation)
+    for marker in pyformat_markers:
+        pyformat_marker = PYFORMAT_PARAMSTYLE_FMT.format(marker)
+        named_marker = NAMED_PARAMSTYLE_FMT.format(marker)
+        _output_operation = _output_operation.replace(pyformat_marker, named_marker)
+
+    return _output_operation
 
 
 def create_arrow_table_from_arrow_file(file_bytes: bytes, description) -> pyarrow.Table:
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -589,17 +589,17 @@ def resolve_databricks_sql_integer_type(integer):
         return DbSqlType.BIGINT
 
 
+TYPE_INFERRENCE_LOOKUP_TABLE = {
+    str: DbSqlType.STRING,
+    int: DbSqlType.INTEGER,
+    float: DbSqlType.FLOAT,
+    datetime.datetime: DbSqlType.TIMESTAMP,
+    datetime.date: DbSqlType.DATE,
+    bool: DbSqlType.BOOLEAN,
+    Decimal: DbSqlType.DECIMAL,
+    type(None): DbSqlType.VOID,
+}
 def infer_types(params: list[DbSqlParameter]):
-    type_lookup_table = {
-        str: DbSqlType.STRING,
-        int: DbSqlType.INTEGER,
-        float: DbSqlType.FLOAT,
-        datetime.datetime: DbSqlType.TIMESTAMP,
-        datetime.date: DbSqlType.DATE,
-        bool: DbSqlType.BOOLEAN,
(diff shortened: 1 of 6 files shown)


benc-db:
Question: why `%(var)*s*?`  Will it always be `s`?

src/databricks/sql/utils.py
```diff
@@ -387,6 +388,36 @@ def inject_parameters(operation: str, parameters: Dict[str, str]):
...
+    Args:
+        operation (str): The operation or SQL text to transform.
+
+    Returns:
+        str
+    """
+
+    _output_operation = operation
+
+    PYFORMAT_PARAMSTYLE_REGEX = r"%\((\w+)\)s"
```
benc-db, line 407: is this too restrictive?  Can you link to pep or whatever that describes this?  In particular I'm wondering if %( thing )s is allowed.

susodapop:
Oh good question. I picked that final `s` because it's consistent with the only other e2e test we have that tests this behaviour. But since `inject_parameters` is just a convenience wrapper around python's string interpolation, it's very possible we have users with invocations like:

```python
query = "SELECT * FROM users WHERE user_id = %(user_id)d"
parameters = {"user_id": 1234}
cursor.execute(query, parameters)
```

I can add an e2e test for this and revise the regex to capture-and-replace any valid format specifier. I wonder just how complex could be the format specifiers customers are using...depending on the edge-cases it could make sense to use a different approach for this.

What I'm thinking is that rather than regexing for the parameter marker, we could simply render it inline and then do `str.replace()` where we replace the *rendered value* with the new parameter marker. Thoughts?

src/databricks/sql/utils.py
susodapop: This is the specification: https://peps.python.org/pep-0249/#paramstyle

It's ambiguous as to what format codes are expected to work. But in reality, since we're doing string interpolation under the covers, it doesn't matter so much what PEP-249 requires. Because if customers have queries that depend on this behaviour their queries will break if we don't do something to cover the use case.

jadewang-db, line 407: do we have an existing regex for inline replacement today, can we reuse that logic?

susodapop: Nope. We're just doing string interpolation so Python handles it for us, in C.

Have a look at `inject_parameters` to see exactly what it does:

https://github.com/databricks/databricks-sql-python/blob/0e7ce57ae4fecc364c4de55d6adace33c39aff5f/src/databricks/sql/utils.py#L386-L387

jadewang-db: can we call to inject_parameters which replace parameters with the native format of parameter

susodapop:
[This is the complete format Mini-Language specification.](https://docs.python.org/3/library/string.html#formatspec). For simplicity, I like the approach of simply rendering the `pyformat` marker with its parameter value and then replacing _that_ with our new marker.

benc-db:
> Oh good question. I picked that final `s` because it's consistent with the only other e2e test we have that tests this behaviour. But since `inject_parameters` is just a convenience wrapper around python's string interpolation, it's very possible we have users with invocations like:
> 
> ```python
> query = "SELECT * FROM users WHERE user_id = %(user_id)d"
> parameters = {"user_id": 1234}
> cursor.execute(query, parameters)
> ```
> 
> I can add an e2e test for this and revise the regex to capture-and-replace any valid format specifier. I wonder just how complex could be the format specifiers customers are using...depending on the edge-cases it could make sense to use a different approach for this.
> 
> What I'm thinking is that rather than regexing for the parameter marker, we could simply render it inline and then do `str.replace()` where we replace the _rendered value_ with the new parameter marker. Thoughts?

So, the only examples I could find all use %()s.  Even for things like dates.  Now technically, because you were just doing python string interpolation before, people might have used crazier things than that, but I don't think we have any obligation to support things outside of the norm.  If they are doing crazy things, the rendered value could be unexpected (consider if they passed you date formatting for example).  I think the simple regex approach is fine.

jadewang-db:
after this change if user use :param style query, it will work right?

susodapop:
@jadewang-db 
> after this change if user use :param style query, it will work right?

Yes, but only when the native parameter approach is used. It will fail at DBR when `use_inline_params=True` because DBR will see a parameter marker and expect that a `TSparkParameter` was passed in the request. But pysql will not have sent a `TSparkParameter` because the inline approach doesn't do that.

src/databricks/sql/utils.py
susodapop: That's what `transform_paramstyle()` already does...

susodapop:
@benc-db 
> So, the only examples I could find all use %()s. Even for things like dates. Now technically, because you were just doing python string interpolation before, people might have used crazier things than that, but I don't think we have any obligation to support things outside of the norm. If they are doing crazy things, the rendered value could be unexpected (consider if they passed you date formatting for example). I think the simple regex approach is fine.

Works for me!

benc-db:
> [This is the complete format Mini-Language specification.](https://docs.python.org/3/library/string.html#formatspec). For simplicity, I like the approach of simply rendering the `pyformat` marker with its parameter value and then replacing _that_ with our new marker.

I'm skeptical that this is possible without having _exactly_ as complicated a regex.  Consider that a date format could lead to the word August being rendered.  In order to know that August is what you're looking for, you would need to extract their formatting code.  But if you have a regex that can extract their formatting code, you may as well just use that for finding and replacing without rendering the pyformat first.  Willing to be proven wrong :P.

benc-db (approved):

susodapop:
> I'm skeptical that this is possible without having exactly as complicated a regex. Consider that a date format could lead to the word August being rendered. In order to know that August is what you're looking for, you would need to extract their formatting code. But if you have a regex that can extract their formatting code, you may as well just use that for finding and replacing without rendering the pyformat first. Willing to be proven wrong :P.

Well you can't ask me to prove you wrong and not expect me to try 😛 

```python
def transform_paramstyle(operation: str, parameters: Dict[str, Any]) -> str:

    COMPLETE_PYFORMAT_REGEX = r'%\((\w+)\)([diouxXeEfFgGcrsa%])'
    INNER_NAME_REGEX = r'\((.*?)\)'
    NAMED_PARAMSTYLE_FMT = ":{}"

    pat = re.compile(COMPLETE_PYFORMAT_REGEX)
    inner_pat = re.compile(INNER_NAME_REGEX)

    pyformat_markers = pat.findall(operation)

    for marker in pyformat_markers:
        _name = re.findall(inner_pat, marker)[0]
        rendered_value = inject_parameters(_name, parameters.get(_name))
        operation = operation.replace(rendered_value, NAMED_PARAMSTYLE_FMT.format(_name))

    return operation
```

tests/e2e/test_parameterized_queries.py
```diff
@@ -173,185 +278,10 @@ def test_use_inline_by_default_with_warning(self, explicit_inline, caplog):
                             "Consider using native parameters." in caplog.text
                         ), "Log message should not be supressed"
 
-    @both_approaches
```
nithinkdb, line 176: Why are we removing all of these?

nithinkdb (approved):
LGTM

tests/e2e/test_parameterized_queries.py
susodapop: We're not...

susodapop: It's just replaced with an extra dimension to the matrix, as described in the PR description.

https://github.com/databricks/databricks-sql-python/blob/a86abd35fe525676369d49b9124bb1c2aaa19fbb/tests/e2e/test_parameterized_queries.py#L52-L56

susodapop:
I pushed some new e2e tests of the inline ordinal parameter approach. And updated the logic that calls `transform_paramstyle`. Using @jadewang-db 's approach means that we now must pass a dictionary of parameters to `transform_parameters`. Whereas the regex approach didn't look at the parameters whatsoever.

As a consequence, `transform_paramstyle` won't work if `cursor.execute()` is passed a list of parameters. So I've revised the logic to make sure we don't do that.

One side-effect of this is that users cannot pass a `List[DbsqlParameter]` and also use the `pyformat` paramstyle. But we wouldn't expect users to do that, because `DbsqlParameter` itself is brand new for the forthcoming 3.0.0 release. i.e. this doesn't break backwards compatibility.

jadewang-db (approved):

PR #277, opened 2023-11-10
Use native parameter approach by default

susodapop:
## Description

This PR updates the connector so that the native parameter approach is enabled by default. It modifies the semantics of `use_inline_params` so that it now accepts three values:

- `False` (default) Native parameter approach will be used.
- `True`: Inline parameter approach will be used, and a warning will be issued
- `"silent"`: Inline parameter approach will be used, and no warning will be logged.

It also moves the warning message from the `execute()` method (where a warning would be logged on every "offending" query execution) and into the client construtor. This way the warning is logged _once_. We'd like to be kind to our friends that examine their logs 😉


src/databricks/sql/client.py
```diff
@@ -222,8 +227,36 @@ def read(self) -> Optional[OAuthToken]:
         logger.info("Successfully opened session " + str(self.get_session_id_hex()))
         self._cursors = []  # type: List[Cursor]
 
-        self._suppress_inline_warning = "use_inline_params" in kwargs
-        self.use_inline_params = kwargs.get("use_inline_params", True)
+        self.use_inline_params = self._set_use_inline_params_with_warning(
+            kwargs.get("use_inline_params")
+        )
+
+    def _set_use_inline_params_with_warning(self, user_value):
```
benc-db, line 234: types?

src/databricks/sql/client.py
```diff
@@ -222,8 +227,36 @@ def read(self) -> Optional[OAuthToken]:
...
+        """Valid values are True, False, and "silent"
+
+        False: Use native parameters
+        True: Use inline parameters and log a warning
+        "silent": Use inline parameters and don't log a warning
+        """
+
+        if user_value is None:
```
benc-db, line 242: declare as Option and set default in parameter?

src/databricks/sql/client.py
```diff
@@ -222,8 +227,36 @@ def read(self) -> Optional[OAuthToken]:
...
+
+        if user_value not in [True, False, "silent"]:
+            raise ValueError(
+                f"Invalid value for use_inline_params: {user_value}. "
+                + 'Valid values are True, False, and "silent"'
+            )
+
+        if user_value is True:
+            logger.warning(
+                "Parameterised queries executed on with this client will use the inlien parameter approach."
```
benc-db, line 253: typo: inlien

benc-db: Also, let's inform that we will deprecate in a future.

src/databricks/sql/client.py
susodapop: Fixed!

benc-db (approved):

PR #279, opened 2023-11-11
Add support for ordinal parameters

susodapop:
## Description

This is the final PR in the implementation for native parameter support in this connector prior to the version 3.0.0 release. It does a few major things:

1. It updates our logic for building `TSparkParameter` objects so that we can set `ordinal=True`.
2. It updates the test matrix for parameters with a new dimension: `ParameterStructure`.
3. It moves all the code for inferring types into a dedicated module rather than mixing it into `utils.py`
4. It refactors all the type inference code into the `TDbsqlParameter` type, as suggested by @benc-db during a previous review.

The `TestParameterizedQueries` test class now defines a matrix with _four_ dimensions:
- Parameter type
- `ParameterApproach`: NATIVE and INLINE
- `ParameterStyle`: `%(pyformat)s` and `:named`
- `ParameterStructure`: NAMED and POSITIONAL

### Passing parameters by position under the inline approach

In #276 we added the ability to run queries written with `pyformat` paramstyle under the native approach, by automatically transforming `pyformat` markers to `named` paramstyle. This behaviour is tested with the `ParameterStyle` dimension of our test matrix. But there is a wrinkle when it comes to parameters passed by position.

Passing positional parameters in a DBAPI compliant way is a new feature in version 3.0.0. However, there has been a workaround under the inline approach due to the specific implementation of our inline value renderer. Under the inline approach, users could pass parameters by position like so:

```python
query = "SELECT * FROM table WHERE field = %s"
params=  [1]
cursor.execute(query, params)
```

And this would produce a query that looks like this:

```sql
SELECT * FROM table WHERE field = 1
```

With this PR, that behaviour still works when `use_inline_params=True`. But it will not work when `use_inline_params=False`. 

When `use_inline_params=False`, the syntax for passing parameters by position looks like this:

```sql
SELECT * FROM table WHERE field = ?
```

Users will need to update their queries containing anonymous markers `%s` → `?`. We can't do this automatically because `%s` is actually valid SQL, whereas a pyformat-style parameter marker is not. If we do a blind substitution without parsing the user input then our transformer would mangle a query like this:

```sql
-- before transformation
SELECT * FROM table WHERE field LIKE '%some' and other_field = %s

-- after transform
SELECT * FROM table WHERE field LIKE '?ome' and other_field = ?
```

Therefore, in our testing matrix we do not test combinations of the inline parameter approach and positional parameter structure.

**Note:** In a follow-up PR I will add documentation and revised docstrings for the `DbsqlParameter` class, as we expect users to begin using this going forward.

### Usage Example

Here's a usage example demonstrating the two ways that parameters can be passed positionally:
1. `List[Primitive]`
2. `List[TDbsqlParameter]`

These three snippets would produce an identical thrift `ExecuteStatement` command:
```python
# Params passed as a list of primitive values
query = "SELECT * FROM table WHERE field IN (?, ?, ?)"
parameters = [1,2,3]
with sql.connect(...) as connection:
  with connection.cursor() as cursor:
    result = cursor.execute(query, parameters)

# Params passed as TDbsqlParameter objects, whose types will be inferred
# This produces the exact same output as the first snippet above
query = "SELECT * FROM table WHERE field IN (?, ?, ?)"
parameters = [
  IntegerParameter(value=1),
  StringParameter(value='hello'),
  DecimalParameter(value=Decimal("23.51", scale=10, precision=2)
]
with sql.connect(...) as connection:
  with connection.cursor() as cursor:
    result = cursor.execute(query, parameters)
```


src/databricks/sql/utils.py
```diff
@@ -387,26 +393,121 @@ def inject_parameters(operation: str, parameters: Dict[str, str]):
...
 
     Returns:
         str
     """
+    output = operation
+    if (
+        param_structure == ParameterStructure.POSITIONAL
+        and _may_contain_inline_positional_markers(operation)
+    ):
+        logger.warning(
```
susodapop, line 501: I'm not sure we really need this log message.

src/databricks/sql/parameters.py
```diff
@@ -0,0 +1,204 @@
...
+    int,
+    float,
+    datetime.datetime,
+    datetime.date,
+    bool,
+    decimal.Decimal,
+    Type[None],
+)
+
+ListOfParameters = Union[List[DbsqlParameter], List[PrimitiveType]]
```
benc-db, line 159: Minor: prefer ParameterList, ParameterDict.  Prepositions look weird in types :P.

src/databricks/sql/client.py
```diff
@@ -433,6 +439,43 @@ def _determine_parameter_approach(
...
 
+    def _only_dbsql_parameters_in_list(self, params: ListOfParameters) -> bool:
+        """Return True if all members of the list are DbsqlParameter instances"""
+        return all([isinstance(i, DbsqlParameter) for i in params])
+
+    def _all_dbsql_parameters_are_named(self, params: List[DbsqlParameter]) -> bool:
+        """Return True if all members of the list have a non-null .name attribute"""
+        return all([i.name is not None for i in params])
+
+    def _list_of_params_can_use_named_structure(self, params: ListOfParameters) -> bool:
```
benc-db, line 450: when your private/protected methods are named properly, and are obvious (i.e. one liners like here), doc strings are superfluous.  No need to delete what you have, but when you've gone to trouble of making the code clean, you don't need as much doc.

src/databricks/sql/client.py
```diff
@@ -433,6 +439,43 @@ def _determine_parameter_approach(
...
+
+    def _determine_parameter_structure(
+        self,
+        parameters: Optional[Union[ListOfParameters, DictOfParameters]],
+    ) -> ParameterStructure:
+
+        if parameters is None:
+            return ParameterStructure.NONE
+
+        if not isinstance(parameters, (list, dict, tuple)):
```
benc-db, line 469: this method is not part of the public interface, so in theory, your code is the only thing invoking it.  In such cases, I would say its being very paranoid/defensive to validate your types at runtime.  Better to spend your effort ensuring that you use type assignments extensively, and rely on mypy to ensure you only call this method with a valid type.

src/databricks/sql/client.py
```diff
@@ -458,8 +501,36 @@ def _prepare_inline_parameters(
 
         return rendered_statement, NO_NATIVE_PARAMS
 
+    def _prepare_dbsql_params_from_list(
+        self, params: ListOfParameters
+    ) -> List[DbsqlParameter]:
+        """Return a list of DbsqlParameter objects from the passed list of params"""
+
+        output = []
+        for p in params:
```
benc-db, line 510: minor, but I think the cleaner way to handle this is to have a factory method (could be your __init__ or could be a create_from(...)) for DbsqlParameter that can take p as a value or as a DbsqlParameter.  Then you don't really need this method, because getting your list is [DbsqlParameter.create_from(p) for p in params]

src/databricks/sql/client.py
```diff
@@ -458,8 +501,36 @@ def _prepare_inline_parameters(
...
+                output.append(DbsqlParameter(value=p))
+
+        return output
+
+    def _prepare_dbsql_params_from_dict(
+        self, params: DictOfParameters
+    ) -> List[DbsqlParameter]:
+        """Return a  list of DbsqlParameter objects from the passed dictionary"""
+
+        output = []
```
benc-db, line 523: As has probably become evident from my other comment, I prefer comprehensions so long as they're not nested.  They make it easier to think/process functionally, because instead of a step by step imperative process that builds up a list with side effects, you can think of the comprehension as a one step data transformation:

```
[DbsqlParameter(name=name, value=value) for (name,value) in params.items()]
```

src/databricks/sql/client.py
```diff
@@ -479,9 +550,20 @@ def _prepare_native_parameters(
...
 
-        return stmt, params
+        if isinstance(params, dict):
+            dbsql_params = self._prepare_dbsql_params_from_dict(params)
+        if isinstance(params, (list, tuple)):
+            dbsql_params = self._prepare_dbsql_params_from_list(params)
+
+        output = []
+
+        for p in dbsql_params:
```
benc-db, line 561: Same comment as elsewhere, prefer list comprehensions for the extremely simple list building.

src/databricks/sql/utils.py
benc-db: If we're going to try to support %s at all, then yes i think we need this message.  As mentioned above, I would instead suggest we explicitly not support it due to conflicts with our like syntax.

benc-db (commented):
Consider the stylistic feedback I've provided here.  My biggest concern is around trying to support %s.

src/databricks/sql/utils.py
```diff
@@ -387,26 +393,118 @@ def inject_parameters(operation: str, parameters: Dict[str, str]):
...
+        names = params.keys()
+    elif isinstance(params, list):
+        names = _dbsqlparameter_names(params)  # type: ignore
+
+    return {name: f":{name}" for name in names}
+
+
+def _may_contain_inline_positional_markers(operation: str) -> bool:
+    """Check for the presence of `%s` in the operation string.
+    If it's present, the operation may contain positional query markers. Return True.
```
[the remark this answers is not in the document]
benc-db: back the named parameter issue, what do we do here?

```
select * from foo like '%(name)s'
```

crud.

src/databricks/sql/parameters/native.py
```diff
@@ -0,0 +1,604 @@
...
+    SMALLINT = auto()
+    STRING = auto()
+    TIMESTAMP = auto()
+    TIMESTAMP_NTZ = auto()
+    TINYINT = auto()
+    ARRAY = auto()
+    MAP = auto()
+    STRUCT = auto()
+
+TAllowedParameterValue = Union[
```
benc-db, line 57: I would get rid of init on base.  a.) you don't want users to instantiate a DbsqlParameterBase, and b.) it forces you to break Liskov substitution principal by having the child types take a more restricted set of inputs.

benc-db (approved):
Consider dropping the __init__ method where I commented.

PR #280, opened 2023-11-15
Document parameter behaviour in version 3

susodapop:
## Description

This is a documentation PR for  the new parameter support in version 3.

- [x] README.parameters.md
- [x] New parameters.py example
- [x] Update changelog

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -212,7 +212,7 @@ The result of the above two examples is identical.
 
 **Note**: `%s` is not compliant with PEP-249 and only works due to the specific implementation of our inline renderer. 
 
-**Note:** This `%s` syntax overlaps with valid SQL syntax around the usage of `LIKE` DML. For example if your query includes a clause like `WHERE field LIKE '%sequence'`, the parameter inlining function will raise an exception because this string appears to include an inline marker but none is provided.
+**Note:** This `%s` syntax overlaps with valid SQL syntax around the usage of `LIKE` DML. For example if your query includes a clause like `WHERE field LIKE '%sequence'`, the parameter inlining function will raise an exception because this string appears to include an inline marker but none is provided. When `use_inline_params=False`, we will pass `%s` occurrences along to the database, allowing it to be used as expected in `LIKE` statements.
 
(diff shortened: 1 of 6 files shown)


susodapop:
If you run `examples/parameters.py` the output looks like this:

```
EXAMPLE 1
Example 1 result in native mode →        Row(name='Jane', age=30, active=True)
Example 1 result in inline mode →        Row(name='Jane', age=30, active=True)

EXAMPLE 2
Example 2 result with pyformat  →        Row(name='Jane', age=30, active=True)
Example 2 result with named     →        Row(name='Jane', age=30, active=True)

EXAMPLE 3
Example 3 result in native mode →        Row(name='Jane', age=30, active=True)
Example 3 result in inline mode →        Row(name='Jane', age=30, active=True)

EXAMPLE 4
Example 4 inferred result       →        2012-10-15 17:57:18+00:00      12.3456
Example 4 explicit result       →        2012-10-15 12:57:18            12.35
```

docs/parameters.md
```diff
@@ -1 +1,255 @@
...
+        query = "SELECT field FROM table WHERE field = %s AND another_field = %s"
+        parameters = ["foo", 20]
+        result = cursor.execute(query, parameters=parameters).fetchone()
+```
+
+The result of the above two examples is identical.
+
+**Note**: `%s` is not compliant with PEP-249 and only works due to the specific implementation of our inline renderer. 
+
+**Note:** This `%s` syntax overlaps with valid SQL syntax around the usage of `LIKE` DML. For example if your query includes a clause like `WHERE field LIKE '%sequence'`, the parameter inlining function will raise an exception because this string appears to include an inline marker but none is provided.
```
benc-db, line 215: you should explain in this note that if they switch to using named parameters, we will pass this along to the database, allowing it to be used as expected in the LIKE statements.  This might be inferred from the other documentation, but better to be explicit here.

benc-db (approved):
Looks good, only one recommendation to take a look at.

docs/parameters.md
susodapop: Added!

PR #281, opened 2023-11-15
Merge native parameter rewrite into main

susodapop:
## Description

Over the past week, we've completely rewritten the native parameter support and tests. Each step of the way has undergone discussion and review in pull requests against the `native-query-params-staging` branch. This PR exists to merge those changes into `main` in advance of the release.

Apart from those changes, this PR also performs a little cleanup of the code:
1. I found that there was an untested case around users passing SQL LIKE wildcards (`%`). I implemented the fix and added tests for it.
2. I found that running the SQLAlchemy reusable test suite was not possible without a few overrides for CTETest and ComponentReflection test. I added these. The test suite now passes 👍 
3. I removed some commented-out code that I missed during native parameters development
4. I refined the type declaration for `TParameterList` → `TParameterSequence` as this was necessary for SQLAlchemy tests and is consistent with the previous behaviour of the connector and PEP-249

The commit history on this branch is self-explanatory.

### Steps

- [x] Verify unit tests pass
- [x] Verify e2e tests pass
- [x] Verify SQLAlchemy reusable dialect tests pass

### DIFF
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
+        python-version: [3.8, 3.9, "3.10", "3.11"]
     steps:
       #----------------------------------------------
       #       check-out repo and set-up python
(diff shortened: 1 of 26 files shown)


susodapop:
SQLAlchemy reusable tests are currently running. These generally take an hour or so.

benc-db (approved):

PR #282, opened 2023-11-16
Enable v3 retries by default

susodapop:
## Description

Ahead of the v3 release later this week, this PR sets `_enable_v3_retries=True` by default and emits a warning log if users override it. This means that anyone installing `databricks-sql-connector==3.0.0` will be automatically upgraded to the new retry behaviour.

I had two fix two problems with the retry behaviour as well:

### GetOperationStatus would keep retrying despite a 404 response

In the case where a GetOperationStatus request raised a 404 there were
two problems:

1. urllib3 would just keep retrying the request, even though it's not
recoverable, until it ran out of configured retry attempts
2. when the max attempts was exceeded, the connector would abandon its
connection from urllib3's connection pool. whereas it should clean up
its connection by draining it and releasing it back to the connection pool.

As a result of 2, the `test_cancel_during_execute` test case would spin
forever because the attempt to close the open session during the test
teardown would _start_ from a connection that wasn't drained and end up
in an evaluation loop.

I fixed this in two ways:
- I updated the retry policy so it won't continually retry GetOperationStatus
requests that fail with a 404.
- I moved the call to self._transport.close() out of the main try: block
and into a finally block so that it's _always_ called.

This allows the failing tests to pass.

### An assertion in `test_close_connection_closes_cursors` used an out-of-date `Exception` subclass

This was an easy fix. Just swap `TApplicationException` (which was raised by thrift's transport implementation) with `SessionAlreadyClosedError` (which is raised by our custom transport implementation).

### DIFF
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
+        python-version: [3.8, 3.9, "3.10", "3.11"]
     steps:
       #----------------------------------------------
       #       check-out repo and set-up python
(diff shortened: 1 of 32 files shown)


susodapop:
I have verified that this PR passes all unit and e2e tests.

yunbodeng-db (approved):

benc-db (approved):

PR #287, opened 2023-11-21
Update default parameterisation behaviour in doc

Falydoor:
With the `3.0.0` release, I believe the documentation about the default mode wasn't updated.


susodapop (approved):
Great catch, thank you!

PR #293, opened 2023-11-28
[PECO-1286] Complex types tests

susodapop:
## Description

This PR adds e2e tests for the complex type behaviour of this connector. The test demonstrates that when a user selects a field which contains a complex type value that this type is passed through the result as a List, `ndarray`, or `dict()` depending on the type of the field in the table.

This can be disabled with a config when building the connection.

As part of an ongoing effort to improve the "hintability" of this connector in a code editor, I've also extracted `_use_arrow_native_complex_types` from `kwargs` and added comments for it.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -61,6 +61,7 @@ class Connection:
         session_configuration: Dict[str, Any] = None,
         catalog: Optional[str] = None,
         schema: Optional[str] = None,
+        _use_arrow_native_complex_types: Optional[bool] = True,
         **kwargs,
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


src/databricks/sql/client.py
```diff
@@ -152,8 +153,13 @@ def read(self) -> Optional[OAuthToken]:
                             experimental_oauth_persistence=DevOnlyFilePersistence("~/dev-oauth.json")
                         )
                 ```
-
-
+            :param _use_arrow_native_complex_types: `bool`, optional
+                Controls whether a complex type field value is returned as a string or as a native Arrow type. Defaults to True.
+                When True:
+                    MAP is returned as List[Tuple[str, Any]]
```
benc-db, line 159: Why is MAP not returned as a Dict[str, any]?

benc-db (approved):

src/databricks/sql/client.py
susodapop: I had the same thought but this seems to be controlled by `pyarrow`. Updating to use a `dict()` is a potential improvement we could implement.

benc-db: Unclear if it would be an improvement.  If this is the native type, let's leave it be.

PR #297, opened 2023-11-30
Bump package version to 3.0.1

susodapop:
## Description

I'd like to release the fix to SQLAlchemy that merged today. This PR bumps to a patch version as all these changes are backwards compatible and introduce no new functionality.


susodapop:
the test publish to pypi failed because of a temporary issue with deployments to test.pypi.org. nothing wrong with this code.

PR #298, opened 2023-12-01
[PECO-1263] Refactor parameter handling logic entirely into parameters module

susodapop:
## Description

NOTE: this PR goes against a new staging branch for the async execution work.

This PR mostly shuffles existing code around. Back when I added native parameter handling, I created the `sql.parameters.native.py` file. This PR gathers all the inline parameter handling code into `sql.parameters.inline.py`. I also added `sql.parameters.choose.py` which exposes a single method that `client.py` now uses to determine the parameter style, structure, and approach for a query execution.

This work is part of the broader effort to introduce the `AsyncExecution` object type. Since the forthcoming `execute_async()` function will be part of the `Connection` object, I needed to relocate the helper methods of `Cursor` (since async executions won't depend on cursors at all). And after looking at them, I realised that most of those helper functions are pure and could be defined anywhere.

This means that when I implement the `execute_async()` method, it can call `prepare_parameters_and_statement()` without needing to duplicate code anywhere.

## Tests

I have run all the e2e and unit tests and verify they still pass.


src/databricks/sql/parameters/choose.py
```diff
@@ -0,0 +1,80 @@
...
+
+def determine_parameter_approach(
+    connection: "Connection", params: Optional["TParameterCollection"]
+) -> ParameterApproach:
+    """Encapsulates the logic for choosing whether to send parameters in native vs inline mode
+
+    If params is None then ParameterApproach.NONE is returned.
+    If self.use_inline_params is True then inline mode is used.
+    If self.use_inline_params is False, then check if the server supports them and proceed.
+        Else raise an exception.
```
benc-db, line 32: Although I believe this is just a refactor and not actually new code, I noticed something:

We claim that if self.use_inline_params is False, we check if the server supports them and proceed, else raise an exception.  None of that is in the current function.  This is precisely why I recommend not detailing how a function works in the doc string, as it is a maintenance burden.  Recommend one sentence on what a function does, and no discussion of how it does.  We don't want other code to depend on the how, just the what.

benc-db (approved):
Approve, but suggest you fix the comment string I mentioned.

src/databricks/sql/parameters/choose.py
susodapop: Fixed!

PR #303, opened 2023-12-19
Fix URI construction in ThriftBackend

### LINKED ISSUE #230: server_hostname string should be less picky
If you include 'https://' or a final '/' in your server_hostname, you currently get a difficult to parse error.  It would be better to either silently strip these pieces off, or to return a very clear error to the user.

NodeJSmith:
closes #230 

This pull request fixes the URI construction in the `ThriftBackend` class of the `thrift_backend.py` file. The current implementation does not handle the case where the `server_hostname` is provided with the `https://` prefix, resulting in duplicate `https://` in the URI. This PR addresses that issue by checking if the `uri` starts with `https://` and adding it if necessary.

Additionally, this PR also fixes the case where the `server_hostname` is provided with a trailing backslash. The current implementation duplicates the trailing backslash in the URI, which is incorrect. This PR removes the duplicate backslash.

Lastly, this PR includes unit tests to verify the correctness of the URI construction.

Changes:
- Updated the URI construction logic in the `ThriftBackend` class.
- Added unit tests for the URI construction.

Please review and merge this pull request.

### DIFF
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -138,7 +138,7 @@ class ThriftBackend:
             uri = kwargs.get("_connection_uri")
         elif server_hostname and http_path:
             uri = "{host}:{port}/{path}".format(
-                host=server_hostname, port=port, path=http_path.lstrip("/")
+                host=server_hostname.rstrip("/"), port=port, path=http_path.lstrip("/")
             )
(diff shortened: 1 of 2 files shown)


susodapop (approved):
Looks great. Thank y ou1

susodapop:
I've run our e2e tests for this change. Will merge once the PR checks pass.

PR #306, opened 2023-12-22
Add support for column comments

cbornet:

PR #308, opened 2023-12-27
Add support for table comments

cbornet:
NB: I didn't find a way to get the comments of temporary views
* temporary views don't appear in `information_schema`
* `DESCRIBE TABLE EXTENDED` doesn't return anything for temporary views
* `SHOW CREATE TABLE` isn't supported for temporary views
* `SHOW TABLE EXTENDED` doesn't report comments for temporary views (it does for default views)
* The thrift protocol `TGetTablesResp` message contains the temporary views but the `REMARKS` field is empty for all tables

The TCK tests `test_comments_unicode` and `test_comments_unicode_full` are still skipped at the moment since they also test column comments which are added in another PR: #306


src/databricks/sqlalchemy/test/_extra.py
```diff
@@ -63,8 +66,39 @@ class DateTimeTZTestCustom(_DateFixture, fixtures.TablesTest):
...
         # so that the value is processed correctly.
         result = connection.scalar(select(literal(self.data, TIMESTAMP)))
         eq_(result, self.data)
+
+
+class TableDDLTestCustom(fixtures.TestBase):
+    """This test confirms that a table comment can be dropped.
+    The difference with TableDDLTest is that the comment value is '' and not None after
+    being dropped.
+    """
```
cbornet, line 79: BTW this is surprising. From the [docs](https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-comment.html) we could expect the comment to be set to NULL.
But that's what I get when I test on databricks cloud.

src/databricks/sqlalchemy/_information_schema.py
```diff
@@ -0,0 +1,23 @@
...
+    Column("commit_action", String),
+    Column("table_owner", String),
+    Column("comment", String),
+    Column("created", DateTime),
+    Column("created_by", String),
+    Column("last_altered", DateTime),
+    Column("last_altered_by", String),
+    Column("data_source_format", String),
+    Column("storage_sub_directory", String),
+    schema="information_schema",
```
cbornet, line 22: Is it OK to use the `information_schema ` ?
From the [docs](https://docs.databricks.com/en/sql/language-manual/sql-ref-information-schema.html), I understand that it's only available when the Unity catalog is used ?

cbornet: I did some tests and it seems that, indeed, the `information_schema` is not available for non-unity workspaces.
I tried to use `SHOW TABLE EXTENDED` instead but the query seems very slow (~10s to get a result...).
So what should be done ? Fallback to STE if `information_schema` returns an error ? Or let `get_multi_table_comment` crash for non-unity workspaces (reflection will catch the error and just not return any comment) ?

cbornet: OK. So it appears other methods (`get_foreign_keys`, `get_pk_constraint`) use individual DTE statements.
So I'll do the same for table comments.
There is probably optimisations possible to try to use the `information_schema` when available and caching the results between comments/PK/FK queries but I'll leave that for potential future work.

cbornet: done

nagyryan:
@susodapop can we get help taking a peek at this one?

NadirJ:
Hello @nagyryan @susodapop @yansonggao-db @andrefurlan-db

NadirJ:
Please let @cbornet or I know if there is anything we can do move this along

src/databricks/sqlalchemy/test/_extra.py
susodapop: I adapted for this in #328 by simply coercing the `''` value into a `NoneType`. This passes SQLAlchemy's reusable tests.

src/databricks/sqlalchemy/_ddl.py
```diff
@@ -61,7 +67,7 @@ def get_column_specification(self, column, **kwargs):
         feature in the future, similar to the Microsoft SQL Server dialect.
         """
         if column is column.table._autoincrement_column or column.autoincrement is True:
-            logger.warn(
+            logger.warning(
```
susodapop, line 70: Good catch 🙏

src/databricks/sqlalchemy/pytest.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
 markers = 
-    reviewed: Test case has been reviewed by databricks
\ No newline at end of file
+    reviewed: Test case has been reviewed by databricks
```
susodapop, line 3: Not sure what change was needed here.

src/databricks/sqlalchemy/test/_extra.py
```diff
@@ -63,8 +66,39 @@ class DateTimeTZTestCustom(_DateFixture, fixtures.TablesTest):
...
+            "test_table",
+            self.metadata,
+            Column("id", Integer, primary_key=True, autoincrement=False),
+            Column("data", String(50)),
+            schema=schema,
+        )
+
+    @requirements.comment_reflection
+    @util.provide_metadata
+    def test_drop_table_comment(self, connection):
```
susodapop, line 94: Since I merged #328 do we still need to override this test?

src/databricks/sqlalchemy/test/_future.py
```diff
@@ -251,36 +249,7 @@ class FutureWeCanSetDefaultSchemaWEventsTest(FutureWeCanSetDefaultSchemaWEventsT
     pass
 
 
-class FutureTableDDLTest(FutureTableDDLTest):
```
susodapop, line 254: Perfect!

src/databricks/sqlalchemy/test/_unsupported.py
```diff
@@ -55,6 +55,7 @@ class SkipReason(Enum):
     TIMEZONE_OPT = "timezone-optional TIMESTAMP fields"
     TRANSACTIONS = "transactions"
     UNIQUE = "UNIQUE constraints"
+    DROP_TBL = "drop table comment"
```
susodapop, line 58: Since `''` is identical to `None` as far as sqlalchemy is concerned, I think we needn't list this as unsupported. We can just document the behaviour. Thoughts?

susodapop (commented):
Overall the change code looks great. I'm running all the tests and going to add some unit tests for the statement rendering.

src/databricks/sqlalchemy/base.py
```diff
@@ -375,6 +374,30 @@ def get_schema_names(self, connection, **kw):
...
+        table_name: str,
+        schema: Optional[str] = None,
+        **kw: Any,
+    ) -> ReflectedTableComment:
+        result = self._describe_table_extended(
+            connection=connection,
+            table_name=table_name,
+            schema_name=schema,
+        )
+        # Type ignore is because mypy knows that self._describe_table_extended *can*
```
susodapop, line 390: We can actually catch this using `isinstance` to avoid the mypy error like we do in `get_pk_constraint`.

src/databricks/sqlalchemy/pytest.ini
susodapop: I reverted.

susodapop:
I pushed a few commits making these changes:

1. I removed the override for `TableDDLTest` since the OG test now passes.
2. I added unit tests of the generated SQL for table comments
3. I moved the logic for parsing table comments out of a DTE result into `_parse.py` so the logic is consistent.
4. Added unit tests for item 2
5. Added an e2e test for using `Inspector` to get a table comment. This is technically a duplicate of portions of `TableDDLTest` but I needed it anyway for debugging the new method in `_parse.py` so might as well keep it.

I concur with your choice not to use the thrift `GetTables` RPC for this. Eventually we will re-implement a lot of these methods using `information_schema` but for now, the output of DESCRIBE TABLE EXTENDED is the path forward.

susodapop:
@cbornet 

Alrighty, I have merged this PR into the staging branch and then opened https://github.com/databricks/databricks-sql-python/pull/329 to merge the staging branch into `main`. I would appreciate you giving that PR a look to make sure everything has been handled to your satisfaction.

src/databricks/sqlalchemy/test/_extra.py
cbornet: Nice 👍

cbornet:
Awesome !
Thanks for your review and enhancements (cool solution for the `''` comment!).

src/databricks/sqlalchemy/test_local/e2e/test_basic.py
```diff
@@ -499,3 +499,45 @@ def test_get_columns(db_engine, sample_table: str):
...
+        yield tbl
+
+        md.drop_all(bind=engine)
+
+    def test_table_comment_reflection(self, inspector: Inspector, table: Table):
+        tbl_name = table.name
+
+        comment = inspector.get_table_comment(tbl_name)
+
+        assert comment == {"text": "table comment"}
```
cbornet, line 543: Could also check column comments with `inspector.get_columns` ?

PR #310, opened 2024-01-02
[PECO-1263] Initial execute_async implementation

susodapop:
## Description

This is a work in progress, but an initial working version of our execute_async interface. The code path for executing and handling thrift responses is very java-esque and un-pythonic. In a subsequent PR I'd like to refactor it for cleanliness, but this is mostly a proof of concept.

1. I moved `ResultSet` out of `client.py` into a dedicated module because otherwise we had a circular import between `thrift_backend.py` and `client.py`
2. I added `ae.py` with the `AsyncExecution` object as defined in the design doc tied to this ticket
3. I added a basic test of the API that works end-to-end

All other e2e tests pass, so this doesn't break the existing thrift execution backend.


src/databricks/sql/ae.py
```diff
@@ -0,0 +1,121 @@
...
+        self.status = status
+
+    status: AsyncExecutionStatus
+    query_id: UUID
+
+    def get_result_or_status(self) -> Union["ResultSet", AsyncExecutionStatus]:
+        """Get the result of the async execution. If execution has not completed, return False."""
+
+        if self.status == AsyncExecutionStatus.FINISHED:
+            self._thrift_fetch_result()
```
benc-db, line 72: Advisory: separate status checking from returning results.  It makes consumers' lives more complicated when a method/function can return two very different types.

src/databricks/sql/client.py
```diff
@@ -1,7 +1,7 @@
 import decimal
 import json
 import os
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, TYPE_CHECKING
```
benc-db, line 4: Why are we doing this pattern?

benc-db (approved):
Approved with suggestions

src/databricks/sql/client.py
susodapop: [This is strictly for type checking](https://stackoverflow.com/questions/61545580/how-does-mypy-use-typing-type-checking-to-resolve-the-circular-import-annotation). It allows us to perform circular imports of types across files without needing to reorganise the code for the sole purpose of satisfying mypy.

src/databricks/sql/ae.py
susodapop: Works for me. The combination method was part of the original design but there's nothing sacred about it. In a follow-up PR I'll split this into two separate methods.

src/databricks/sql/client.py
benc-db: I would suggest that this is generally a code smell.  I don't think its worth blocking on, since you are only relying on it to provide types for type checking, but whenever you hit a situation like this, it's worth considering if there is another organization scheme that brings related types closer together, or structuring the code into layers with a single dependency direction.

src/databricks/sql/ae.py
benc-db: Another option is to have a more structured type:

```
@dataclass
class ExecutionResult:
   status: AsyncExecutionStatus
   results: Optional[ResultSet]
```

benc-db: At least going with a dataclass like above means the consumer never has to match again type, only status.

PR #312, opened 2024-01-08
[PECO-1263] Extract asyncexecution from thrift backend

susodapop:
## Description

**Note: This PR should merge _after_ #311 since it depends on the commits in that PR**

To clean up the dependencies between `client.py` → `thrift_backend.py` → `ae.py`, this PR makes it so that `AsyncExecution` knows how to hydrate itself from a `TExecuteStatementResp`.

This way, `thrift_backend.py` doesn't need to import anything from `ae.py`. In theory, `thrift_backend.py` should never care about the pep-249 interface or about the concept of `AsyncExecution`s. There is an exception to this already where thrift_backend doesn't know how to fetch results unless it is provided a `client.Cursor` object, which looks funky. To get around this without making an unneccessary roundtrip to create a new "cursor", I use a `FakeCursor()` object here. Once we refactor thrift_backend result fetching into a separate interface the `FakeCursor` will be removed.

## What's next

I'm going to clean up the user interface for this to separate `get_result_or_status` into separate methods. Also need to add tests for picking up a running execution that _doesn't_ return results immediately.

### DIFF
--- a/src/databricks/sql/ae.py
+++ b/src/databricks/sql/ae.py
@@ -2,6 +2,7 @@ from enum import Enum
 from typing import Optional, Union, TYPE_CHECKING
 from databricks.sql.results import ResultSet
 
+from dataclasses import dataclass
 
 if TYPE_CHECKING:
     from databricks.sql.thrift_backend import ThriftBackend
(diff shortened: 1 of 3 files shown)


benc-db (approved):

susodapop:
Rebased onto the tip of peco-1263-staging after #311 merged.

PR #313, opened 2024-01-08
[PECO-1263] Separate get_status and poll_for_status methods

susodapop:
## Description

Per guidance from @benc-db, I've separated the `get_result_or_status` method into separate calls. The hello world for this interface now looks like this:

```python
        with self.connection() as conn:
            ae = conn.execute_async("select :param `col`", {"param": 1})
            while ae.is_running:
                ae.poll_for_status()
                time.sleep(1)
```

## What's next?

I need to teach AsyncExecution how to pick up a running execution that was started by another thread.

### DIFF
--- a/src/databricks/sql/ae.py
+++ b/src/databricks/sql/ae.py
@@ -104,7 +104,7 @@ class AsyncExecution:
     def poll_for_status(self) -> None:
         """Check the thrift server for the status of this operation and set self.status
 
-        This will result in an error if the operaiton has been canceled or aborted at the server"""
+        This will result in an error if the operation has been canceled or aborted at the server"""
         self._thrift_get_operation_status()
 
     def cancel(self) -> None:
--- a/src/databricks/sql/ae.py
+++ b/src/databricks/sql/ae.py
@@ -15,6 +15,10 @@ from uuid import UUID
 from databricks.sql.thrift_api.TCLIService import ttypes
 
 
+class AsyncExecutionException(Exception):
+    pass
+
+
 @dataclass
 class FakeCursor:
     active_op_handle: Optional[ttypes.TOperationHandle]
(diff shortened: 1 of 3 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -775,10 +775,7 @@ def _create_arrow_table(self, t_row_set, lz4_compressed, schema_bytes, descripti
                 num_rows,
             ) = convert_column_based_set_to_arrow_table(t_row_set.columns, description)
         elif t_row_set.arrowBatches is not None:
-            (
-                arrow_table,
-                num_rows,
-            ) = convert_arrow_based_set_to_arrow_table(
+            (arrow_table, num_rows,) = convert_arrow_based_set_to_arrow_table(
```
susodapop, line 778: This was a `black` reformat unrelated to this PR.

tests/e2e/test_execute_async.py
```diff
@@ -1,50 +1,62 @@
...
+LONG_RUNNING_QUERY = """
 SELECT SUM(A.id - B.id)
 FROM range(1000000000) A CROSS JOIN range(100000000) B 
 GROUP BY (A.id - B.id)
 """
 
+
 class TestExecuteAsync(PySQLPytestTestCase):
+    @pytest.fixture
+    def long_running_ae(self, scope="function") -> AsyncExecution:
```
susodapop, line 20: For simplicity I added this pytest fixture so I don't need to duplicate this code for my "raises an exception" assertions.

src/databricks/sql/ae.py
```diff
@@ -83,18 +86,26 @@ def __init__(
...
             return self._result_set
-        else:
-            self._thrift_get_operation_status()
-            return self.status
+
+    def poll_for_status(self) -> None:
+        """Check the thrift server for the status of this operation and set self.status
+
+        This will result in an error if the operation has been canceled or aborted at the server"""
+        self._thrift_get_operation_status()
```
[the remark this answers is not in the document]
benc-db: My instinct is yes, just rename the other method.  The more nuanced answer is, it depends on if you think the private version of the method will evolve in ways that the public version could be refactored to adapt for without changing the interface.  The benefit of a private method is that you should feel comfortable changing private interfaces without breaking consumers...however python doesn't really enforce private, so the waters get muddied.  So yeah, I say just rename the existing method.

benc-db (approved):

PR #314, opened 2024-01-09
[PECO-1263] Add get_async_execution method

susodapop:
## Description

This PR adds a way to "pick up" a running execution using only its `query_id` and `query_secret`. The interface looks like this:

```python
with self.connection() as conn:
    query_id = "01eeae6b-xxxx-1513-89a3-4668a032ed77"
    query_secret = "01eeae6b-xxxx-179d-8faf-0f39dc3788f8"
    ae = conn.get_async_execution(query_id, query_secret)
```

To achieve this I had to update `AsyncExecution`'s `__init__` method to accept `None` as its initial `status` value. Prior to this, `AsyncExecution` was always created with the `from_thrift_response` method which already knows the initial value.

I also had to introduce a new `FakeExecuteStatementResponse` dataclass. This is a workaround for the way `thrift_backend.py` is currently written. `thrift_backend.py` was written with the assumption that we have the original `TExecuteStatementResp` from when a query began by the time we fetch results of that query.

But for a "picked up" execution, we don't have the original `TExecuteStatementResp` because we only use the `query_id` and `query_secret` to pick up the running execution. This is a problem because result fetching depends on configuration data present in `TResultSetMetadata`. If we had the original `TExecuteStatementResp` then `thrift_backend.py` knows how to use its properties to gather the configuration it needs to fetch results. But since we don't have the original `TExecuteStatementResp` in the case of a "picked up" execution, we need a way to trick `thrift_backend.py` into thinking otherwise.

For this situation, the Thrift server includes the `TGetResultSetMetadataReq` and `TGetGetResultSetMetadataResp` messages and `thrift_backend.py` helpfully includes a way to invisibly fetch this if the original `TExecuteStatementResp.directResults` property is false-y.

To hook into this, I created the `FakeExecuteStatementResp` which only possesses the properties necessary to make `thrift_backend.py` use its normal code-path. It sets `.directResults=False` and sets `operationHandle` to the `THandleIdentifier` for the current `AsyncExecution`.

As discussed internally, we need to refactor `thrift_backend.py` to not depend on the assumption of synchronous execution within the same thread. But for now, the `FakeExecuteStatmentResp` makes `AsyncExecution` behave exactly the same way in the original thread as it does for a thread where the execution was "picked up". 

## What's next?

After this I need to add documentation for this feature and update the changelog.

### DIFF
--- a/src/databricks/sql/ae.py
+++ b/src/databricks/sql/ae.py
@@ -1,5 +1,6 @@
 from enum import Enum
 from typing import Optional, Union, TYPE_CHECKING
+from databricks.sql.exc import RequestError
 from databricks.sql.results import ResultSet
 
 from dataclasses import dataclass
@@ -119,7 +120,11 @@ class AsyncExecution:
 
         This will result in an error if the operation has been canceled or aborted at the server"""
 
-        _output = self._thrift_backend._poll_for_status(self.t_operation_handle)
+        try:
+            _output = self._thrift_backend._poll_for_status(self.t_operation_handle)
+        except RequestError as e:
+            if "RESOURCE_DOES_NOT_EXIST" in e.message:
+                raise AsyncExecutionException("Query not found: %s" % self.query_id) from e
         self.status = _toperationstate_to_ae_status(_output.operationState)
 
     def _thrift_fetch_result(self) -> None:
--- a/src/databricks/sql/ae.py
+++ b/src/databricks/sql/ae.py
@@ -73,7 +73,7 @@ class AsyncExecution:
         connection: "Connection",
         query_id: UUID,
         query_secret: UUID,
-        status: AsyncExecutionStatus,
+        status: Optional[AsyncExecutionStatus] = None,
         execute_statement_response: Optional[ttypes.TExecuteStatementResp] = None,
     ):
         self._connection = connection
@@ -82,6 +82,9 @@ class AsyncExecution:
         self.query_id = query_id
         self.query_secret = query_secret
         self.status = status
+
+        if self.status is None:
+            self.poll_for_status()
 
     status: AsyncExecutionStatus
     query_id: UUID
--- a/src/databricks/sql/ae.py
+++ b/src/databricks/sql/ae.py
@@ -117,7 +117,7 @@ class AsyncExecution:
 
     def serialize(self) -> str:
         """Return a string representing the query_id and secret of this AsyncExecution.
-        
+
         Use this to preserve a reference to the query_id and query_secret."""
         return f"{self.query_id}:{self.query_secret}"
 
@@ -130,7 +130,10 @@ class AsyncExecution:
             _output = self._thrift_backend._poll_for_status(self.t_operation_handle)
         except RequestError as e:
             if "RESOURCE_DOES_NOT_EXIST" in e.message:
-                raise AsyncExecutionException("Query not found: %s" % self.query_id) from e
+                raise AsyncExecutionException(
+                    "Query not found: %s. Result may have already been fetched."
+                    % self.query_id
+                ) from e
         self.status = _toperationstate_to_ae_status(_output.operationState)
 
     def _thrift_fetch_result(self) -> None:
--- a/src/databricks/sql/ae.py
+++ b/src/databricks/sql/ae.py
@@ -25,6 +25,12 @@ class FakeCursor:
     active_op_handle: Optional[ttypes.TOperationHandle]
 
 
+@dataclass
+class FakeExecuteStatementResponse:
+    directResults: bool
+    operationHandle: ttypes.TOperationHandle
+
+
 class AsyncExecutionStatus(Enum):
     """An enum that represents the status of an async execution"""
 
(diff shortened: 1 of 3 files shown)


tests/e2e/test_execute_async.py
```diff
@@ -53,10 +83,112 @@ def test_cant_get_results_after_cancel(self, long_running_ae: AsyncExecution):
...
+        assert poll_count > 0
+
+    def test_get_async_execution_and_get_results_without_direct_results(
+        self, long_ish_ae: AsyncExecution
+    ):
+        while long_ish_ae.is_running:
+            time.sleep(1)
+            long_ish_ae.poll_for_status()
 
-    def test_staging_operation(self):
```
susodapop, line 57: This diff looks wacky here. I'm removing the `test_staging_operation` for the moment.

tests/e2e/test_execute_async.py
```diff
@@ -53,10 +83,112 @@ def test_cant_get_results_after_cancel(self, long_running_ae: AsyncExecution):
...
+            ae = conn.execute_async(DIRECT_RESULTS_QUERY, {"param": 1})
+            query_id, query_secret = ae.serialize().split(":")
+            ae.get_result()
+
+        with self.connection() as conn:
+            with pytest.raises(AsyncExecutionException, match="Query not found"):
+                ae_late = conn.get_async_execution(query_id, query_secret)
+
+    def test_get_async_execution_and_fetch_results(self, long_ish_ae: AsyncExecution):
+        """This tests currently _fails_ because of how result fetching is factored.
```
susodapop, line 174: ope, need to fix this comment as it's not accurate anymore. this test passes.

tests/e2e/test_execute_async.py
```diff
@@ -53,10 +83,112 @@ def test_cant_get_results_after_cancel(self, long_running_ae: AsyncExecution):
...
+    def test_serialize(self, long_running_ae: AsyncExecution):
+        serialized = long_running_ae.serialize()
+        query_id, query_secret = serialized.split(":")
+
+        with self.connection() as conn:
+            ae = conn.get_async_execution(query_id, query_secret)
+            assert ae.is_running
+
+    def test_get_async_execution_no_results_when_direct_results_were_sent(self):
+        """It remains to be seen whether results can be fetched repeatedly from a "picked up" execution.
```
susodapop, line 161: self nit: will remove this comoment. results _can_ be fetched repeatedly until they've been expired by the thrift server.

src/databricks/sql/ae.py
```diff
@@ -83,6 +84,9 @@ def __init__(
         self.query_secret = query_secret
         self.status = status
 
+        if self.status is None:
```
benc-db, line 87: I don't love having potentially long running ops in the init.  For one thing, it means one more thing that must be mocked when unit testing, regardless of whether the test needs to interact with status directly.  When I have initialization that is non-trivial, but must be complete for an object to be functional, I tend to put that in a factory method, and try to make the constructor hidden...not sure if we have that capability in python though.  Take this comment with a grain of salt, because depending on the user experience, this point may be outweighed by trying to make the client experience simpler.

src/databricks/sql/ae.py
```diff
@@ -111,12 +115,25 @@ def _thrift_cancel_operation(self) -> None:
...
+        Use this to preserve a reference to the query_id and query_secret."""
+        return f"{self.query_id}:{self.query_secret}"
+
     def poll_for_status(self) -> None:
         """Check the thrift server for the status of this operation and set self.status
 
         This will result in an error if the operation has been canceled or aborted at the server"""
 
-        _output = self._thrift_backend._poll_for_status(self.t_operation_handle)
+        try:
```
benc-db, line 129: Perfect example of why I don't love complexity in my constructors.  An ideal constructor says, if you give me all the required parameters as input, you will get an instance of this object as object; here, however, our constructor could fail and throw an exception to the user.  If you instead use a factory pattern, you have the choice of propagating the error, or just giving the user None, as you could name the method something like 'get_if_exists(...)'.

src/databricks/sql/ae.py
susodapop: How about this:

I'll add a new `AsyncExecutionStatus.UNKNOWN` and make that the default.  Then modify the `.status` class member to become a property that accesses a private `._status` member. If `._status==AsyncExecutionStatus.UNKNOWN`, this will fire-off the `poll_for_status` and set it.

This way, the long-running op won't happen until a user actually tries to do something which depends on the status.

src/databricks/sql/ae.py
susodapop: That makes sense to me.

I _do_ have a factory function. It's the `Connection.get_async_execution()` method. I can push this "does it exist" checking into that function like you describe.

src/databricks/sql/ae.py
benc-db: is there a reason to have it pretend to be a property?  As opposed to just get_status()?  This might be a result of other languages I've programmed in, but in my mind, a property is ideally either a.) a field, or b.) something directly computable from fields.  Making it a function suggests that work will be done to get the value.

susodapop: I have no objection to that. Either way is technically "Pythonic".

benc-db (approved):
Approve with suggestions

susodapop:
Thanks! Will apply these suggestions and merge tomorrow. Documentation to follow.

susodapop:
Running our e2e and unit tests before mergin.

susodapop:
Tests all pass. Merging.

PR #315, opened 2024-01-10
Added support for tinyint to _parse.py

TimTheinAtTabs:
Using SqlAlchemy to reflect on a table having a tinyint column throws an error because tinyint type was missing from the type map. Adding a single line to the code seems to have fixed the bug.

### DIFF
--- a/src/databricks/sqlalchemy/_parse.py
+++ b/src/databricks/sqlalchemy/_parse.py
@@ -282,6 +282,7 @@ def get_pk_strings_from_dte_output(
 GET_COLUMNS_TYPE_MAP = {
     "boolean": sqlalchemy.types.Boolean,
     "smallint": sqlalchemy.types.SmallInteger,
+    "tinyint": sqlalchemy.types.SmallInteger,
     "int": sqlalchemy.types.Integer,
     "bigint": sqlalchemy.types.BigInteger,
     "float": sqlalchemy.types.Float,
--- a/src/databricks/sqlalchemy/_parse.py
+++ b/src/databricks/sqlalchemy/_parse.py
@@ -282,7 +282,7 @@ def get_pk_strings_from_dte_output(
 GET_COLUMNS_TYPE_MAP = {
     "boolean": sqlalchemy.types.Boolean,
     "smallint": sqlalchemy.types.SmallInteger,
-    "tinyint": sqlalchemy.types.SmallInteger,
+    "tinyint": type_overrides.TINYINT,
     "int": sqlalchemy.types.Integer,
     "bigint": sqlalchemy.types.BigInteger,
     "float": sqlalchemy.types.Float,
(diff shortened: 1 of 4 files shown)


susodapop (approved):
Thank you! We'll need to add a couple tests for this behaviour before merging.

susodapop:
Please push an update to your commit including the DCO sign-off as described [here](https://github.com/databricks/databricks-sql-python/blob/main/CONTRIBUTING.md#sign-your-work).

src/databricks/sqlalchemy/_parse.py
```diff
@@ -282,6 +282,7 @@ def get_pk_strings_from_dte_output(
 GET_COLUMNS_TYPE_MAP = {
     "boolean": sqlalchemy.types.Boolean,
     "smallint": sqlalchemy.types.SmallInteger,
+    "tinyint": sqlalchemy.types.SmallInteger,
```
susodapop, line 285: This should be  `type_overrides.TINYINT`, not `SmallInteger`, no?

susodapop (changes requested):
Sorry for the chatter. I clicked the approve button prematurely. One comment for you:

Before we can merge:
1. We need to add e2e / unit tests for this. I can do this for you if you're not able to do so easily.
2. We can't merge this change without a DCO sign-off on your commit. Please amend your commit and push it to your branch.

src/databricks/sqlalchemy/_parse.py
TimTheinAtTabs: I do not know enough of the code to make the judgement call. I used my best guess to figure out the quickest solution that could work.

susodapop: Understood. Databricks has both `SMALLINT` and `TINYINT` types. The code you're editing has to do with column reflection of existing tables. These are mapped to SqlAlchemy types named `SmallInteger` (which is built-in to SQLAlchemy) and `TINYINT` (which is a custom type implemented in our dialect). In Python, a `TINYINT` in our dialect behaves _exactly_ the same as a `SmallInteger` (Python doesn't have any size restrictions for integers). That's why your version of the code _works for your use case_. But we should reflect the type properly back as `type_overrides.TINYINT. Otherwise SQLAlchemy will not properly reflect back the true schema of the table, which will be an issue for people using SQLAlchemy/Alembic to programmatically manage their table schemas.

susodapop:
I've pushed a fix that properly tests this behaviour. Can you update your commit with the DCO sign-off?

src/databricks/sqlalchemy/test_local/test_types.py
```diff
@@ -111,7 +111,7 @@ def test_numeric_renders_as_decimal_with_precision(self):
         )
 
     def test_numeric_renders_as_decimal_with_precision_and_scale(self):
-        return self._assert_compiled_value_explicit(
+        self._assert_compiled_value_explicit(
```
susodapop, line 114: I found during testing that this test case would always pass because prefixing an assertion with `return` always evaluates as a pass. When I removed the `return`, the test failed for TINYINT types (should have been caught during initial development). This test now passes.

benc-db (approved):

PR #319, opened 2024-01-14
Customization to allow customization of OAuthManager to use another logic to open the login screen in browser

vaimdev:
As mentioned above


susodapop (commented):
I'm closing this pull request.

Given that we already support providing custom credentials providers (example [here](https://github.com/databricks/databricks-sql-python/blob/main/examples/custom_cred_provider.py)) we would rather not expose a second way to do the same thing.

If you have any issues constructing your own credentials provider you're welcome to open an issue on this repository or speak with Databricks support.

susodapop:
Given that we already

Issue #318, opened 2024-01-12
Add support for proxies per-connection that override environment variables


jakewski:
For cases where we might want a dedicated proxy for the Databricks SQL connector, but not override the global proxy in an application

I have a branch that implements this, just need permissions to push it

Issue #317, opened 2024-01-12
tests/unit/tests.py has broken tests and does not run in CI


jakewski:
pretty much the title

```======================================================================================== short test summary info ========================================================================================
FAILED tests/unit/tests.py::ClientTestSuite::test_auth_args - ValueError: too many values to unpack (expected 4)
FAILED tests/unit/tests.py::ClientTestSuite::test_authtoken_passthrough - AssertionError: ('Authorization', 'Bearer tok') not found in [('User-Agent', 'PyDatabricksSqlConnector/3.0.1')]
FAILED tests/unit/tests.py::ClientTestSuite::test_close_uses_the_correct_session_id - AttributeError: 'bytes' object has no attribute 'sessionHandle'
FAILED tests/unit/tests.py::ClientTestSuite::test_closing_connection_closes_commands - databricks.sql.exc.Error: You must provide at least one staging_allowed_local_path when initialising a connection to perform ingestion commands
FAILED tests/unit/tests.py::ClientTestSuite::test_context_manager_closes_connection - AttributeError: 'bytes' object has no attribute 'sessionHandle'
FAILED tests/unit/tests.py::ClientTestSuite::test_cursor_keeps_connection_alive - AttributeError: 'bytes' object has no attribute 'sessionHandle'
FAILED tests/unit/tests.py::ClientTestSuite::test_disable_pandas_respected - databricks.sql.exc.Error: You must provide at least one staging_allowed_local_path when initialising a connection to perform ingestion commands
FAILED tests/unit/tests.py::ClientTestSuite::test_execute_parameter_passthrough - databricks.sql.exc.Error: You must provide at least one staging_allowed_local_path when initialising a connection to perform ingestion commands
FAILED tests/unit/tests.py::ClientTestSuite::test_executemany_parameter_passhthrough_and_uses_last_result_set - databricks.sql.exc.Error: You must provide at least one staging_allowed_local_path when initialising a connection to perform ingestion commands
FAILED tests/unit/tests.py::ClientTestSuite::test_executing_multiple_commands_uses_the_most_recent_command - databricks.sql.exc.Error: You must provide at least one staging_allowed_local_path when initialising a connection to perform ingestion commands
FAILED tests/unit/tests.py::ClientTestSuite::test_finalizer_closes_abandoned_connection - AttributeError: 'bytes' object has no attribute 'sessionHandle'
[...]

PR #320, opened 2024-01-16
Support proxy per-connection rather than just environment variables

jakewski:
# Problem

I opened this issue: https://github.com/databricks/databricks-sql-python/issues/318. My use case involves using the Databricks SQL connector in a web application where we may not want to set the proxy for all global traffic. The current method of using environment variables would not allow that.

# Solution

Proxy specification works same way as the Python `requests` module - it will default to whatever is set in the environment variable, but it can manually be overridden for a single request / connection if desired.

### DIFF
--- a/src/databricks/sql/auth/thrift_http_client.py
+++ b/src/databricks/sql/auth/thrift_http_client.py
@@ -1,7 +1,7 @@
 import base64
 import logging
 import urllib.parse
-from typing import Dict, Union
+from typing import Dict, Optional, Union
 
 import six
 import thrift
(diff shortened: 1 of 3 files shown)


jakewski:
I would like to add a test for this in client tests - but seems they don't actually run in CI. Any guidance would be much appreciated! https://github.com/databricks/databricks-sql-python/issues/317

susodapop:
Thanks for this contribution. We have rather poor testing facilities for proxies at present (we need proper CI for this repository -- that's coming soon). Will review and perform some manual testing in the next couple weeks.

jakewski:
> Thanks for this contribution. We have rather poor testing facilities for proxies at present (we need proper CI for this repository -- that's coming soon). Will review and perform some manual testing in the next couple weeks.

sounds good, appreciate the review!

PR #322, opened 2024-01-20
[PECO-1263] Add documentation for execute_async

susodapop:
## Description

This PR adds documentation for the new `execute_async` method.

Note: this is a draft as I realised in the course of writing the doc that we need to add the `.is_available` property. More fullsome PR description incoming once that is done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release History
 
+## 3.1.0 (TBD)
+
+- Add `execute_async` and `get_async_execution` methods to `Connection` 
+
 ## 3.0.1 (2023-12-01)
 
 - Other: updated docstring comment about default parameterization approach (#287)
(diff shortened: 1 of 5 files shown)


dhirschfeld:
Does this method have particular benefits over just running a blocking call in a thread?

susodapop:
Yes, the benefit is to recover a running execution. Imagine you are running a long query (1+ hour) and in the middle, your network cable comes unplugged or your computer runs out of battery. The query continues running on your cluster (incurring cost) but you have no way to recover it or see its results. You can cancel it from the Databricks SQL web UI but apart from that you have no choice but to re-run it and wait another hour for results.

In a second usage scenario, if you deploy the connector in a serverless function or lambda. You can wake the lambda to begin executing the query, commit its handle to state, sleep the lambda for some period of time, and wake up to fetch the results. This isn't possible using threads.

PR #325, opened 2024-01-22
[PECO-1263] Implement an .is_available property for AsyncExecution status

susodapop:
## Description

This PR implements an additional boolean flag on `AsyncExecution` that indicates whether the result of this execution can be fetched from a separate thread.

I determined this is necessary while documenting the new `execute_async()` behaviour. For context, when we send a `TExecuteStatementReq` to the thrift server, there are **two** possible return conditions:

1. If the query completes within five seconds, the resulting `TExecuteStatementResp` will actually include the results! 
2. If the query takes longer than five seconds to complete, the resulting `TExecuteStatementResp` will only include the `query_id` for the query, which we can use to poll for the results until they are ready.

In case 1, the result is sent within the initial `TExecuteStatementResp` _and they cannot be fetched a second time. What this means for users is that if they are going to call `.serialize()` to persist a `query_id` and `secret` for use by another thread, they need a way to verify that the results will actually be available to another thread.

This new `.is_available` property makes that easier.

### DIFF
--- a/src/databricks/sql/ae.py
+++ b/src/databricks/sql/ae.py
@@ -1,7 +1,7 @@
 from enum import Enum
 from typing import Optional, Union, TYPE_CHECKING
 from databricks.sql.exc import RequestError
-from databricks.sql.results import ResultSet
+from databricks.sql.results import ResultSet, execute_response_contains_direct_results
 
 from datetime import datetime
 
@@ -81,6 +81,7 @@ class AsyncExecution:
     ]
     _last_sync_timestamp: Optional[datetime] = None
     _result_set: Optional["ResultSet"] = None
+    _is_available: bool = True
 
     def __init__(
         self,
@@ -101,6 +102,8 @@ class AsyncExecution:
 
         if execute_statement_response:
             self._execute_statement_response = execute_statement_response
+            if execute_response_contains_direct_results(execute_statement_response):
+                self._is_available = False
         else:
             self._execute_statement_response = FakeExecuteStatementResponse(
                 directResults=False, operationHandle=self.t_operation_handle
@@ -224,6 +227,15 @@ class AsyncExecution:
     def last_sync_timestamp(self) -> Optional[datetime]:
         """The timestamp of the last time self.status was synced with the server"""
         return self._last_sync_timestamp
+
+    @property
+    def is_available(self) -> bool:
+        """Indicates whether the result of this query can be fetched from a separate thread.
+
+        Only returns False if the query returned its results directly when `execute_async` was called.
+        """
+
+        return self._is_available
 
     @classmethod
     def from_thrift_response(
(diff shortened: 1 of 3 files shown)


benc-db:
>If the query completes within five seconds, the resulting TExecuteStatementResp will actually include the results!
If the query takes longer than five seconds to complete, the resulting TExecuteStatementResp will only include the query_id for the query, which we can use to poll for the results until they are ready.

Oh man, why?  This is really unfortunate behavior.

src/databricks/sql/ae.py
```diff
@@ -225,6 +228,15 @@ def last_sync_timestamp(self) -> Optional[datetime]:
         """The timestamp of the last time self.status was synced with the server"""
         return self._last_sync_timestamp
 
+    @property
+    def is_available(self) -> bool:
```
benc-db, line 232: How long is it available for?  How does this limitation work in the CUJ of recovering from a user-space crash?

susodapop: I'm going to check with the thrift server folks to get that answer. I believe it's available for a few hours. With cloud fetch enabled it's 24 hours.

What do you think of the name `is_available`? I'd like a name that's reflective but also easy enough to type.

benc-db: Its really a tough one; I think this might be a little misleading, because isn't the answer 'Unknown' for crash recovery CUJ?

susodapop: I don't think so. Because under this situation, if the client crashes it will do so _before_ the `query_id` is returned. Which means there would be no way to pick up the execution anyway.

benc-db: Ah, so if it gets a query_id, then it must be available, unless its been a long time?  Although, again, available doesn't sound quite right because the query could still be executing.  Maybe just returned_as_direct_result?

susodapop: Good idea. I've done the rename and pushed.

Note: we could get around this by allowing users to disable `directResults` when the `TExecuteStatementReq` is emitted.

jadewang-db:
> is really unfortunate behavior.

will this cause any issues?

benc-db:
> > is really unfortunate behavior.
> 
> will this cause any issues?

Jade, this is the issue:
>In case 1, the result is sent within the initial TExecuteStatementResp and they cannot be fetched a second time. What this means for users is that if they are going to call .serialize() to persist a query_id and secret for use by another thread, they need a way to verify that the results will actually be available to another thread.

src/databricks/sql/ae.py
```diff
@@ -81,6 +81,7 @@ class AsyncExecution:
     ]
     _last_sync_timestamp: Optional[datetime] = None
     _result_set: Optional["ResultSet"] = None
+    _returned_as_direct_result: bool = False
```
jadewang-db, line 84: do we need change get_result method to also check this flag?

PR #327, opened 2024-01-23
[PECO-1414] Change AAD openid config url to Databricks workspace

jackyhu-db:
To make pysql work with Azure AD OAuth and Databricks OAuth in Azure, we have to use Databricks workspace OAuth well-known URL which works with both.


susodapop:
Please add a record to the changelog.

jackyhu-db:
this requirement is not required any more.

Issue #326, opened 2024-01-23
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'


rahul-dwiwedi:
Getting "TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'" while executing query on databricks.

**Environment:**
OS: Mac
Python: 3.9.6

**Requirements:**
```
certifi==2023.11.17  
charset-normalizer==3.3.2
databricks-sql-connector==3.0.1
et-xmlfile==1.1.0
idna==3.6
lz4==4.3.3
numpy==1.26.3
oauthlib==3.2.2
openpyxl==3.1.2
pandas==2.2.0
pyarrow==14.0.2
python-dateutil==2.8.2
python-dotenv==1.0.0
pytz==2023.3.post1
requests==2.31.0
six==1.16.0
SQLAlchemy==2.0.25
thrift==0.16.0
typing_extensions==4.9.0
tzdata==2023.4
urllib3==2.1.0
```

**Traceback:**
```
Traceback (most recent call last):
  File "/Users/rahul/Desktop/Projects/Data-Access-RND/databricks/connect2.py", line 28, in <module>
    data = result.fetchall()
  File "/Users/rahul/Desktop/Projects/Data-Access-RND/databricks/venv/lib/python3.9/site-packages/sqlalchemy/engine/result.py", line 1325, in fetchall
    return self._allrows()
  File "/Users/rahul/Desktop/Projects/Data-Access-RND/databricks/venv/lib/python3.9/site-packages/sqlalchemy/engine/result.py", line 551, in _allrows
    rows = self._fetchall_impl()
  File "/Users/rahul/Desktop/Projects/Data-Access-RND/databricks/venv/lib/python3.9/site-packages/sqlalchemy/engine/cursor.py", line 2103, in _fetchall_impl
    return self.cursor_strategy.fetchall(self, self.cursor)
  File "/Users/rahul/Desktop/Projects/Data-Access-RND/databricks/venv/lib/python3.9/site-packages/sqlalchemy/engine/cursor.py", line 1139, in fetchall
    self.handle_exception(result, dbapi_cursor, e)
  File "/Users/rahul/Desktop/Projects/Data-Access-RND/databricks/venv/lib/python3.9/site-packages/sqlalchemy/engine/cursor.py", line 1080, in handle_exception
    result.connection._handle_dbapi_exception(
[...]

Issue #203, opened 2023-08-18
Can not use the urllib3 v1: ImportError: cannot import name 'BaseHTTPResponse' from 'urllib3' 


trancongtuanmanh:
Upgrading to the new version of the databricks package introduces the following error:

ImportError: cannot import name 'BaseHTTPResponse' from 'urllib3'

Environment: 
- DBR 12.2LTS 
- dbt-databricks==1.4.2
- python3.9

trancongtuanmanh:
I have a way to work around, try to install urllib3>=2.0.0, but this is not a right way to solve this problem.

susodapop:
This was closed already by https://github.com/databricks/databricks-sql-python/pull/197. Install databricks-sql-connector==2.9.2

trancongtuanmanh:
> This was closed already by #197. Install databricks-sql-connector==2.9.2

Thank you ! It is ok at 1.26.9 version

Issue #190, opened 2023-08-10
Add urllib3 as an explicit dependency


jacobus-herman:
Hi!

Can you please add `urllib3` as an explicit dependency to your package? I see that in the latest version, i.e. `2.9.0`, you are using `BaseHTTPResponse` from `urllib3` and this is only available in `urllib3>=2.0.0`. So can you either add this dependency or change the code to accommodate older versions of `urllib3`? :)

susodapop:
Thanks for opening this issue! Excellent catch.

dthemg:
An issue with this is that `botocore` requires `urllib3` version 1, you can't have both installed:
```
botocore 1.31.29 requires urllib3<1.27,>=1.25.4, but you have urllib3 2.0.4 which is incompatible
```

susodapop:
A fix for this has merged to main and will release today as part of 2.9.2

susodapop:
Release 2.9.2 is now live on Pypi

PR #330, opened 2024-01-25
Update pyproject.toml to fix issues seen in dbt-databricks

benc-db:
pandas 2.2.0 causes several issues, notably:
```
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```
and urllib < 1.26 gives an error about Retry not having allowed_methods in its `__init__`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,13 +2,13 @@
 
 [[package]]
 name = "alembic"
-version = "1.12.1"
+version = "1.13.1"
 description = "A database migration tool for SQLAlchemy."
 optional = true
-python-versions = ">=3.7"
+python-versions = ">=3.8"
 files = [
-    {file = "alembic-1.12.1-py3-none-any.whl", hash = "sha256:47d52e3dfb03666ed945becb723d6482e52190917fdb47071440cfdba05d92cb"},
-    {file = "alembic-1.12.1.tar.gz", hash = "sha256:bca5877e9678b454706347bc10b97cb7d67f300320fa5c3a94423e8266e2823f"},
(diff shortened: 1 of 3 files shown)


susodapop:
This closes #326

pyproject.toml
```diff
@@ -26,7 +26,7 @@ numpy = [
 sqlalchemy = { version = ">=2.0.21", optional = true }
 openpyxl = "^3.0.10"
 alembic = { version = "^1.0.11", optional = true }
-urllib3 = ">=1.0"
+urllib3 = ">=1.26"
```
susodapop, line 29: My only concern with this pin version is anyone using an older urllib3 than 1.26, as these were the users who voiced issues when we originally released with the version pinned to `urllib3>=2.0.0`.

Looking through past issues, it seems that as long as we still support `urllib3==1.27.0` then the previous issues would still be avoided:

- [evidence 1](https://github.com/databricks/databricks-sql-python/issues/203)
- [evidence 2](https://github.com/databricks/databricks-sql-python/issues/190)

I also checked the default version of `urllib3` in a DBSQL notebook and it's 1.26.14, so I think this change is acceptable.

susodapop (commented):
Rather than re-pin the dependency, we could instead write a version-contingent fix for this that checks the `urllib3` version at runtime rather than locking out older users.

This can't merge until we add the new `poetry.lock` file as well. Instructions for this are within CONTRIBUTING.md (but I'll take care of it for you.

poetry.lock
```diff
@@ -2,13 +2,13 @@
 
 [[package]]
 name = "alembic"
-version = "1.12.1"
+version = "1.13.1"
```
susodapop, line 5: bumped alongside sqlalchemy

susodapop (approved):

susodapop:
Synced with `main` to avoid merge conflicts on the README

PR #331, opened 2024-01-25
[PECO-1435] Restore `tests.py` to the test suite

susodapop:
## Description

This closes https://github.com/databricks/databricks-sql-python/issues/317. Somewhere about 18 months ago we move to from using unittest to pytest as our test runner. These tests were contained in a file called `tests.py` which would not be auto-discovered by `pytest` because it doesn't have the prefix `test_` in its filename.

13 of these tests failed at-first. Because we've done 18 months of development since then and changed some of the internal interfaces that these unit tests exercise. So I've gone through and updated as many of the tests as I could within a couple hours to get them to pass. Now only 2 of them fail.

I chose not to fix those remaining two because they deal with the internal interface for handling arrow batches which I'm about to rewrite anyway. I'll pick these up in that effort next week.

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -7,6 +7,8 @@ import itertools
 from decimal import Decimal
 from datetime import datetime, date
 
+from databricks.sql.thrift_api.TCLIService.ttypes import TOpenSessionResp
+
 
 import databricks.sql
 import databricks.sql.client as client
@@ -33,13 +35,16 @@ class ClientTestSuite(unittest.TestCase):
     @patch("%s.client.ThriftBackend" % PACKAGE_NAME)
     def test_close_uses_the_correct_session_id(self, mock_client_class):
         instance = mock_client_class.return_value
-        instance.open_session.return_value = b'\x22'
+
+        mock_open_session_resp = MagicMock(spec=TOpenSessionResp)()
+        mock_open_session_resp.sessionHandle.sessionId = b'\x22'
+        instance.open_session.return_value = mock_open_session_resp
 
         connection = databricks.sql.connect(**self.DUMMY_CONNECTION_ARGS)
         connection.close()
 
         # Check the close session request has an id of x22
-        close_session_id = instance.close_session.call_args[0][0]
+        close_session_id = instance.close_session.call_args[0][0].sessionId
         self.assertEqual(close_session_id, b'\x22')
 
     @patch("%s.client.ThriftBackend" % PACKAGE_NAME)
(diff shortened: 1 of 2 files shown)


tests/unit/test_client.py
```diff
@@ -16,6 +22,51 @@
 from tests.unit.test_thrift_backend import ThriftBackendTestSuite
 from tests.unit.test_arrow_queue import ArrowQueueSuite
 
+class ThriftBackendMockFactory:
+
+    @classmethod
+    def new(cls):
+        ThriftBackendMock = Mock(spec=ThriftBackend)
+        ThriftBackendMock.return_value = ThriftBackendMock
```
benc-db, line 30: What?  A mock that returns itself when invoked?

benc-db (approved):

tests/unit/test_client.py
susodapop: Not quite. This is a mock factory. When you call the `new()` method, it builds a fresh mock and hands it to you.

Originally I just declared the mock once at the top of the file. But there's one test that makes assertions about method call counts, and since the same mock was only instantiated once, that test _passed_ when I ran it alone and _failed_ when I ran the whole suite. So I wrapped it all into a factory and called it a day.

Didn't want to spend too much time on this since I'm confident these are failures due to API changes (that's why I had to start applying `PropertyMock` patches where they weren't previously required.

benc-db: I mean specifically you assigning a Mock to the label `ThriftBackendMock`, and then:
```
ThriftBackendMock.return_value = ThriftBackendMock
```

How does that not make it its own return value?

susodapop: oh whoops, I misunderstood your comment (didn't see the context at first)

Yeah, I thought this was strange as well. Basically our `client` isn't consistent about whether it wants an instantiated `ThriftBackend` or not (I'll be fixing that in the next few weeks). And our `ThriftBackend` isn't a "good" Python object in the sense that it has a bunch of properties that are defined at runtime but can't be inspected by the `unittest.mock` library before running an individual test.

In a perfect world, we should be able to call `MagicMock(autospec=ThriftBackend)` and it will give us a mock that, when called, returns an "instantiated" instance of itself. But autospec won't work because of those unpublished properties. So I had to code that logic myself on this line.

benc-db (approved):

tests/unit/test_client.py
susodapop: tl;dr if we don't include this line, calling `ThriftBackendMock()` will return a new, generic `Mock()` object rather than an instance of our true mock. So I include this logic to make sure that whether you call `ThriftBackend.execute_command()` or `ThriftBackend().execute_command()` you still get the `MockTExecuteResponse()`.

PR #332, opened 2024-01-26
[PECO-1436] [sqlalchemy] Include sqlalchemy __version__ in user-agent

susodapop:
## Description

This pr updates our user-agent setting logic for the sqlalchemy dialect so that we include the sqlalchemy package version. 

I used this as an opportunity to reorganise the two (now three) user agent tests into their own test class.


benc-db (approved):

susodapop:
I'm double-checking the correct format. Will hold off on merging this until next week.

PR #341, opened 2024-02-01
Redact the URL query parameters from the urllib3.connectionpool logs

mkazia-db:
If LogLevel is set to DEBUG, urllib3 is logging all the request URLs. These URLs include pre-signed URLs which are sensitive and should not be logged. This implements a redacting filter for logging that will replace param values in the URL with <REDACTED>. The URLs will be logged with URL path and param names and without the original param values.

### DIFF
--- a/src/databricks/sql/__init__.py
+++ b/src/databricks/sql/__init__.py
@@ -25,7 +25,11 @@ class RedactUrlQueryParamsFilter(logging.Filter):
         record.msg = self.redact(str(record.msg))
         if isinstance(record.args, dict):
             for k in record.args.keys():
-                record.args[k] = self.redact(record.args[k])
+                record.args[k] = (
+                    self.redact(record.args[k])
+                    if isinstance(record.arg[k], str)
+                    else record.args[k]
+                )
         else:
             record.args = tuple(
                 (self.redact(arg) if isinstance(arg, str) else arg)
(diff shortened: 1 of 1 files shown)


src/databricks/sql/__init__.py
```diff
@@ -8,6 +8,35 @@
...
+        else:
+            record.args = tuple(
+                (self.redact(arg) if isinstance(arg, str) else arg)
+                for arg in record.args
+            )
+
+        return True
+
+
+logging.getLogger("urllib3.connectionpool").addFilter(RedactUrlQueryParamsFilter())
```
jackyhu-db, line 38: can we enable this via a flag or environment variable?

jackyhu-db (approved):

src/databricks/sql/__init__.py
```diff
@@ -8,6 +8,35 @@
...
+        super().__init__()
+
+    def redact(self, string):
+        return re.sub(self.pattern, self.mask, str(string))
+
+    def filter(self, record):
+        record.msg = self.redact(str(record.msg))
+        if isinstance(record.args, dict):
+            for k in record.args.keys():
+                record.args[k] = self.redact(record.args[k])
```
jackyhu-db, line 28: can you add a check to ensure records.args[k] is a str before redacting it?

src/databricks/sql/__init__.py
mkazia-db: This should be the default behavior, secure by default. Can't rely on developers to enable a flag. Even disabling by flag will be a security issue.

src/databricks/sql/__init__.py
mkazia-db: done, please check.

andrefurlan-db (approved):

PR #352, opened 2024-02-13
Prep for Test Automation

benc-db:
This PR introduces a new workflow for running E2E tests triggered by PRs that targets our test infra.  
It also includes necessary changes to tests in order to work with the new infra.  
For the most part these are minor, with the one exception being the core E2E tests that were previously based on unittest; they have been converted to pytest to simplify the adoption of environment variable-based connection to the test resources.  Unfortunately this means replacing a lot of `self.assertEquals` and similar with tests based on `assert`.

There is one potentially meaningful change in the sql client, where I found a test failing due to the server returning a list of headers, while the code is expecting a json string.  After making this change, the test in question started passing, while no other tests failed, but I'm not certain this change is compatible with old versions of the DBR, so I will update it to be backwards compatible before merging.

### DIFF
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -1,5 +1,5 @@
 name: Code Quality Checks
-on: 
+on:
   push:
     branches:
       - main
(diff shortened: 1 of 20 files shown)


yunbodeng-db (commented):
@andrefurlan-db please review this PR.
@benc-db this is awesome. Let's do this!

yunbodeng-db (approved):

Issue #353, opened 2024-02-15
Proxy authentication not working


amir-haroun:
Hello,

We are trying to use the databricks sql connector behind an https proxy but this is failing with this error:
WARNING:urllib3.connectionpool:Retrying (DatabricksRetryPolicy(total=25, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProxyError('Cannot connect to proxy.', OSError('Tunnel connection failed: 407 Proxy Authentication Required'))': /sql/1.0/warehouses/xxxx

I think that there is an issue with the current implementation.

PR #354, opened 2024-02-15
Update the proxy authentication

amir-haroun:
This PR solves the proxy auth issue reported in #353.
The idea is to use the *proxy_headers* argument of the *ProxyManager* instead of the *Proxy-Authorization* http header according to the [API docs](https://urllib3.readthedocs.io/en/stable/reference/urllib3.poolmanager.html#urllib3.ProxyManager)

### DIFF
--- a/src/databricks/sql/auth/thrift_http_client.py
+++ b/src/databricks/sql/auth/thrift_http_client.py
@@ -14,7 +14,7 @@ from http.client import HTTPResponse
 from io import BytesIO
 
 from urllib3 import HTTPConnectionPool, HTTPSConnectionPool, ProxyManager
-
+from urllib3.util import make_headers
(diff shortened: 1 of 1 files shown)


amir-haroun:
@rcypher-databricks, @arikfr , @yansonggao-db , @andrefurlan-db , @jackyhu-db ,@benc-db 
Can anyone take a look to this PR?

kravets-levko (approved):

kravets-levko:
@amir-haroun I was testing these changes a lot, and the code seems to work fine. I would like to ask you - can you add some tests for this code? Or maybe you can suggest some ideas on how to avoid regressions in future. Thank you!

amir-haroun:
@kravets-levko,
Not sure what can be added as unit tests here (especially for auth...) but we can verify if the proxies are set using `urllib.getproxies().get('http')`

PR #356, opened 2024-02-16
fixes for cloud fetch - part un

andrefurlan-db:
This is a first set of fixes for cloud fetch.
- Throw when failed to download file
- Retry properly while downloading file
- Add a bunch of debug logs
- Added URL redacting
- Prevent thread issues

TODO:
- check rows context to avoid duplicates
- http connection pools for cloud storage, proxies, etc.

Backported to version 2


src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -97,9 +98,11 @@ def get_next_downloaded_file(
             self.download_handlers.pop(idx)
             # Return True upon successful download to continue loop and not force a retry
             return result
-        # Download was not successful for next download item, force a retry
+        # Download was not successful for next download item. Fail
```
yunbodeng-db, line 101: The original code retries immediately. What is the difference?

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -171,3 +202,34 @@ def decompress_data(compressed_data: bytes) -> bytes:
                 uncompressed_data += data
                 start += num_bytes
         return uncompressed_data
+
+
+def http_get_with_retry(url, max_retries=5, backoff_factor=2, download_timeout=60):
```
yunbodeng-db, line 207: In here it does not bother reading retry-after header? Or there is none because this is directly getting from cloud files?

tests/unit/test_download_manager.py
```diff
@@ -137,71 +137,4 @@ def test_find_next_file_index_one_scheduled_next_row_8000(self, mock_submit):
...
-           return_value=False)
-    def test_check_if_download_successful_other_reason(self, mock_is_file_download_successful):
-        manager = self.create_download_manager()
-        handler = downloader.ResultSetDownloadHandler(manager.downloadable_result_settings, self.create_result_link())
-
-        status = manager._check_if_download_successful(handler)
-        mock_is_file_download_successful.assert_called()
-        assert not status
-        assert manager.fetch_need_retry
+        assert manager._find_next_file_index(8000) is None
```
yunbodeng-db, line 140: By deleting the old tests and changing the logic, I think we missed retry coverage?

src/databricks/sql/cloudfetch/download_manager.py
andrefurlan-db: the original code was wrong. Now I am properly retrying.

andrefurlan-db: and most importantly, actually failing. We cannot continue if a file fails to download after the retries

src/databricks/sql/cloudfetch/downloader.py
andrefurlan-db: this is not thrift. Just cloud files.

tests/unit/test_download_manager.py
andrefurlan-db: I removed the old retry logic. No point in testing it now.

src/databricks/sql/cloudfetch/download_manager.py
yunbodeng-db: The original code returns None in case of failure?

yunbodeng-db (approved):

Issue #372, opened 2024-03-15
`databricks.sql.connect` hangs in a long retrying loop when an invalid access token is used


mbelak-dtml:
## Steps to reproduce
Supply any invalid access token to `databricks.sql.connect`, e.g.
```python
sql.connect(
     server_hostname=my_workspace_url,
     http_path=http_path_to_my_warehouse,
     access_token="this-is-not-a-valid-access-token"
)
```

## Current behavior
The call gets stuck in a long retrying loop, eventually failing with a `RequestError`, which does not indicate an authentication error.

## Expected behavior
The function immediately raises an exception indicating an authentication error.

mbelak-dtml:
Related
https://github.com/databricks/databricks-sql-python/issues/346#issuecomment-1967984742

susodapop:
The fix for this has merged.

Issue #346, opened 2024-02-04
Provide actionable feedback when internal operations are failing rather than silently retrying for over 10mins


dhirschfeld:
I'm trying to debug a problem where my request appears to hang indefinitely.

If I pause the program in the debugger it's deep in a retry loop in `urllib`
```python
>>> response
<urllib3.response.HTTPResponse at 0x187b9786ec0>

>>> response.status
403
```
...and it seems the default timeout for requests is 15mins
```python
>>> self
<databricks.sql.auth.thrift_http_client.THttpClient at 0x187b97858a0>

>>> self._THttpClient__timeout
900.0
```

dhirschfeld:
To all intents and purposes the library has hung - there is no feedback that the operation is failing and being retried multiple times or any indication of why it is failing.

Maybe this issue is more of a request for better feedback in the case of failing HTTP operations.

dhirschfeld:
The 403 is pretty fishy.

Debugging a bit further I can see the token being passed in the headers:
```python
>>> headers
{'Content-Type': 'application/x-thrift',
 'Content-Length': '167',
 'User-Agent': 'PyDatabricksSqlConnector/3.0.1 (sqlalchemy + )',
 'Authorization': 'Bearer <snip>d6b6'}
```

The token being passed is the same as my default environment variable:
```powershell
❯ echo $env:DATABRICKS_TOKEN
<snip>d6b6
```

***BUT*** in my calling script I explicitly set the correct token to use:
```python
os.environ['DATABRICKS_TOKEN'] = '<snip>60cd'
```

...so it seems the token is somehow not being picked up?

susodapop:
Do you know what kind of Thrift request is being sent that results in this 403 failing with the 403? Is it an `OpenSession`?

> there is no feedback that the operation is failing and being retried multiple times or any indication of why it is failing.

Is this true even when you have your log level set at `DEBUG`? `urllib3` and `databricks-sql-connector` both emit debug messages when they determine a request is retry-able.

Adding this to your script will help with the repro:

```python
import logging

logging.getLogger("databricks.sql").setLevel(logging.DEBUG)
logging.getLogger("urllib3").setLevel(logging.DEBUG)
```

> The 403 is pretty fishy.

Yeah, generally a 403 should halt everything since authentication errors are by-nature not recoverable. The DBX team should investigate. Likely the fix will be an additional conditional check added to the `should_retry()` method in `DatabricksRetryPolicy`.

[...]

PR #373, opened 2024-03-15
Don't retry network requests that fail with code 403

susodapop:
## Description

This adds a check to `should_retry()` which immediately raises an exception if a 403 code is received from Databricks and an associated test. Prior to this change, the connector would continue retrying until the retry counter was exceeded.

I think this has existed for some time but wasn't obvious to users until exponential backoff was implemented in #349. Because prior to #349, it would retry over and over again until it ran out of retries and then raise an exception. But now that backoff is in-force, it takes longer for it to exhaust the retry counter.

Fixes these:
- https://github.com/databricks/databricks-sql-python/issues/372
- https://github.com/databricks/databricks-sql-python/issues/346

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release History
 
+# x.x.x (TBD)
+
+- Don't retry requests that fail with code 403
+
 # 3.1.0 (2024-02-16)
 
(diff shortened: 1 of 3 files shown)


benc-db:
looks good, running integration tests now.

benc-db (approved):

PR #379, opened 2024-03-26
Fix cookie setting

benc-db:
Small revamp of PR that @susodapop put out in October.  I suspect this might be breaking now (see https://github.com/databricks/dbt-databricks/issues/622)

Also, I removed a test that happens to still pass successfully, even though it shouldn't pass after my previous change to retry behavior.  For whatever reason, the mocking of the status code is not making it into the retry logic.

### DIFF
--- a/src/databricks/sql/auth/thrift_http_client.py
+++ b/src/databricks/sql/auth/thrift_http_client.py
@@ -191,7 +191,8 @@ class THttpClient(thrift.transport.THttpClient.THttpClient):
 
         # Saves the cookie sent by the server response
         if "Set-Cookie" in self.headers:
-            self.setCustomHeaders(dict("Cookie", self.headers["Set-Cookie"]))
+            self._headers["Cookie"] = self.headers["Set-Cookie"]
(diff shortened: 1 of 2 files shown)


kravets-levko:
@benc-db there is a task to remove cookies handling at all - [PECO-1064]. We already removed it in Node connector a while ago, and IIRC Go driver didn't even have similar code. Do you know any use cases where cookies may be needed in PySQL?

benc-db:
> @benc-db there is a task to remove cookies handling at all - [PECO-1064](https://databricks.atlassian.net/browse/PECO-1064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ). We already removed it in Node connector a while ago, and IIRC Go driver didn't even have similar code. Do you know any use cases where cookies may be needed in PySQL?

If we can just delete it, I'll go ahead and do that.

kravets-levko (approved):

Issue #381, opened 2024-03-30
Fixing a couple type problems. (adding py.typed, typing connect, returning Any from fetchall (which I failed to fix!))


wyattscarpenter:
I'm trying to use databricks-sql-python in a statically-typed python project, and I'm running into some issues. I'm using `mypy --strict`.

1. Despite this project having type annotations, my typechecker (mypy) will not check them (this causes problems for the type implications of the rest of my code). I'm pretty sure that, as per https://peps.python.org/pep-0561/#packaging-type-information, you can just add a py.typed file to the sql/ and sqlalchemy/ folders. I don't use the sqlalchemy side so I haven't tested that, but the sql one seems to work perfectly when I modify the copy of databricks-sql-python on my machine.

2. I get `error: Call to untyped function "connect" in typed context  [no-untyped-call]`. I'm pretty sure you can just change the type signature of `def connect` in the sql folder's `__init.py__` to `def connect(server_hostname, http_path, access_token=None, **kwargs)  -> "Connection"`. This works perfectly when I test it on my machine.

3. I get `error: Returning Any from function declared to return "List[Row]"  [no-any-return]` when I have the line ` return cursor.execute(query, sql_params_dict).fetchall()`. Honestly, I have no idea how to fix this. [fetchall](https://github.com/databricks/databricks-sql-python/blob/7fade51097d943c7f0fbe2b43f762bf837656f49/src/databricks/sql/client.py#L927) looks like it has the right type signature to me! Maybe this is a bug in mypy somehow? Or maybe this is a symptom of the previous issue not getting completely fixed...

I can make a pull request about this, at least the first two, but maybe these ideas are bad and you haven't done them yet because it would break something else; let me know!

wyattscarpenter:
I realized error №3 here was not fixed by solving №s 1 and 2, but also I realized what was causing it, leading to my fix in #384

wyattscarpenter:
Since #382 and #384 have been merged, I'm fairly certain this problem is completely fixed. 🎉

PR #382, opened 2024-03-30
Fixing a couple type problems: how I would address most of #381

wyattscarpenter:


benc-db:
What do the empty 'typed' files do?

benc-db:
> What do the empty 'typed' files do?
It looks like they fail our mypy step...do you have any suggestions on how to update the workflow so that this will pass?

wyattscarpenter:
The empty py.typed files are simply a python convention that allow typecheckers to use the type annotations of your package, when typechecking project that depend on your package. (Seems kind of redundant to me, but I didn't make the conventions!) https://peps.python.org/pep-0561/#packaging-type-information

I think probably the mypy step is failing for some other reason, because the error message says
>  error: --install-types failed (no mypy cache directory)
> Error: Process completed with exit code 2.
Which seems unrelated. I'll see if I can replicate this on my machine, later.

Ah, hold on, the unrelated error message is a bug in [mypy](https://github.com/python/mypy/issues/10768), so I will have to replicate this error on my machine, just to see what the actual error is!

There are also unrelated errors above in the script, but quite possibly those are unrelated.

wyattscarpenter:
I think the problems should be fixed now. It had to do with importing the type so that the typechecker could see it, while making sure not to cause a circular import.

benc-db (approved):

benc-db:
Running integration tests, then will merge.

PR #384, opened 2024-04-02
fix the return types of the classes' __enter__ functions

wyattscarpenter:
so that the type information is preserved in context managers eg with-as blocks. This fixes the last issue I was having in #381


benc-db (approved):
Should be a strictly positive change, but running the integration tests just to be safe, then will merge.

wyattscarpenter:
Thanks!

PR #385, opened 2024-04-05
Add explicit heartbeat interval for GetOperationStatus

benc-db:
In my investigation, it looks as though the current interval between GetOperationStatus requests is enforced by the server, as it comes between sending and receiving the request, and turning on debug logging for urllib3 shows no retries.  This PR adds an entry point for configuring how long to wait client side between sending successive GetOperationStatus requests on 200 response.  I'm initially proposing to set this value at 25 seconds, so that when adding to the 5s server-side, we are polling every 30s for statement completion.


jackyhu-db (approved):

src/databricks/sql/thrift_backend.py
```diff
@@ -55,6 +55,7 @@
 
 TIMESTAMP_AS_STRING_CONFIG = "spark.thriftserver.arrowBasedRowSet.timestampAsString"
 DEFAULT_SOCKET_TIMEOUT = float(900)
+DEFAULT_STATEMENT_HEARTBEAT_INTERVAL = float(25)
```
kravets-levko, line 58: For me 25 seconds delay between GetOperationStatus requests looks a bit too much. I mean - if query execution finishes faster than this delay - client will still wait that 25 seconds. In Nodejs we poll for operation status with 100ms interval - which also may be too low, but looks way more reasonable

benc-db: 100ms?  Do you actually see that in practice?  Today we wait 0, and the server holds up for up to 5 seconds.  Seems weird to care about subsecond latency on queries that we know take more than 5 seconds.

PR #397, opened 2024-06-06
Update Thrift package

### LINKED ISSUE #394: Unpin Thrift
Thrift version is currently pinned at `^0.16.0`, Thrift skipped all the way to `0.20.0` this year without any breaking changes in the [release notes](https://github.com/apache/thrift/blob/master/CHANGES.md) related to Python so it seems like it would be safe to upgrade.

m1n0:
Resolves #394


m1n0:
This is a very simple change, could we get workflow approval for tests to run please?

susodapop (commented):
This change LGTM.

kravets-levko (approved):

m1n0:
Thanks for the review and approval! All tests fail on an seemingly unrelated error though, any advice?

susodapop:
@m1n0 interesting! I checked out your branchy and tried the unit tests and everything passed. Looks like something changed on github that results in these failures.

susodapop:
Looks like this is caused by a broader issue in virtualenv that is affecting many other developers: https://github.com/pypa/virtualenv/issues/2735

The solution is to force github actions to rebuild its virtualenvs (rather than caching them). I'm not quite clear how to force that since I can't see the GH Actions panel for this repo anymore. But this should be enough information to fix things, @kravets-levko

susodapop:
Yep, the tests are failing because `actions/cache@v2` is pulling a cached version of the environment which was most recently rebuilt using a version of `filelock` that has since been yanked. [Here](https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache) are the instructions for blowing away this cache so that GH Actions can rebuild it. This will cause the tests to pass.

m1n0:
Thanks everyone!

PR #403, opened 2024-06-30
move py.typed to correct places

wyattscarpenter:
https://peps.python.org/pep-0561/ says 'For namespace packages (see PEP 420), the py.typed file should be in the submodules of the namespace, to avoid conflicts and for clarity.'. Previously, when I added the py.typed file to this project (https://github.com/databricks/databricks-sql-python/pull/382) I was unaware this was a namespace package, so I added the py.typed file to the top-level folder (the namespace). As PEP 561 warns us, this does create conflicts; other libraries in the databricks namespace package (such as, in my case, databricks-vectorsearch) are then treated as though they are typed, which they are not, leading to unnecessary type errors in those other packages. This commit moves the py.typed file to the correct places, the submodule folders, fixing that problem.

### DIFF
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -134,7 +134,7 @@ class CloudFetchQueue(ResultSetQueue):
         schema_bytes,
         max_download_threads: int,
         start_row_offset: int = 0,
-        result_links: List[TSparkArrowResultLink] = None,
+        result_links: List[TSparkArrowResultLink] = [],
         lz4_compressed: bool = True,
         description: List[List[Any]] = None,
     ):
@@ -161,13 +161,12 @@ class CloudFetchQueue(ResultSetQueue):
                 start_row_offset
             )
         )
-        if result_links is not None:
-            for result_link in result_links:
-                logger.debug(
-                    "- start row offset: {}, row count: {}".format(
-                        result_link.startRowOffset, result_link.rowCount
-                    )
+        for result_link in result_links:
+            logger.debug(
+                "- start row offset: {}, row count: {}".format(
+                    result_link.startRowOffset, result_link.rowCount
                 )
+            )
 
         self.download_manager = ResultFileDownloadManager(
             self.max_download_threads, self.lz4_compressed
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -7,7 +7,7 @@ from collections import OrderedDict, namedtuple
 from collections.abc import Iterable
 from decimal import Decimal
 from enum import Enum
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, Union
 import re
 
 import lz4.frame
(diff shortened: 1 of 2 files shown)


kravets-levko (approved):

kravets-levko:
Hi @wyattscarpenter! CI now reports this error: `src/databricks/sql/cloudfetch/downloader.py: error: Source file found twice under different module names: "downloader" and "databricks.sql.cloudfetch.downloader"` - do you have any ideas what causes this?

src/databricks/sql/utils.py
```diff
@@ -134,7 +134,7 @@ def __init__(
         schema_bytes,
         max_download_threads: int,
         start_row_offset: int = 0,
-        result_links: List[TSparkArrowResultLink] = None,
+        result_links: List[TSparkArrowResultLink] = [],
```
kravets-levko, line 137: I'm a bint uncertain about this change. In JS I would just approve it 🙂  But in Python, default values are created once and thus are shared between function calls. So if one day somebody will decide to modify `self.result_links` (or other place where it is passed to) without paying extra attention - it will cause udesired side effects. Maybe better keep a default value `None` and change the type to `Optional[List[TSparkArrowResultLink]]`. WDYT?

wyattscarpenter: Boy, I learn something surprising about Python every day...!

I agree that [] is a bad default value, then. `Optional[List[TSparkArrowResultLink]]` is a fine type for the parameter to have. The logic for dealing with it being None, which I removed, would have to be put back in. (The code right below that code would also, I think, have to be indented, to go within the `if not None` if statement—that code being outside that if statement was the error mypy was originally upset about which caused me to touch this code at all, I think.)

For your consideration: if you want to get rid of the `= None` altogether, you can. There would be no default for that argument, and it would become a required parameter. It would be moved to before start_row_offset (as it would now be a positional parameter). I grepped the example code and didn't see any example usages in which the parameter it was omitted, but I don't know enough about the code to say if it should be optional.

So, either of those options seem fine to me.

wyattscarpenter:
@kravets-levko the "Source file found twice under different module names" error is, ultimately, caused by mypy not dealing with the namespaces correctly. It's either a historical bug or behavior they decided to change later, since it actually works fine without workarounds or flags on the latest version of mypy: https://github.com/wyattscarpenter/databricks-sql-python/tree/upgrade_mypy (I'm thinking I'll do a pr from that branch later, since the version upgrade touches a number of other files and therefore might be inconvenient to review at the same time as this pr; but I could also make squash it into a commit in this pr if you would prefer.)

src/databricks/sql/utils.py
kravets-levko: If you don't mind - let's keep `None` as default value, with all those `if <var> is not None`

wyattscarpenter: I have now done so.

kravets-levko (approved):

kravets-levko:
Thank you @wyattscarpenter!

wyattscarpenter:
Thanks!

PR #405, opened 2024-07-01
[PECO-1751] Refactor CloudFetch downloader: handle files sequentially

kravets-levko:
[PECO-1751]

This PR is an attempt to fix various CloudFetch-related issues (no data returned, duplicated rows returned, also maybe some performance and memory issues).

Replaces databricks/databricks-sql-python#362

#### The problem

In current implementation, all the links from a single TRowSet are added to a concurrent thrrad pool. The pool downloads them in a random order (all the tasks have the same priority and as a result - same chance to be executed first). To maintain the order of results, `startRowOffset`/`rowCount` fields from each CloudFetch link are used: library keeps track of the current row number, and use it to pick the right CloudFetch link (looking for the file where the current row is within [startRowOffset; startRowOffset + rowCount]).

This solution has several caveats. First of all, data can be fetched only from beginning to the end. With a concurrent thread pool, you never know which file will be downloaded first. In the worst case, while the user is waiting for the very first file, the library may download all the other ones and keep them in memory because the user may need them in future. This increases the latency (on average it will be okay, but we have no control over it), and also memory consumption.
Another problem with this approach is that if any of the files cannot be downloaded (even after retries) - there is no need to download the remaining files, the user won’t be able to process them anyway. But because files are downloaded in arbitrary order - nobody knows how many files will be downloaded before the user reaches the failed one.

Also, instead of utilizing the power of Futures, a custom synchronization and error handling is implemented - which may be a cause of synchronization issues. Additionally, retries are implemented outside of the download task, so when the task fails - outer code has to reschedule it, which is also a great source of sporadic errors.

#### The solution

This PR changes CloudFetch downloader to use a queue. Downloader keeps a list of pending links (not scheduled), and current tasks. Number of tasks is limited, so new files are scheduled only when previous task is completed and extracted from queue. As user requests next files, downloader will pick the first task from the queue, and schedule the new one to run in background - to keep the queue full. Then, downloader will wait for the task it picked from the queue, and then return it to user. Tasks are still running in in parallel in background. Also, each task itself is reponsible for handling errors (e.g. retry failed downloads), so when task completes - it is either eventually successfull, or failed after all possible retries. 

With this approach, the proper order of files is automatically assured. All errors are either handled in downloader or propagated to user. If some file cannot be downloaded due to error - library will not download the remaining ones (like it did previously). Utilizing Futures solves all possible issues with threads synchronization. Because new files are downloaded only when user consumes previous ones - library will not keep the whole dataset in memory.

[PECO-1751]: https://databricks.atlassian.net/browse/PECO-1751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/databricks/sql/cloudfetch/download_manager.py
+++ b/src/databricks/sql/cloudfetch/download_manager.py
@@ -1,52 +1,27 @@
 import logging
 
-from concurrent.futures import ThreadPoolExecutor
-from dataclasses import dataclass
+from concurrent.futures import ThreadPoolExecutor, Future
 from typing import List, Union
 
 from databricks.sql.cloudfetch.downloader import (
     ResultSetDownloadHandler,
     DownloadableResultSettings,
+    DownloadedFile,
 )
 from databricks.sql.thrift_api.TCLIService.ttypes import TSparkArrowResultLink
 
 logger = logging.getLogger(__name__)
 
 
-@dataclass
-class DownloadedFile:
-    """
-    Class for the result file and metadata.
-
-    Attributes:
-        file_bytes (bytes): Downloaded file in bytes.
-        start_row_offset (int): The offset of the starting row in relation to the full result.
-        row_count (int): Number of rows the file represents in the result.
-    """
-
-    file_bytes: bytes
-    start_row_offset: int
-    row_count: int
-
-
 class ResultFileDownloadManager:
-    def __init__(self, max_download_threads: int, lz4_compressed: bool):
-        self.download_handlers: List[ResultSetDownloadHandler] = []
-        self.thread_pool = ThreadPoolExecutor(max_workers=max_download_threads + 1)
-        self.downloadable_result_settings = DownloadableResultSettings(lz4_compressed)
-        self.fetch_need_retry = False
-        self.num_consecutive_result_file_download_retries = 0
-
-    def add_file_links(
-        self, t_spark_arrow_result_links: List[TSparkArrowResultLink]
-    ) -> None:
-        """
-        Create download handler for each cloud fetch link.
-
-        Args:
-            t_spark_arrow_result_links: List of cloud fetch links consisting of file URL and metadata.
-        """
-        for link in t_spark_arrow_result_links:
+    def __init__(
+        self,
+        links: List[TSparkArrowResultLink],
+        max_download_threads: int,
+        lz4_compressed: bool,
+    ):
+        self._pending_links: List[TSparkArrowResultLink] = []
+        for link in links:
             if link.rowCount <= 0:
                 continue
             logger.debug(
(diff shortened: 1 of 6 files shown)


src/databricks/sql/cloudfetch/download_manager.py
```diff
@@ -73,143 +52,49 @@ def get_next_downloaded_file(
...
-            return result
-        else:
+        task = self._download_tasks.pop(0)
+        # Future's `result()` method will wait for the call to complete, and return
+        # the value returned by the call. If the call throws an exception - `result()`
+        # will throw the same exception
+        file = task.result()
+        if (next_row_offset < file.start_row_offset) or (
+            next_row_offset > file.start_row_offset + file.row_count
+        ):
```
jackyhu-db, line 71: if `next_row_offset` does not match file's offset, is it intended to return the file?

kravets-levko: Short answer - yes. The idea is that as user iterates through records from beginning to end, we fetch data sequentially, and fail if next portion of data cannot be fetched. So there are never gaps in the data, and even tracking the current row number should be redundant. Eventually, the whole `next_row_offset` thing should be removed, but for now I keep it and this log for debug purposes

jackyhu-db: I see, but this will violate the API contract. If the client request the file/chunk whose offset is different from current downloaded file, they will get the wrong file. If we want to do this as iterator and does not respect offset, we should remove it from the parameter or create another method if we want to backward compatible?

kravets-levko: This class is internal to the library, so I wouldn't worry that much. Library's public interface allows only iterating through rows from beginning to end. So user cannot request an arbitrary row. Here, the previous implementation needed a `next_row_offset` to find the correct batch among the downloaded ones, because previous implementation was downloaded batches concurrently and in a random order. But in fact, all the batches already come in a proper order. So if we make sure they're downloaded sequentially (as in new implementation) - there should be no such case when user tries to get next batch and it doesn't match the current row number. And, as a result, `next_row_offset` should not be needed too. I can remove it right now, but I would like to have this log we're duscussing right now - for debug purposes. And after users confirm that new code works properly - I'm going to completely remove `next_row_offset` in the next version

jackyhu-db: If the `next_row_offset` will not be respected and may return the file does not match, I think we should remove it as even if it is used by the client and they call it but get the wrong result, which is also wrong and breaks the contract.

kravets-levko: @jackyhu-db Yes, I totally agree with you on this. And, as I said, I want to keep it for now for debugging purposes. Currently `next_row_offset` is just passed all the way along from the `ThriftBackend` class and is not used for any other purpose except this log message. After we confirm that users no longer have issues with CloudFetch - I will remove it completely

jackyhu-db: If this is case, can you add a new download manager instead of rewriting existing one and add a flag/option to switch to the new one? Then user can switch it back to the old one if they find some problems.

kravets-levko: Users can disable CloudFetch feature - that's what they already do when they have issues with this feature. This download manager is for CloudFetch only.

Library does not allow to fetch data from arbitrary place, because Thrift server doesn't support so

jackyhu-db (approved):

PR #406, opened 2024-07-02
Upgrade mypy

wyattscarpenter:
This commit removes the flag (and cd step) from https://github.com/databricks/databricks-sql-python/commit/f53aa37a34dc37026d430e71b5e0d1b871bc5ac1 which we added to get mypy to treat namespaces correctly. This was apparently a bug in mypy, or behavior they decided to change. To get the new behavior, we must upgrade mypy. (This also allows us to remove a couple `# type: ignore` comment that are no longer needed.)

This commit runs changes the version of mypy and runs `poetry lock`; note that I did not run `poetry lock --no-update`, because I didn't think about it at the time (I could go do that instead if need be). It also conforms the whitespace of files in this project to the expectations of various tools and standards (namely: removing trailing whitespace as expected by git and enforcing the existence of one and only one newline at the end of a file as expected by unix and github.) It also uses https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade codebase due to a change in mypy behavior. For a similar reason, it also fixes newly-found type (or otherwise) errors:

* "Return type 'Retry' of 'new' incompatible with return type 'DatabricksRetryPolicy' in supertype 'Retry'"
* databricks/sql/auth/retry.py:225: error: object has no attribute update  [attr-defined]
* /test_param_escaper.py:31: DeprecationWarning: invalid escape sequence \) [as it happens, I think it was also wrong for the string not to be raw, because I'm pretty sure it wants all of its backslashed single-quotes to appear literally with the backslashes, which wasn't happening until now]
* ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject [this is like a numpy version thing, which I fixed by being stricter about numpy version]


src/databricks/sql/auth/retry.py
```diff
@@ -222,7 +224,7 @@ def new(self, **urllib3_incremented_counters: typing.Any) -> Retry:
         urllib3_init_params.update(**urllib3_incremented_counters)
 
         # Include urllib3's current state in our __init__ params
-        databricks_init_params["urllib3_kwargs"].update(**urllib3_init_params)  # type: ignore
+        databricks_init_params["urllib3_kwargs"] = urllib3_init_params.copy()
```
kravets-levko, line 227: This modification actually changes the behavior of the code (replacing the whole dict is not the same as adding new/updating existing keys - even if for now it works because initial dict is empty). Let's revert to updating

kravets-levko (changes requested):
Looks good except one notice. Also, I'm curious - where all that formatting changes come from? I mean - they look okay (especially al that trailing whitespaces and blank lines), but why didn't we catch them earlier? 🤔

kravets-levko (approved):

PR #410, opened 2024-07-08
[PECO-1760] Certifi removes GLOBALTRUST root certificate

jprakash-db:
Updated the certify version to "2024.7.4"

[Dependabot report:](https://github.com/databricks/databricks-sql-python/security/dependabot/16)

Package | Affected versions | Patched version
-|-|-
certifi | >= 2021.05.30, < 2024.07.04 | 2024.07.04

Certifi 2024.07.04 removes root certificates from "GLOBALTRUST" from the root store. These are in the process of being removed from Mozilla's trust store.

GLOBALTRUST's root certificates are being removed pursuant to an investigation which identified "long-running and unresolved compliance issues". Conclusions of Mozilla's investigation can be found [here](https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/XpknYMPO8dI).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1202,5 +1202,4 @@ sqlalchemy = ["sqlalchemy"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.8.0"
-content-hash = "983b6405d99b6e4e931868c6eaf9136bcc64a3588d1a5db763358ee5753e4fdf"
-
+content-hash = "bb289639e9f2630940a6699c69905db0166d9ae50afb5485508f48bbf899c77c"


kravets-levko (changes requested):
@jprakash-db Dependency upgrade looks good 👍  But please remove the `.idea` folder from the PR. Also, what are that two files in `src/databricks/sqlalchemy/test/`? Are they relevant to this PR?

jprakash-db:
@kravets-levko Fixed the issues

pyproject.toml
```diff
@@ -27,6 +27,7 @@ sqlalchemy = { version = ">=2.0.21", optional = true }
 openpyxl = "^3.0.10"
 alembic = { version = "^1.0.11", optional = true }
 urllib3 = ">=1.26"
+certifi = "2024.7.4"
```
kravets-levko, line 30: Okay, so I checked where this dependency come from. It is a transitive dependency of `requests` package. Also, I checked the `requests` itself - it's latest version also uses old `certifi`. So I think in this case it's okay to add a consraint to our project and wait for `requests` upgrade. Two more things to do:

1. add a comment that once `requests` updates `certifi` to the version that will satisfy us - we should roll this change back and upgrae `requests instead
2. also, instead of pinning the exact version - better relax the constraint a bit (e.g. "^2024.7.4")

kravets-levko (changes requested):

Issue #413, opened 2024-07-11
Too large queries produce MaxRetryError


rth:
Previously when a too big query was made https://github.com/databricks/databricks-sql-python/issues/383  we got 0 rows as output (as discussed in that issue) . With the changes in https://github.com/databricks/databricks-sql-python/pull/405 for me it now produces an MaxRetryError which is better but the error message is misleading (and also retying so many times is slow).

The minimal code I'm using is,

```py
    from databricks import sql as databricks_sql

    db = databricks_sql.connect(
        server_hostname=os.getenv("DATABRICKS_SERVER_HOSTNAME"),
        http_path=os.getenv("DATABRICKS_HTTP_PATH"),
        access_token=os.getenv("DATABRICKS_TOKEN"),
        _tls_no_verify=True
    )
    cursor = db.cursor()
    cursor.execute("<my-query>")
    data = cursor.fetchall()
```

If the query is small, it works with no warnings. 

If the query is too big  produces the following MaxRetryError with a nested SSLError. There is no way to detect a too big query in the HTTP response status without retying N times and reaching MaxRetryError?  And also I have the impression that _tls_no_verify is not passed somewhere in this case, which produces those SSLError. cc @kravets-levko 

<details>

```
urllib3.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "python3.10/site-packages/requests/adapters.py", line 667, in send
    resp = conn.urlopen(
  File "python3.10/site-packages/urllib3/connectionpool.py", line 873, in urlopen
    return self.urlopen(
  File "python3.10/site-packages/urllib3/connectionpool.py", line 873, in urlopen
    return self.urlopen(
  File "python3.10/site-packages/urllib3/connectionpool.py", line 873, in urlopen
    return self.urlopen(
  [Previous line repeated 2 more times]
  File "site-packages/urllib3/connectionpool.py", line 843, in urlopen
    retries = retries.increment(
  File "site-packages/urllib3/util/retry.py", line 519, in increment
    raise MaxRetryError(_pool, url, reason) from reason  # type: ignore[arg-type]
[...]

PR #414, opened 2024-07-11
Disable SSL verification for CloudFetch links

kravets-levko:
Fixes databricks/databricks-sql-python#413

Follow up on databricks/databricks-sql-python#405

~~Even though this is not a proper fix and it should be implemented as an option configured by user, this will allow us to unblock the upcoming release. The proper fix will be implemented as a follow up~~

### DIFF
--- a/src/databricks/sql/cloudfetch/downloader.py
+++ b/src/databricks/sql/cloudfetch/downloader.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 
 import requests
 from requests.adapters import HTTPAdapter, Retry
+from ssl import SSLContext, CERT_NONE
 import lz4.frame
 import time
 
@@ -65,9 +66,11 @@ class ResultSetDownloadHandler:
         self,
         settings: DownloadableResultSettings,
         link: TSparkArrowResultLink,
+        ssl_context: SSLContext,
     ):
         self.settings = settings
         self.link = link
+        self._ssl_context = ssl_context
 
     def run(self) -> DownloadedFile:
         """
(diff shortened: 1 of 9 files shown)


src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -95,7 +95,9 @@ def run(self) -> DownloadedFile:
         try:
             # Get the file via HTTP request
             response = session.get(
-                self.link.fileLink, timeout=self.settings.download_timeout
+                self.link.fileLink,
+                timeout=self.settings.download_timeout,
+                verify=False,
```
rth, line 100: You probably want to check this by default and let the user disable the check no?

kravets-levko: Yes, you're right. But I briefly checked the codebase - a proper fix will reguire quite some changes to be made. I'm already starting to work on it, but won't be able to deliver by Monday (we scheduled a release for Monday, and I don't want to block it). Disabling SSL verification is not good, but also not so critical

jackyhu-db (approved):

src/databricks/sql/cloudfetch/downloader.py
jackyhu-db: we should enable SSL by default

rth: Maybe a simple temporary  solution could be to add a, top level private variable `_SSL_VERIFY=True` in that file, use it, and then users could disable it via,
```py
import databricks.sql.cloudfetch.downloader

databricks.sql.cloudfetch.downloader._SSL_VERIFY = False
```
It's not very nice but at least there would be no promises that this is a public API, you'll be free to change it in the future, and it would solve the problem short term for users?

kravets-levko:
@jackyhu-db (cc @rth) Please check out my second take on this - https://github.com/databricks/databricks-sql-python/pull/414/commits/0cd439b6ff0770841b84f47a6a76ec3283fcd988 We already have flag `_tls_no_verify` that disables SSL cert validation (but still - does not disable SSL itself) for Thrift requests. That flag defaults to False and user should explicitly set it when needed. I decided to extend it for CloudFetch downloader as well. Please let me know what do you think. @rth it would be great if you can actually try this fix. Thanks!

PR #417, opened 2024-07-24
[PECO-1803] Splitting the PySql connector into the core and the non core part

jprakash-db:
[PECO-1803]

### Related Links

databricks_sqlalchemy split is present in this PR - https://github.com/databricks/databricks-sqlalchemy/pull/1

## Description
databricks-sql-python library is split so that package size can be reduced for the end user based on their requirements
Particularly pyarrow is the heavy component that is planned to be kept optional 
existing library split into 
 - databricks-sql-connector ( This is kept, so that the existing users import flow does not change )
 - databricks-sql-connector-core ( This is the lightweight library that separates the core part )

## Tasks Completed

- [x] Refractored the code into its respective folders based on the proposed design doc
- [x] pyproject.toml file has been changed to reflect the proper dependencies for the split
- [x] Made sure that all the existing e2e and units tests are working pre and post spit, ensuring parity
- [x] Added benchmarking queries to test the performance of pre and post split and a dashboard has been created for visualization
- [x] Dependency tests are also added to check how the library behaves when certain libraries are not available and the user requests their functions

## How to Test

Testing pipeline remains the same as it is before the split. 
pytest can be used to directly run both the integration as well as unit tests, by `pytest [directory_name or file_name]`

### Addition of dist folder in this repo

Github actions have been setup in the databricks_sqlalchemy repo to run tests using the databricks_sql_connector_core. For running those tests currently we need the .whl file in the dist folder and for temporary testing it has been added to the PR.

Once the library gets published to a public repository such as PyPi then databricks_sqlalchemy will automatically download it from that repo and run the tests using Github actions

## Performance Comparison - Benchmarking

The pre-split and post-split preformance comparison has been made using the large and small queries to make sure their is no regression of performance
Dashboard has been created so that everytime the benchmarking is run the result are stored in the benchfood, and comparisons can be made easily
Screenshot 2024-09-03 at 2 48 19 PM

[PECO-1803]: https://databricks.atlassian.net/browse/PECO-1803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


madhav-db (approved):

databricks_sql_connector_core/src/databricks_sql_connector_core/sql/auth/auth.py
```diff
@@ -107,9 +107,11 @@ def get_python_sql_connector_auth_provider(hostname: str, **kwargs):
         tls_client_cert_file=kwargs.get("_tls_client_cert_file"),
         oauth_scopes=PYSQL_OAUTH_SCOPES,
         oauth_client_id=kwargs.get("oauth_client_id") or client_id,
-        oauth_redirect_port_range=[kwargs["oauth_redirect_port"]]
-        if kwargs.get("oauth_client_id") and kwargs.get("oauth_redirect_port")
-        else redirect_port_range,
+        oauth_redirect_port_range=(
```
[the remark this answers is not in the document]
jprakash-db: @madhav-db  This is the black formatter auto formatting

databricks_sql_connector_core/README.md
```diff
@@ -65,8 +65,8 @@ or to a Databricks Runtime interactive cluster (e.g. /sql/protocolv1/o/123456789
 
 ## Contributing
 
-See [CONTRIBUTING.md](CONTRIBUTING.md)
+See [CONTRIBUTING.md](../CONTRIBUTING.md)
```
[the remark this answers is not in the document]
jprakash-db: @madhav-db In the new folder structure Contributing.md file is one level above. So the new link is necessary

databricks_sql_connector_core/src/databricks_sql_connector_core/sqlalchemy/__init__.py
```diff
@@ -0,0 +1,6 @@
+try:
```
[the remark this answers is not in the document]
jprakash-db: @madhav-db In this split databricks_sqlalchemy is kept optional and the requirement is to achieve this without changing the existing import structure for users. That is why this is necessary

.gitignore
```diff
@@ -48,7 +48,7 @@ __pycache__/
 .Python
 build/
 develop-eggs/
-dist/
+#dist/
```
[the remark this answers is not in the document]
jprakash-db: @kravets-levko I had added the reason for it in the PR description. I have to test the Github actions in the databricks_sqlalchemy - [PR Link](https://github.com/databricks/databricks-sqlalchemy/pull/1) repo, so till the time we don't publish the library to PyPi this was a temporary fix

databricks_sql_connector_core/README.md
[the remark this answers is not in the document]
jackyhu-db: We should not move the `README.md`, otherwise people will not see anything when they open this git repo

jackyhu-db (changes requested):
We should not move the README.md, otherwise people will not see anything when they open this git repo. Please keep the original README.md and update it with this new package split.

databricks_sql_connector_core/src/databricks_sql_connector_core/sql/auth/auth.py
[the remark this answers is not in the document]
jackyhu-db: please avoid change unnecessary file.

databricks_sql_connector_core/src/databricks_sql_connector_core/sql/auth/retry.py
```diff
@@ -14,7 +14,7 @@
 from urllib3 import Retry
 from urllib3.util.retry import RequestHistory
 
-from databricks.sql.exc import (
+from databricks_sql_connector_core.sql.exc import (
```
jackyhu-db, line 17: please do not change the namespace and also keep `databricks` namespace as the root.

databricks_sql_connector_core/src/databricks_sql_connector_core/sql/client.py
```diff
@@ -1196,7 +1220,44 @@ def fetchmany_arrow(self, size: int) -> pyarrow.Table:
 
         return results
 
-    def fetchall_arrow(self) -> pyarrow.Table:
+    def merge_columnar(self, result1, result2):
```
jackyhu-db, line 1223: can you move the code change into a separated PR? Let this PR only focus the repo restructure.

databricks_sql_connector_core/src/databricks_sql_connector_core/sql/client.py
```diff
@@ -1196,7 +1220,44 @@ def fetchmany_arrow(self, size: int) -> pyarrow.Table:
...
+        """
+        Function to merge / combining the columnar results into a single result
+
+        :param result1:
+        :param result2:
+        :return:
+        """
+        merged_result = [result1[i] + result2[i] for i in range(len(result1))]
```
jackyhu-db, line 1231: are the len of the results same? Can you add a check before merge?

databricks_sql_connector_core/pyproject.toml
```diff
@@ -14,23 +14,17 @@ thrift = ">=0.16.0,<0.21.0"
...
-sqlalchemy = { version = ">=2.0.21", optional = true }
 openpyxl = "^3.0.10"
+databricks_sqlalchemy = { version = ">=1.0.0", optional = true }
 alembic = { version = "^1.0.11", optional = true }
 urllib3 = ">=1.26"
 
 [tool.poetry.extras]
-sqlalchemy = ["sqlalchemy"]
-alembic = ["sqlalchemy", "alembic"]
+databricks_sqlalchemy = ["databricks_sqlalchemy"]
```
jackyhu-db, line 26: why do we need `databricks_sqlalchemy` in `sql_connector_core`, it should NOT have this dependency or declaration.

databricks_sql_connector_core/src/databricks_sql_connector_core/sql/__init__.py
```diff
@@ -1,6 +1,6 @@
 import datetime
 
-from databricks.sql.exc import *
+from databricks_sql_connector_core.sql.exc import *
```
jackyhu-db, line 3: we cannot change the package name, otherwise it will break existing consumers.

examples/interactive_oauth.py
```diff
@@ -1,4 +1,4 @@
-from databricks import sql
+from databricks_sql_connector_core import sql
```
jackyhu-db, line 1: all the examples should still work w/o any code change, we should not change these. If it is required, it means this change will break all the customer codes.

kravets-levko:
As the first iteration, it is more or less okay. Still, there are few major things to fix.

The core package should have most of the dependencies, as well as tests, and so on. The dependencies like `pyarrow` should be optional for it, and thus not installed by default. The core package should keep the original project's structure - in other words, everything inside of `databricks_sql_connector_core` should remain the same as it is now, but just moved into own folder.

The full package should basically have no source files, and the onlt dependency of the core package, but will all its optional dependencies enabled. Please refer to Poetry docs - https://python-poetry.org/docs/pyproject/#extras (look for `extras=["all"]` config of the dependency)

Why it is important and how it should work? When you publish the package, Poetry collects files from the folder where `pyproject.toml` file is located, puts them into archive and uploads to Pypi under the name specified in `pyproject.toml`. Later, when package is installed, `pip` basically just unpacks that archive as is to the folder where all packages are located (`.../lib/python3/site-packages/`). Let's imagine that we created two packages which both contain their sources in the folder `xyz`. After installing both of them, both will be corrupted because their sources will be extracted to the same path on target machine.

So with the configuration described above, we will have all the source code located in the core package. After installation, it will be extracted to the same folders as earlier, so users will not have to update their imports. Optional dependencies will not be installed by default, but if user already has them installed (e.g. `pyarrow`) - correcsponding features will immediately become (remain) available.

Installing the full package will not add any new source files to the user's machine. Everything is already installed with the core package. The full package should just tell `pip` to install optional dependencies.

README.md
```diff
@@ -65,8 +65,8 @@ or to a Databricks Runtime interactive cluster (e.g. /sql/protocolv1/o/123456789
 
 ## Contributing
 
-See [CONTRIBUTING.md](CONTRIBUTING.md)
+See [CONTRIBUTING.md](../CONTRIBUTING.md)
```
jackyhu-db, line 68: Both `README.md` and `CONTRIBUTING.md` are top level, this change is not required.

databricks_sql_connector/README.md
jackyhu-db, line 1: Is this "[src/databricks/sql/auth/__init__.py → databricks_sql_connector/README.md]" intended? I guess not and I saw it was empty file.

databricks_sql_connector/dist/databricks_sql_connector-1.0.0.tar.gz
jackyhu-db, line 1: we should not put any build artifact or `dist` folder in the source repo, please remove it.

databricks_sql_connector_core/dist/databricks_sql_connector_core-1.0.0.tar.gz
jackyhu-db, line 1: same as above, please remove them.

databricks_sql_connector/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[tool.poetry]
+name = "databricks-sql-connector"
+version = "1.0.0"
```
jackyhu-db, line 3: It should be next major version of the existing `databricks-sql-connector`, I believe it is `4.0.0`

databricks_sql_connector_core/src/databricks/sql/client.py
```diff
@@ -1,7 +1,12 @@
 from typing import Dict, Tuple, List, Optional, Any, Union, Sequence
 
 import pandas
-import pyarrow
+
+try:
+    import pyarrow
+except ImportError:
+    pyarrow = None
+
```
jackyhu-db, line 9: can we move this to the last import

databricks_sql_connector_core/src/databricks/sql/thrift_api/TCLIService/TCLIService.py
jackyhu-db, line 1: can you avoid changing this file if the change is only for format? Let us try to minimize the change in this PR by avoid unrelated or unnecessary change.

databricks_sql_connector_core/src/databricks/sql/auth/auth.py
```diff
@@ -107,9 +107,11 @@ def get_python_sql_connector_auth_provider(hostname: str, **kwargs):
...
         oauth_scopes=PYSQL_OAUTH_SCOPES,
         oauth_client_id=kwargs.get("oauth_client_id") or client_id,
-        oauth_redirect_port_range=[kwargs["oauth_redirect_port"]]
-        if kwargs.get("oauth_client_id") and kwargs.get("oauth_redirect_port")
-        else redirect_port_range,
+        oauth_redirect_port_range=(
+            [kwargs["oauth_redirect_port"]]
+            if kwargs.get("oauth_client_id") and kwargs.get("oauth_redirect_port")
+            else redirect_port_range
+        ),
```
jackyhu-db, line 114: Let us try to minimize the change in this PR by avoid unrelated or unnecessary change (e.g. foramatting).

databricks_sql_connector_core/src/databricks/sql/auth/authenticators.py
```diff
@@ -4,11 +4,17 @@
...
+    infer_cloud_from_host,
+)
 
 # Private API: this is an evolving interface and it will change in the future.
 # Please must not depend on it in your applications.
-from databricks.sql.experimental.oauth_persistence import OAuthToken, OAuthPersistence
+from databricks.sql.experimental.oauth_persistence import (
+    OAuthToken,
+    OAuthPersistence,
+)
```
jackyhu-db, line 17: same as above. Please avoid format change in this PR.

jackyhu-db (commented):
Please remove all the code format change in this PR to minimize the changes in this PR

databricks_sql_connector_core/src/databricks/sql/auth/authenticators.py
jprakash-db: @jackyhu-db I tried running the black autoformatter. This is the result it is giving, is there some other way to do this ?

databricks_sql_connector/pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+description = "Databricks SQL Connector for Python"
+authors = ["Databricks <databricks-sql-connector-maintainers@databricks.com>"]
+license = "Apache-2.0"
+readme = "README.md"
+include = ["CHANGELOG.md"]
+
+[tool.poetry.dependencies]
+python = "^3.8.0"
+pyarrow = ">=14.0.1,<17"
+databricks_sql_connector_core = { version = ">=1.0.0"}
```
kravets-levko, line 13: This package should basically have no other dependencies except `databricks_sql_connector_core`. Installation of optional deps should be achieved by adding `extras=["all"]` option (like - `databricks_sql_connector_core = { version = ">=1.0.0", extras=["all"]}`. Have you tried it? If there are any caveats with this approach - please share your thoughts

databricks_sql_connector/pyproject.toml
kravets-levko: I'm actually not so sure about it. I think, if the library behavior remains the same for user - no need to change major version number. At least for now we can keep a current version number, and decide how to increment it before the next release

databricks_sql_connector_core/src/databricks/sql/auth/authenticators.py
kravets-levko: It may be that after your changes some newer version of Black was installed, and it formats files slightly different. You can check this by checking out `main` to a new folder, setting up a devloop, and then comparing to what you have here

databricks_sql_connector_core/src/databricks/sql/auth/authenticators.py
```diff
@@ -25,12 +31,10 @@ class CredentialsProvider(abc.ABC):
     for authenticating requests to Databricks REST APIs"""
 
     @abc.abstractmethod
-    def auth_type(self) -> str:
-        ...
```
kravets-levko, line 29: Not related to this PR and shouldn't be fixed here, but I've already seen this so much all over our codebase so far. Even though ellipsis literal _can_ be used as a function body, it's not a replacement for `pass`. `pass` actually creates function with no body, while ellipsis literal creates a function with body which consists of that ellipsis literal. It's not any different from using any other literal for this purpose, e.g. `def f(): "hello, world"`. I think after we're done with this splitting, we should go over the code and replace these with `pass`

databricks_sql_connector_core/src/databricks/sql/client.py
```diff
@@ -22,6 +27,7 @@
     ParamEscaper,
     inject_parameters,
     transform_paramstyle,
+    ArrowQueue,
```
kravets-levko, line 30: Is this import used anywhere in this file? If not - please remove it

databricks_sql_connector_core/src/databricks/sql/thrift_api/TCLIService/TCLIService.py
kravets-levko: Even more - this file should not be changed at all because it is generated from `.thrift` definitions (as well as other files in this folder)

databricks_sql_connector_core/src/databricks/sql/utils.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import annotations
-
+import pytz
```
kravets-levko, line 2: This import also seems to be a leftover. If not used - please remove it

databricks_sql_connector_core/tests/e2e/common/decimal_tests.py
```diff
@@ -1,14 +1,28 @@
...
+    pyarrow = None
 import pytest
 
 
-class DecimalTestsMixin:
-    decimal_and_expected_results = [
+def decimal_and_expected_results():
+    if pyarrow is None:
+        return []
+
```
kravets-levko, line 13: If the whole test suite is skipped when pyarrow not available - no need to ass additional checks to these functions as they will not be executed anyway (same for other helper functions here)

databricks_sql_connector_core/tests/e2e/common/predicates.py
```diff
@@ -8,9 +8,12 @@
 
 
 def pysql_supports_arrow():
-    """Import databricks.sql and test whether Cursor has fetchall_arrow."""
-    from databricks.sql.client import Cursor
-    return hasattr(Cursor, 'fetchall_arrow')
+    """Checks if the pyarrow library is installed or not"""
+    try:
+        __import__("pyarrow")
```
kravets-levko, line 13: `__import__` is not meant to be used by programmers. Please use regular `import` instead

databricks_sql_connector_core/tests/e2e/test_complex_types.py
```diff
@@ -4,6 +4,9 @@
 from tests.e2e.test_driver import PySQLPytestTestCase
 
 
+@pytest.mark.skipif(
+    pytest.importorskip("pyarrow"), reason="Skipping because pyarrow is not installed"
+)
```
kravets-levko, line 9: Okay, so I've seen at least three ways to conditionally skip tests when `pyarrow` not available 🙂 Please choose one of them and stick with it everywhere.

Also, please note that here `pytest.importskip` is not used correctly (see https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip). It should be used in a test body to skip the remaining part of it. `@pytest.mark.skipif` decorator seems to be a better fit if we want to skip the whole test case, but in this case it's better to use it with the predicate you created to check the presense of `pyarrow`

poetry.lock
kravets-levko, line 1: Lockfile should be kept - it's important to have reproducible installs

kravets-levko (changes requested):
Folder structure seems to be good. Two major things to improve - figure out if we can better split dependencies, and also lockfile should be restored/recreated for both new pyprojects (please see my comments)

README.md
```diff
@@ -65,8 +65,8 @@ or to a Databricks Runtime interactive cluster (e.g. /sql/protocolv1/o/123456789
...
 
 ## License
 
-[Apache License 2.0](LICENSE)
+[Apache License 2.0](../LICENSE)
```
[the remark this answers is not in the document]
jprakash-db: fixed

README.md
jprakash-db: fixed

databricks_sql_connector/dist/databricks_sql_connector-1.0.0.tar.gz
jprakash-db: fixed

databricks_sql_connector_core/dist/databricks_sql_connector_core-1.0.0.tar.gz
jprakash-db: fixed

databricks_sql_connector/README.md
jprakash-db: fixed

databricks_sql_connector_core/src/databricks/sql/client.py
jprakash-db: fixed

databricks_sql_connector_core/src/databricks/sql/client.py
jprakash-db: fixed

databricks_sql_connector_core/src/databricks/sql/utils.py
jprakash-db: fixed

databricks_sql_connector_core/tests/e2e/common/decimal_tests.py
jprakash-db: fixed

databricks_sql_connector_core/tests/e2e/common/predicates.py
jprakash-db: fixed

databricks_sql_connector_core/tests/e2e/test_complex_types.py
jprakash-db: fixed

jprakash-db: fixed

databricks_sql_connector_core/src/databricks/sql/auth/retry.py
```diff
@@ -77,7 +77,7 @@ class DatabricksRetryPolicy(Retry):
         commands like ExecuteStatement. This is passed to urllib3 by extending its status_forcelist
 
     :param urllib3_kwargs:
-        Dictionary of arguments that are passed to Retry.__init__. Any setting of Retry() that
+        Dictionary of arguments that are passed to Retry.__init__.py. Any setting of Retry() that
```
kravets-levko, line 80: These `.py` suffixes (?) should not be here - `Retry.__init__` is a magic method, not a file name

databricks_sql_connector_core/src/databricks/sql/client.py
```diff
@@ -118,7 +125,7 @@ def __init__(
 
                 from databricks.sql.experimental.oauth_persistence import OAuthPersistence, OAuthToken
                 class MyCustomImplementation(OAuthPersistence):
-                    def __init__(self, file_path):
+                    def __init__.py(self, file_path):
```
kravets-levko, line 128: Same here - `__init__` is a method name, not a file name. Also, how did this code even run? It has to throw a syntax error

databricks_sql_connector_core/src/databricks/sql/thrift_backend.py
```diff
@@ -56,7 +60,7 @@
 TIMESTAMP_AS_STRING_CONFIG = "spark.thriftserver.arrowBasedRowSet.timestampAsString"
 DEFAULT_SOCKET_TIMEOUT = float(900)
 
-# see Connection.__init__ for parameter descriptions.
+# see Connection.__init__.py for parameter descriptions.
```
kravets-levko, line 63: Another one here - `__init__` is a method name, not a file name

databricks_sql_connector/pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[tool.poetry]
+name = "databricks-sql-connector"
+version = "3.3.0"
```
jackyhu-db, line 3: As we change the package structure in this release. I would suggest to bump the major version to `4.0.0` or at least it should be greater than current release `3.4.0`,  so it has to start with `3.5.0`. @kravets-levko wdyt?

jackyhu-db (approved):

kravets-levko (approved):

PR #419, opened 2024-07-30
[PECO-1801] Make OAuth as the default authenticator if no authentication setting is provided

jackyhu-db:
- Make OAuth as the default authenticator if no authentication setting is provided
- Update `README.md` to remove the `access_token` authenticator from the example


kravets-levko (approved):
LGTM 👍

PR #425, opened 2024-08-13
[PECO-1857] Use SSL options with HTTPS connection pool

kravets-levko:
This issue seems to be a regression after databricks/databricks-sql-python#131 

SSL options (`_tls_no_verify`, `_tls_trusted_ca_file`, `_tls_client_cert_file`, `_tls_client_cert_key_file`, `_tls_client_cert_key_password`) are not used for requests to Thrift backend: they are just not passed to connection pool

### DIFF
--- a/src/databricks/sql/auth/thrift_http_client.py
+++ b/src/databricks/sql/auth/thrift_http_client.py
@@ -26,9 +26,6 @@ class THttpClient(thrift.transport.THttpClient.THttpClient):
         uri_or_host,
         port=None,
         path=None,
-        cafile=None,
-        cert_file=None,
-        key_file=None,
         ssl_options: SSLOptions = None,
         max_connections: int = 1,
         retry_policy: Union[DatabricksRetryPolicy, int] = 0,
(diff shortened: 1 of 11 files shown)


yunbodeng-db:
When you make the change please let @benc-db know so he can change his code to adapt.

src/databricks/sql/auth/thrift_http_client.py
```diff
@@ -48,13 +48,11 @@ def __init__(
...
             if self.scheme == "https":
-                self.certfile = cert_file
-                self.keyfile = key_file
-                self.context = (
-                    ssl.create_default_context(cafile=cafile)
-                    if (cafile and not ssl_context)
-                    else ssl_context
-                )
+                if self._ssl_options is not None:
+                    # TODO: Not sure if those options are used anywhere - need to double-check
```
benc-db, line 52: have you double checked :P?

benc-db (approved):

src/databricks/sql/auth/thrift_http_client.py
kravets-levko: Not yet - decided that it's not that critical for now. But will do - I promise 😅

PR #434, opened 2024-09-02
Relax `pyarrow` pin

dhirschfeld:


kravets-levko:
@dhirschfeld Thank you for submitting this! Please update `poetry.lock` file as well - CI fails because lockfile doesn't match `pyproject.toml`

dhirschfeld:
> @dhirschfeld Thank you for submitting this! Please update `poetry.lock` file as well - CI fails because lockfile doesn't match `pyproject.toml`

Yeah, I plan to circle back to do it properly as soon as I can squeeze it in.

PR #436, opened 2024-09-03
Added Json type to the sqlalchemy parser

jprakash-db:
## Description

Added the Json type in the sqlalchemy types file to handle the compilation of Json to string

### DIFF
--- a/src/databricks/sqlalchemy/_types.py
+++ b/src/databricks/sqlalchemy/_types.py
@@ -33,6 +33,7 @@ def process_literal_param_hack(value: Any):
 @compiles(sqlalchemy.types.Unicode, "databricks")
 @compiles(sqlalchemy.types.UnicodeText, "databricks")
 @compiles(sqlalchemy.types.Uuid, "databricks")
+@compiles(sqlalchemy.types.Json, "databricks")
 def compile_string_databricks(type_, compiler, **kw):
     """
     We override the default compilation for Enum(), String(), Text(), and Time() because SQLAlchemy


src/databricks/sqlalchemy/_types.py
```diff
@@ -33,6 +33,7 @@ def process_literal_param_hack(value: Any):
 @compiles(sqlalchemy.types.Unicode, "databricks")
 @compiles(sqlalchemy.types.UnicodeText, "databricks")
 @compiles(sqlalchemy.types.Uuid, "databricks")
+@compiles(sqlalchemy.types.Json, "databricks")
```
yunbodeng-db, line 36: We should add a test to cover the new type? I think we have existing test covering other types already.

src/databricks/sqlalchemy/test_local/test_types.py
```diff
@@ -128,6 +128,7 @@ def test_numeric_renders_as_decimal_with_precision_and_scale(self):
     sqlalchemy.types.INT: DatabricksDataType.INT,
     sqlalchemy.types.SMALLINT: DatabricksDataType.SMALLINT,
     sqlalchemy.types.TIMESTAMP: DatabricksDataType.TIMESTAMP,
+    sqlalchemy.types.JSON: DatabricksDataType.STRING,
```
yunbodeng-db, line 131: Others with more knowledge can chime in, but I wonder if you need to do something similar to TinyIntegerTest to see if the type mapping will actually work. I think this addition makes sure the type will be recognized and translated, but I actually do not know how you would know the value is rendered as a literal string properly.

kravets-levko:
@jprakash-db Generally the idea you implemented looks correct - in Databricks, JSON values are stored as strings, so we have to give SQLAlchemy a hint about it. However, it would be nice to have tests to check that data are actually serialized correctly (as JSON-encoded string in this case). Currently we only have tests that verify it the type is mapped properly, but IMHO it's not enough. Can you please check if we have any tests to check that all our customized data types are actually properly serialized, and add JSON values to it? If we don't have such test suite - can you please create it, and it least check for JSON, Enum ans UUID types being properly serialized?

src/databricks/sqlalchemy/_types.py
```diff
@@ -33,6 +33,7 @@ def process_literal_param_hack(value: Any):
 @compiles(sqlalchemy.types.Unicode, "databricks")
 @compiles(sqlalchemy.types.UnicodeText, "databricks")
 @compiles(sqlalchemy.types.Uuid, "databricks")
+@compiles(sqlalchemy.types.JSON, "databricks")
 def compile_string_databricks(type_, compiler, **kw):
     """
     We override the default compilation for Enum(), String(), Text(), and Time() because SQLAlchemy
```
kravets-levko, line 39: Please update this comment as well - to match decorators list for this function

jprakash-db: ok, adding this

PR #440, opened 2024-09-13
[PECO-1926] Create a non pyarrow flow to handle small results for the column set

jprakash-db:
[PECO-1926]

## Description

To handle the flow of result when pyarrow is not present. This setup is particularly used for users who want to handle small quantities of data 

## Task Completed

- [x] Introduced the new non arrow flow for data to be handled by the users not having pyarrow
- [x] ColumnQueue is the new ResultSetQueue that I have added to handle the column set data 
- [x] Entire pipeline of handling the column set data to finally converting it to the TRow type for the user has been implemented
- [x] Added tests for the ColumnQueue to ensure that it is working as intended

[PECO-1926]: https://databricks.atlassian.net/browse/PECO-1926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -25,7 +25,7 @@ from databricks.sql.utils import (
     ParamEscaper,
     inject_parameters,
     transform_paramstyle,
-    ArrowQueue,
+    ColumnTable,
     ColumnQueue
 )
 from databricks.sql.parameters.native import (
@@ -1152,10 +1152,10 @@ class ResultSet:
         column_names = [c[0] for c in self.description]
         ResultRow = Row(*column_names)
         result = []
-        for row_index in range(len(table[0])):
+        for row_index in range(table.num_rows):
             curr_row = []
-            for col_index in range(len(table)):
-                curr_row.append(table[col_index][row_index])
+            for col_index in range(table.num_columns):
+                curr_row.append(table.get_item(col_index,  row_index))
             result.append(ResultRow(*curr_row))
 
         return result
@@ -1235,11 +1235,11 @@ class ResultSet:
         :return:
         """
 
-        if len(result1) != len(result2):
-            raise ValueError("The number of columns in both results must be the same")
+        if result1.column_names != result2.column_names:
+            raise ValueError("The columns in the results don't match")
 
-        merged_result = [result1[i] + result2[i] for i in range(len(result1))]
-        return merged_result
+        merged_result = [result1.column_table[i] + result2.column_table[i] for i in range(result1.num_columns)]
+        return ColumnTable(merged_result, result1.column_names)
 
     def fetchmany_columnar(self, size: int):
         """
@@ -1250,8 +1250,8 @@ class ResultSet:
             raise ValueError("size argument for fetchmany is %s but must be >= 0", size)
 
         results = self.results.next_n_rows(size)
-        n_remaining_rows = size - len(results[0])
-        self._next_row_index += len(results[0])
+        n_remaining_rows = size - results.num_rows
+        self._next_row_index += results.num_rows
 
         while (
                 n_remaining_rows > 0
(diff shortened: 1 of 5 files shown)


src/databricks/sql/utils.py
```diff
@@ -96,11 +102,33 @@ def build_queue(
...
+        length = min(num_rows, self.n_valid_rows - self.cur_row_index)
+        # Slicing using the default python slice
+        next_data = [
+            column[self.cur_row_index : self.cur_row_index + length]
+            for column in self.columnar_table
+        ]
+        self.cur_row_index += length
+        return next_data
+
+    def remaining_rows(self):
```
jackyhu-db, line 122: Is this standard naming? The name seems this is read only operation

src/databricks/sql/client.py
```diff
@@ -1210,7 +1227,42 @@ def fetchmany_arrow(self, size: int) -> pyarrow.Table:
...
+    def fetchmany_columnar(self, size: int):
+        """
+        Fetch the next set of rows of a query result, returning a Columnar Table.
+        An empty sequence is returned when no more rows are available.
+        """
+        if size < 0:
+            raise ValueError("size argument for fetchmany is %s but must be >= 0", size)
+
+        results = self.results.next_n_rows(size)
+        n_remaining_rows = size - len(results[0])
```
jackyhu-db, line 1249: what if there is no column in the result?

src/databricks/sql/utils.py
```diff
@@ -560,6 +588,29 @@ def convert_decimals_in_arrow_table(table, description) -> pyarrow.Table:
     return table
 
 
+def convert_to_assigned_datatypes_in_column_table(column_table, description):
+    for i, col in enumerate(column_table):
```
jackyhu-db, line 592: I would recommend not change `column_table`, instead, return a table with new assigned data types, input argument should be immutable, if the function crashes in the middle, the state of column_table becomes inconsisent.

src/databricks/sql/utils.py
```diff
@@ -595,6 +653,26 @@ def _convert_column_to_arrow_array(t_col):
     raise OperationalError("Empty TColumn instance {}".format(t_col))
 
 
+def _covert_column_to_list(t_col):
+    supported_field_types = (
```
jackyhu-db, line 657: can we define it as the static or constant variable if this is constant.

src/databricks/sql/utils.py
jprakash-db: The names are defined in the already existing ResultSetQueue from which I inherited

src/databricks/sql/utils.py
jprakash-db: The function is not part of a class, so I don't think we can make it static. In Python I believe we can't make anything constant, so I have replaced the names with ALL CAPS to indicate that this is a constant

src/databricks/sql/utils.py
jprakash-db: Fixed it

jackyhu-db (approved):

src/databricks/sqlalchemy/pytest.ini
```diff
@@ -0,0 +1,4 @@
+
```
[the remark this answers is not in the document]
jprakash-db: Removed it

src/databricks/sql/thrift_backend.py
```diff
@@ -621,6 +624,12 @@ def _get_metadata_resp(self, op_handle):
 
     @staticmethod
     def _hive_schema_to_arrow_schema(t_table_schema):
+
+        if pyarrow is None:
+            raise ImportError(
+                "pyarrow is required to convert Hive schema to Arrow schema"
+            )
+
```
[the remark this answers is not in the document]
jprakash-db: fixed

kravets-levko (approved):

PR #442, opened 2024-09-17
"Export" `DatabricksDialect` from `databricks.sqlalchemy`

Kilo59:
Although python doesn't really have an "Export" command, type checkers like mypy will warn if an imported object isn't part of module's `__all__` list.

https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport


kravets-levko:
Thank you @Kilo59! Your change seems good. We currently have a code freeze due to some refactoring, but after that I'll merge your PR, and it will become available with the next release

Kilo59:
@kravets-levko any updates here?

PR #447, opened 2024-10-03
[PECO-1961] On non-retryable error, ensure PySQL includes useful information in error

shivam2680:
## Description
This PR addresses an issue where the original exception context was not included in the error message. 

### Changes made
Appended original exception context `self.error` in the user-friendly error message

### Testing
Simulated authentication error (Non-retry) and verified exception was being raised with error message.

### DIFF
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -357,7 +357,12 @@ class RequestErrorInfo(
             user_friendly_error_message = "{}: {}".format(
                 user_friendly_error_message, self.error_message
             )
-        return user_friendly_error_message
+        try:
+            error_context = str(self.error)
+        except:
+            error_context = ""
+
+        return user_friendly_error_message + ". " + error_context
 
 
 # Taken from PyHive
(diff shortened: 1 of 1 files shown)


src/databricks/sql/utils.py
```diff
@@ -357,7 +357,12 @@ def user_friendly_error_message(self, no_retry_reason, attempt, elapsed):
             user_friendly_error_message = "{}: {}".format(
                 user_friendly_error_message, self.error_message
             )
-        return user_friendly_error_message
+        try:
+            error_context = str(self.error)
+        except:
+            error_context = ""
+
+        return user_friendly_error_message + ". " + error_context
```
kravets-levko, line 365: `str()` may fail for many weird reasons. Maybe let's either set some default message (`error_context = "unknown error"`), or catch the exception and use its message:

```suggestion
        try:
            error_context = str(self.error)
        except  Exception as e:
            error_context = str(e)

        return user_friendly_error_message + ". " + error_context
```

shivam2680: a default message makes sense. I'll update.

kravets-levko (approved):

Issue #432, opened 2024-08-27
Support for numpy >= 2.0.2


envelio-cb:
I am working on upgrading our ecosystem to use numpy >= 2.0.2 which currently breaks at your package.

Afaik you don't use numpy in your code except for testing. Thus, I am wondering if you can either drop the dependency and add it as dev-only dependency or if you can be less strict in the versioning of numpy and support numpy >= 2.0.2.

Here is also a list of the numpy ecosystem compatibility with numpy 2.0:
https://github.com/numpy/numpy/issues/26191

Kind regards

sshevlyagin:
+1

jslorrma:
+1

Bendik-ml:
+1 👯‍♂️

tblatrille:
+1

jprakash-db:
Pandas and numpy has compatiblity issues, numpy version 2 is only supported after pandas 2.2.2 and any version before that is not supported. - https://pandas.pydata.org/docs/whatsnew/v2.2.2.html#pandas-2-2-2-is-now-compatible-with-numpy-2-0

PR #428, opened 2024-08-22
wip: relax upper pyarrow bound

rauerhans:
implements https://github.com/databricks/databricks-sql-python/issues/427

Issue #427, opened 2024-08-22
Update for pyarrow 17


rauerhans:
We need pyarrow 17 in one of our projects, because of other dependency constraints. However, we also need 3.3.0, because we are facing unstable database connections with this library's version < 3. So it'd be lovely to relax the dependency on Pyarrow < 17 to include the [latest stable version of 17](https://arrow.apache.org/docs/python/index.html).

arredond:
(copying over from the PR comments)

I see the fix here is to move to pyarrow<18.0 as the upper bound. Would you consider not adding an upper bound at all? Upper bounds really make dependency management much, much harder for complex projects.

[Here's](https://iscinumpy.dev/post/bound-version-constraints/) a nice discussion on the topic.

Also, I see this has already been a problem for other users in the past: #389 , #390 , #55

Issue #88, opened 2023-02-10
Strict numpy dependency


bjgbeelen:
Hi!

I'm running into an issue that this package requires a very specific numpy package `1.23.4`. Aside from wondering why it needs numpy in the first place (I didn't dive into that), I'm wondering if this needs to be this specific. 

Context:
We are using NixPkgs to provide reproducible environments. I'm trying to compile the latest `dbt-databricks` version and this is a dependency. From the latest stable NixPkgs I get numpy version `1.23.3`. I can override it but building this specific version of numpy myself is quite heavy. And it makes me wonder if it can be solved in a different way by relaxing the constraint :-)

Thanks for thinking along!

betodealmeida:
Same here, I'm having problems updating the Databricks driver in Apache Superset because of this. Can we relax the versioning here?

https://github.com/databricks/databricks-sql-python/blob/e3f3a6febc903399ceb2db858faa142c754ad38c/pyproject.toml#L22-L25

susodapop:
There's a PR for this here: #74 we're working on merging it ASAP.

susodapop:
Also @betodealmeida we'd like to improve the Superset experience which currently doesn't use our built-in SQLAlchemy dialect (just released in v2.4.0). Are you interested in helping us test the new version?

betodealmeida:
> Also @betodealmeida we'd like to improve the Superset experience which currently doesn't use our built-in SQLAlchemy dialect (just released in v2.4.0). Are you interested in helping us test the new version?

Yeah, I'm happy to bump the version and test it! Will do it this weekend.

susodapop:
@betodealmeida Unfortunately that won't work! Superset needs to be configured to use our in-built dialect. At present if you just bump the version of `databricks-sql-connector` nothing will change. Once we have concrete steps for testing I'll ping you here.

vikrantpuppala:
Closing this as the issue is quite stale - please reopen if help is required

PR #452, opened 2024-10-07
Remove upper caps on dependencies

arredond:
## What

Like many Poetry-managed projects, this package sets upper caps on dependencies both via caret (`^`) and lesser-than signs (`<=`). This PR relaxes these requirements to just establishing lower bounds, not upper ones.

## Why

Poetry heavily believes in SemVer but many projects do not strictly follow it. For instance, [pyarrow](https://pypi.org/project/pyarrow/#history) has bumped major versions **eight** times in the last 2 years without really making breaking changes for the vast majority of users.

Many well-regarded devs in the Python ecosystem are calling against upper-bounds in projects due to the complications it generates for end users ([here's](https://iscinumpy.dev/post/bound-version-constraints/) an excellent discussion on the topic from a PyPa member).

Personally, I find myself in "dependency hell" in a project where we use SQLAlchemy to interface with many kinds of databases (the area where SQLAlchemy excels, for sure). The latest dialects we're trying to support require SQLAlchemy >= 2.0, but our current `databricks-sql-connector`version of `2.9.3` doesn't allow it. We also can't upgrade to the latest `3.4.0` because an upper limit on `pyarrow` has been set (which didn't exist before). A similar experience happens with Numpy >= 2.0.

## Related issues / PRs

Many versions of this issue have popped up before for this package (especially for `pyarrow` and `numpy`):
#427 
#55 
#428 
#74 
#88
#432

## Note

If this PR seems too ambitious, I'm happy to provide another one that only removes or, at least, bumps the upper cap for `pyarrow` and `numpy`. However, I believe removing upper caps in a general sense is the way to go. Major version bumps in Python are very frequently backwards compatible, and it's a much better experience for users to fix problems when they occur than to prevent them from upgrading (and eventually ditching your package altogether).

Here's a quote in the linked article from Brett Cannon, Python Steering Council Member and packaging expert:

> Libraries/packages should be setting a floor, and if necessary excluding known buggy versions, but otherwise don’t cap the maximum version as you can’t predict future compatibility

Or another quote from the author of the article himself:

> If you absolutely must set upper limits, you should release a new version as soon as possible with a higher cap when a dependency updates (ideally before the dependency releases the update). If you are committing to this, why not just quickly release a patch release with caps only after an actual conflict happens? It will be less common, and will help you quickly sort out and fix incompatibilities, rather than hiding your true compatibilities and delaying updates. **You want users to use the latest versions of your libraries if there’s a problem, so why can’t you offer the same consideration to the libraries you depend on and use?**

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1202,4 +1202,4 @@ sqlalchemy = ["sqlalchemy"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.8.0"
-content-hash = "caa1c9e9001579a6f3ad2ca2ef0c6481ba6a860de8b72c137a7f4e9f44ea16ad"
+content-hash = "9d8a91369fc79f9ca9f7502e2ed284b66531c954ae59a723e465a76073966998"
--- a/poetry.lock
+++ b/poetry.lock
@@ -1202,4 +1202,4 @@ sqlalchemy = ["sqlalchemy"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.8.0"
-content-hash = "31066a85f646d0009d6fe9ffc833a64fcb4b6923c2e7f2652e7aa8540acba298"
+content-hash = "caa1c9e9001579a6f3ad2ca2ef0c6481ba6a860de8b72c137a7f4e9f44ea16ad"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,16 +16,16 @@ pandas = [
 ]
 pyarrow = ">=14.0.1"
 
-lz4 = ">=4.0.2"
-requests = ">=2.18.1"
-oauthlib = ">=3.1.0"
+lz4 = "^4.0.2"
+requests = "^2.18.1"
+oauthlib = "^3.1.0"
 numpy = [
     { version = ">=1.16.6", python = ">=3.8,<3.11" },
     { version = ">=1.23.4", python = ">=3.11" },
 ]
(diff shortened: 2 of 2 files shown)


arredond:
@andrefurlan-db @benc-db @jackyhu-db @kravets-levko @rcypher-databricks @yunbodeng-db pinging any of you in case this has gone unseen

susodapop (approved):
Change LGTM and the motivation is a solid one.

hauntsaninja:
@jackyhu-db any chance you could take a look at this? I'm interested in removing the preemptive cap for numpy 2 (see also https://github.com/databricks/databricks-sql-python/issues/432 )

jackyhu-db:
@arredond thanks for making the change. Can you reduce the scope of the change? Only change the `pyarrow` and `numpy`?

arredond:
Of course, I'll make a change to this PR reducing the scope.

However, at the risk of repeating myself and being annoying, please consider removing all upper caps in the future. Numpy and pyarrow have been the problem this time but it's very likely that in the near future you'll be seeing this same issue with, say, `requests`.

Once again copying that quote from Brett Cannon in the PR description, since these well-established folks make the case much better than I do:

> Libraries/packages should be setting a floor, and if necessary excluding known buggy versions, but otherwise don’t cap the maximum version as you can’t predict future compatibility

jackyhu-db (approved):

jprakash-db (approved):

PR #460, opened 2024-10-25
[PECO-2051] Add custom auth headers into cloud fetch request

jackyhu-db:


examples/data_types.py
kravets-levko, line 1: @jackyhu-db Is this related to this PR?

kravets-levko (approved):

examples/data_types.py
jackyhu-db: Good catch, let me fix it.

Issue #458, opened 2024-10-22
If the token is null, the connection hangs


wallystart:
Hi, today we were troubleshooting a process that uses databricks-sql-python. The connection hangs if the token is null.

Evidence to reproduce:
![image](https://github.com/user-attachments/assets/1943154d-2f1c-4da0-ba11-6a16f6f27214)
(The cell never ends)

Generating token it ends correctly:
![image](https://github.com/user-attachments/assets/03ea3615-ef01-4e68-88e2-3ef16e68d13a)

We are using databricks-sql-connector==3.4.0

Thanks!

kravets-levko:
Hi @wallystart! In v3.4.0, when you omit PAT, U2M OAuth will be used. It means that library will try to open browser tab and ask to log in. If browser wasn't open for some reason - it may look like the whole thing hung up. I'm curious how you run your script, and in which environment. Also, please enable debug mode (add following lines in beginning of your script) and share the full log you'll get from the library:

```python
import logging

logging.basicConfig(level=logging.DEBUG)
```

kravets-levko:
Hi @wallystart! Yes, logs clearly indicate that library tries to use U2M OAuth flow, which implies opening link in a browser. Unfortunately, that's expected behavior, so if you need to use the library in a notebook - you have to provide access token.

Although, I will keep this issue open - we definitely need to handle this use case (maybe, throw an error if library cannot open an OAuth link).

[PECO-2054]

susodapop:
@kravets-levko one way to handle this is to check whether the user _passed_ a `None` value versus the `None` default being used. Easiest way to achieve this is with a singleton object which you could call `NotPassed`. If you then set the default value in `.connect()` to this value, you can assert that nothing was passed and bail to the oauth flow. Whereas if the user passed a None value you can raise a nice exception that tells them something in their code is out of whack.

susodapop:
I opened a pull request that implements the pattern described above: https://github.com/databricks/databricks-sql-python/pull/462

PR #462, opened 2024-10-29
[issue-458] Log if the user passes a NoneType access token

susodapop:
## Description

This PR implements the pattern described [here](https://github.com/databricks/databricks-sql-python/issues/458#issuecomment-2442887279) to emit a log message if the user _passes_ a None value for their `access_token`, since this is likely not intentional.

This doesn't change the behavior of the connector. In the case of the linked issue, it will not prevent the browser from hanging. But it _will_ notify the user about what's happening behind the scenes.

Relates #458

### DIFF
--- a/src/databricks/sql/__init__.py
+++ b/src/databricks/sql/__init__.py
@@ -10,7 +10,7 @@ paramstyle = "named"
 
 import re
 
-from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


jackyhu-db (approved):
Thanks for the fix.

PR #479, opened 2024-12-11
Incorrect rows in inline fetch result

jprakash-db:
## Description
For inline results there is only support for FETCH_NEXT, FETCH_PRIOR and FETCH_FIRST. FETCH_ABSOLUTE support is not present , so in highly concurrent environment sometimes server moves the pointer for the FETCH_NEXT but the client does not receive that data and any further FETCH_NEXT results in a mismatch in the row offset

## Fix
Till the time FETCH_ABSOLUTE support is added for inline on server side
-  Error is thrown when there is an incorrect row offset in the values fetched
- Changing the behaviour to not retry when there is a FETCH_NEXT orientation in FetchResults

## Testing
Added unit tests to check the non retryable behaviour across different status codes

### DIFF
--- a/src/databricks/sql/auth/retry.py
+++ b/src/databricks/sql/auth/retry.py
@@ -32,7 +32,7 @@ class CommandType(Enum):
     CLOSE_SESSION = "CloseSession"
     CLOSE_OPERATION = "CloseOperation"
     GET_OPERATION_STATUS = "GetOperationStatus"
-    FETCH_RESULTS_ORIENTATION_FETCH_NEXT = "FetchResultsOrientation_FETCH_NEXT"
+    FETCH_RESULTS_INLINE_FETCH_NEXT = "FetchResultsInline_FETCH_NEXT"
     OTHER = "Other"
 
     @classmethod
@@ -363,10 +363,10 @@ class DatabricksRetryPolicy(Retry):
         if status_code == 501:
             raise NonRecoverableNetworkError("Received code 501 from server.")
 
-        if self.command_type == CommandType.FETCH_RESULTS_ORIENTATION_FETCH_NEXT:
+        if self.command_type == CommandType.FETCH_RESULTS_INLINE_FETCH_NEXT:
             return (
                 False,
-                "FetchResults with FETCH_NEXT orientation are not idempotent in inline mode and is not retried",
+                "FetchResults in INLINE mode with FETCH_NEXT orientation are not idempotent and is not retried",
             )
 
         # Request failed and this method is not retryable. We only retry POST requests.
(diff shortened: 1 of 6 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -1046,7 +1046,7 @@ def fetch_results(
 
         resp = self.make_request(self._client.FetchResults, req)
         if resp.results.startRowOffset > expected_row_start_offset:
-            logger.warning(
+            raise ValueError(
                 "Expected results to start from {} but they instead start at {}".format(
```
jackyhu-db, line 1050: can you also add the code to stop retry on fetch request next when it is inline mode

jackyhu-db: as FetchResultNext is not idempotent, so you should not retry regardless of the status code.

src/databricks/sql/auth/retry.py
```diff
@@ -362,6 +363,12 @@ def should_retry(self, method: str, status_code: int) -> Tuple[bool, str]:
         if status_code == 501:
             raise NonRecoverableNetworkError("Received code 501 from server.")
 
+        if self.command_type == CommandType.FETCH_RESULTS_ORIENTATION_FETCH_NEXT:
+            return (
+                False,
+                "FetchResults with FETCH_NEXT orientation are not idempotent and is not retried",
+            )
+
```
jackyhu-db, line 371: It is not idempotent only in INLINE mode

jprakash-db: @jackyhu-db Updated the error message

jackyhu-db: my point is it should return false (stop retrying) if if it NOT INLINE mode.

jprakash-db: @jackyhu-db The issue is that in INLINE FETCH_NEXT is causing the issue and we don't want to retry. Why stop retrying when in NOT INLINE ? Anyways not INLINE uses cloudfetch which does not use FetchResults call

src/databricks/sql/thrift_backend.py
```diff
@@ -156,6 +156,8 @@ def __init__(
         )
 
         # Cloud fetch
+        self._use_cloud_fetch = kwargs.get("use_cloud_fetch", True)
```
jackyhu-db, line 159: `use_cloud_fetch` is passed as the argument of  `execute_command`, please respect that instead of the args in the constructor.

image

jprakash-db: @jackyhu-db updated it

jackyhu-db (approved):

jackyhu-db (commented):
nit: remove this change.

PR #480, opened 2024-12-15
Fixed the Issue of TCP not closing

jprakash-db:
## Description

Currently while closing a Connection the HTTPConnectionPool is not explicitly closed and just the buffer is emptied upon closing the connection. This creates the issue of the TCP connection being in ESTABLISHED phased indefinetly and not moving to the CLOSED phase.

## Fix
- Modified the code to only close the connection in case en error occurs while making a request
- Setting the pool as None to ensure that the Python Garbage cleanup takes care of the final closure and releasing of the port

## Testing 

Tested by monitoring the connection port through the lifecycle of the query to verify its working


samikshya-db:
Are there plans of merging this, Jothi? Looks like this is indeed a nice improvement on the driver.

PR #484, opened 2025-01-02
Remove a mypy error message workaround that should be unnecessary in mypy 1.15.0

wyattscarpenter:
This pr should be good to go next time there's another release of mypy. I'm just keeping it a draft til then.

See https://github.com/python/mypy/pull/17485 or https://github.com/python/mypy/issues/10768 for more information on the problem and workaround that used to be here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ pyarrow = ["pyarrow"]
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.1.2"
-mypy = "^1.10.1"
+mypy = "^1.15.0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -9,7 +9,7 @@ packages = [{ include = "databricks", from = "src" }]
 include = ["CHANGELOG.md"]
 
 [tool.poetry.dependencies]
-python = "^3.8.0"
+python = "^3.9.0"
```
wyattscarpenter:
```suggestion
python = ">=3.9.0"
```

PR #488, opened 2025-01-16
Merging changes from v3.7.1 release

jprakash-db:
## Description
Merging changes from v3.7.1 release

### DIFF
--- a/src/databricks/sql/__init__.py
+++ b/src/databricks/sql/__init__.py
@@ -68,7 +68,7 @@ DATETIME = DBAPITypeObject("timestamp")
 DATE = DBAPITypeObject("date")
 ROWID = DBAPITypeObject()
 
-__version__ = "4.0.0"
+__version__ = "3.7.0"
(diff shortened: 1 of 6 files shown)


src/databricks/sql/__init__.py
```diff
@@ -68,7 +68,7 @@ def __repr__(self):
 DATE = DBAPITypeObject("date")
 ROWID = DBAPITypeObject()
 
-__version__ = "3.7.0"
+__version__ = "4.0.0"
```
jackyhu-db, line 71: should not be 3.7.1

jprakash-db: In the main the final bump up version has not been raised yet. So that is why the main has 3.7.0 and the Changelog has TBD.

jackyhu-db: I know, please this really confuse. can you do a separated PR?

jprakash-db: fixed

jackyhu-db (approved):

PR #497, opened 2025-01-31
Updated retry timeout

jprakash-db:
## Description
- Set the retry timeout to be the max of retry_after and retry_max_delay
- Increased the retry_max_delay


jackyhu-db (approved):

benc-db:
Going to test with my local repro to verify the fix.

benc-db (approved):

benc-db:
Looks reasonable when I test with dbt-databricks.  Can we add a debug log statement that says how long we're going to sleep for?

jprakash-db:
@benc-db Added the logging

PR #504, opened 2025-02-11
Add support for python 3.13 & remove various version caps

dbaxa:
Add support for python 3.13 & remove various version caps

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,8 +18,8 @@ lz4 = "^4.0.2"
 requests = "^2.18.1"
 oauthlib = "^3.1.0"
 numpy = [
-    { version = "^1.16.6", python = ">=3.8,<3.11" },
-    { version = "^1.23.4", python = ">=3.11" },
+    { version = ">=1.16.6", python = ">=3.8,<3.11" },
+    { version = ">=1.23.4", python = ">=3.11" },
 ]
 openpyxl = "^3.0.10"
 urllib3 = ">=1.26"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,8 @@ include = ["CHANGELOG.md"]
 python = "^3.8.0"
 thrift = ">=0.16.0,<0.21.0"
 pandas = [
-    { version = ">=1.2.5,<2.3.0", python = ">=3.8" }
+    { version = ">=1.2.5,<2.3.0", python = ">=3.8,<3.13" },
+    { version = ">=2.2.3,<2.3.0", python = ">=3.13" }
 ]
 lz4 = "^4.0.2"
 requests = "^2.18.1"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -12,18 +12,22 @@ include = ["CHANGELOG.md"]
...
+    { version = ">=1.2.5,<2.3.0", python = ">=3.8,<3.13" },
+    { version = ">=2.2.3,<2.3.0", python = ">=3.13" }
 ]
 lz4 = "^4.0.2"
 requests = "^2.18.1"
 oauthlib = "^3.1.0"
 numpy = [
-    { version = "^1.16.6", python = ">=3.8,<3.11" },
-    { version = "^1.23.4", python = ">=3.11" },
+    { version = ">=1.16.6", python = ">=3.8,<3.11" },
```
dbaxa, line 22: For some reason https://github.com/databricks/databricks-sql-python/commit/01e998cabe4a4972d5b45f7936219d924706c75c#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711 reverted the change made in d690516 to remove the version caps.

jprakash-db:
@dbaxa  Python3.13 testing can be included. As far as relaxing the numpy version we reverted it as it had compatibility issues with pandas  - https://pandas.pydata.org/docs/whatsnew/v2.2.2.html#pandas-2-2-2-is-now-compatible-with-numpy-2-0

dbaxa:
@jprakash-db can you provide more details about the compatibility issues? In our use case we didn't see any problems with using python 3.13 pandas 2.2.3 & numpy 2.2.3 when using a modified version of `databricks-sql-python` locally.

dhirschfeld:
As `databricks-sql-python` doesn't directly import `numpy` we shouldn't be specifying it as a dependency here.

In this code base we only rely on `pandas` functionality, we don't care how pandas implements that functionality under the hood.

Our contract is with the public API of `pandas` and we trust the `pandas` devs to ensure their library works correctly. If `pandas` doesn't work with a modern `numpy` it is for them to pin `numpy` as it is their dependency.

`pandas` claims to support `numpy>=1.26`:
 https://github.com/pandas-dev/pandas/blob/main/pyproject.toml#L30-L31
  
...so if that is not the case we should open an issue (or PR) in the `pandas` repo to apply the pin there.

Packaging bugs occur (all the time!) but if this pin here is to work around a mis-specified dependency that `pandas` once had at one time, the cure is worse that the cold. Pinning to an old `numpy` means that `databricks-sql-connector` can't be used in any modern analytics environment which isn't great for a package designed to support analytics.

samikshya-db:
Thanks for raising this : [dhirschfeld](https://github.com/dhirschfeld), this makes sense to me. I am closing this in favour of https://github.com/databricks/databricks-sql-python/pull/510

PR #505, opened 2025-02-16
Support for enforcing embedded schema

jprakash-db:
## Description

As a part of query optimization Databricks returns cached results when the underlying query is the same. This creates issue when the same query is used but with a different alias, in this case the result is the same but the metadata is different. But Databricks returns the same metadata as a part of query optimisation

## Added 
Added option to enforce schema to make sure the metadata is in updated state

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -737,6 +737,7 @@ class Cursor:
         self,
         operation: str,
         parameters: Optional[TParameterCollection] = None,
+        enforceEmbeddedSchema=False,
     ) -> "Cursor":
         """
         Execute a query and wait for execution to complete.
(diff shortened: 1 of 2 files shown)


gopalldb (approved):

src/databricks/sql/client.py
```diff
@@ -737,6 +737,7 @@ def execute(
         self,
         operation: str,
         parameters: Optional[TParameterCollection] = None,
+        enforceEmbeddedSchema=False,
```
jackyhu-db, line 740: can you rename it to `enforce_embedded_schema_correctness` to be consistent with the meaning of the thrift API. "enforce schema" has different with "enforce schema with correctness". And please use snake case which is python style.

PR #511, opened 2025-02-19
Added info logging for response

jprakash-db:
## Description
Added logging for status code and method details for easy debugging

### DIFF
--- a/src/databricks/sql/auth/retry.py
+++ b/src/databricks/sql/auth/retry.py
@@ -345,6 +345,8 @@ class DatabricksRetryPolicy(Retry):
         if a retry would violate the configured policy.
         """
 
+        logger.info(f"Received status code {status_code} for {method} request")
+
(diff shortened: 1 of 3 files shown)


src/databricks/sql/auth/retry.py
```diff
@@ -345,6 +345,8 @@ def should_retry(self, method: str, status_code: int) -> Tuple[bool, str]:
         if a retry would violate the configured policy.
         """
 
+        logger.info(f"Received status code {status_code} for {method} request")
```
benc-db, line 348: When we get 4xx or 5xx, do we only get a status code, or do we also get an error message of some sort?

jprakash-db: @benc-db I have updated such that the error messages from the server are printed and not our custom messages

benc-db (approved):

PR #515, opened 2025-02-25
Fix : Unable to poll results in async API

jprakash-db:
## Description
Polling using the async API returned the `RESOURCE_NOT_FOUND` error for async executed queries

## Fix
Force retention of data on the server side by not returning the results in the `direct results` within the `execute_query` response

## Testing
Added e2e tests

### DIFF
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -899,9 +899,11 @@ class ThriftBackend:
             statement=operation,
             runAsync=True,
             # For async operation we don't want the direct results
-            getDirectResults=ttypes.TSparkGetDirectResults(
-                maxRows=0 if async_op else max_rows,
-                maxBytes=0 if async_op else max_bytes,
+            getDirectResults=None
+            if async_op
+            else ttypes.TSparkGetDirectResults(
+                maxRows=max_rows,
+                maxBytes=max_bytes,
             ),
             canReadArrowResult=True if pyarrow else False,
             canDecompressLZ4Result=lz4_compression,
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -898,8 +898,10 @@ class ThriftBackend:
             sessionHandle=session_handle,
             statement=operation,
             runAsync=True,
+            # For async operation we don't want the direct results
             getDirectResults=ttypes.TSparkGetDirectResults(
-                maxRows=max_rows, maxBytes=max_bytes
+                maxRows=0 if async_op else max_rows,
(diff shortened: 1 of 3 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -898,8 +898,10 @@ def execute_command(
             sessionHandle=session_handle,
             statement=operation,
             runAsync=True,
+            # For async operation we don't want the direct results
             getDirectResults=ttypes.TSparkGetDirectResults(
-                maxRows=max_rows, maxBytes=max_bytes
+                maxRows=0 if async_op else max_rows,
```
samikshya-db, line 903: we should add a unit test

jayantsing-db, line 903: Instead of setting maxRows as 0, I would recommend not specifying the direct result object in the request altogether. This is to make the intention clear of not expecting the direct results.

jprakash-db: Added e2e tests

jprakash-db: Have made the object None

src/databricks/sql/thrift_backend.py
```diff
@@ -898,8 +898,12 @@ def execute_command(
...
             runAsync=True,
-            getDirectResults=ttypes.TSparkGetDirectResults(
-                maxRows=max_rows, maxBytes=max_bytes
+            # For async operation we don't want the direct results
+            getDirectResults=None
+            if async_op
+            else ttypes.TSparkGetDirectResults(
+                maxRows=max_rows,
+                maxBytes=max_bytes,
             ),
```
jayantsing-db, line 907: do we have a linter for python driver yet? if not, could you please open an issue to community or internally?

jprakash-db: @jayantsing-db I can raise a new PR for that

tests/e2e/common/large_queries_mixin.py
```diff
@@ -94,7 +94,7 @@ def test_long_running_query(self):
         scale_factor = 1
         with self.cursor() as cursor:
             while duration < min_duration:
-                assert scale_factor < 512, "Detected infinite loop"
+                assert scale_factor < 1024, "Detected infinite loop"
```
jayantsing-db, line 97: Could you explain this change?

jprakash-db: For long running queries the previous upper limit of scale factor was not sufficient to run queries for atleast 5 minutes, hence increased it

samikshya-db (approved):

PR #521, opened 2025-02-27
Updates runner used to run DCO check to use databricks-protected-runner

madhav-db:
Updates runner used to run DCO check to use databricks-protected-runner.

Before this, we were running into IP based access issues. This should resolve them.


jprakash-db (approved):
Thanks for making these changes, LGTM

madhav-db:
jenkins merge

PR #530, opened 2025-03-07
[ES-1372353] make user_agent_header part of public API

shivam2680:
## Description

- Renamed `_user_agent_entry` in connect call args to `user_agent_entry`.
- removed this from list of internal arguments in docstring.

## How is this tested?
Verified agent population in query history when using `user_agent_entry`

ES-1372353

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -234,7 +234,7 @@ class Connection:
             if user_agent_entry is not None:
                 logger.warning(
                     "[WARN] Parameter '_user_agent_entry' is deprecated; use 'user_agent_entry' instead. "
-                    "This parameter will be removed in the next release."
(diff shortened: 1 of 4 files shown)


samikshya-db (approved):

src/databricks/sql/client.py
```diff
@@ -122,6 +122,9 @@ def __init__(
                 port of the oauth redirect uri (localhost). This is required when custom oauth client_id
                 `oauth_client_id` is set
 
+            user_agent_entry: `str`, optional
+                Tag to add to User-Agent header. For use by partners. If not specified, it will use the default user agent PyDatabricksSqlConnector
```
samikshya-db, line 126: we need a better documentation here.

jprakash-db (approved):
Thanks for making the changes. LGTM

PR #533, opened 2025-03-07
Support multiple timestamp formats in non arrow flow

jprakash-db:
## Description
Adding support for multiple timestamp formats parsing using [python-dateutil](https://pypi.org/project/python-dateutil/#description)

## Testing
Added unit tests for the most popular timestamp formats used ata Databricks


src/databricks/sql/utils.py
```diff
@@ -628,6 +628,26 @@ def convert_decimals_in_arrow_table(table, description) -> "pyarrow.Table":
     return table
 
 
+def datetime_parser(datetime_string):
+    formats = [
+        "%Y-%m-%d %H:%M:%S.%f",  # MySQL and PostgreSQL format
+        "%Y-%m-%d %H:%M:%S",  # Common SQL format
+        "%Y-%m-%d",  # MySQL Date only
+        "%m/%d/%Y %I:%M:%S %p",  # US format MSSQL Server
```
samikshya-db, line 636: this does not contain `"2023-09-10T20:45:00Z"`. Can we add more types? Also do we support `TimeStampMicroTZVector`?

jprakash-db: @samikshya-db We can add as many as we need. I have added only the popular ones

samikshya-db: As long as you confirm that we have everything that comes in from the server, I am good.

PR #554, opened 2025-05-13
Enhance Cursor close handling and context manager exception management to prevent server side resource leaks

madhav-db:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Other

## Description

This PR attempts to fix an issue where database connections were not properly closing, resulting in "lame duck" sessions that kept SQL warehouses active unnecessarily.

When clients disconnected, operation handles were not properly being closed on the server side, potentially leading to resource leaks. The `Cursor.close()` methodwas simply nullifying the operation handle reference without properly closing it on the thrift backend.

Besides this, in `Cursor.exit()`, added better error handling to prevent exceptions during cursor closure from suppressing original exceptions, which is particularly important for user introduced interruptions (eg. Keyboard interrupt). Before this fix, when users interrupted long-running queries, the exception from cursor.close() could mask the KeyboardInterrupt, preventing the application from properly responding to cancellation requests and potentially leaving resources uncleaned.

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

Added new unit tests to verify behaviour

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -1179,13 +1179,10 @@ class Cursor:
                 self.thrift_backend.close_command(self.active_op_handle)
             except RequestError as e:
                 if isinstance(e.args[1], CursorAlreadyClosedError):
-                    # This is the specific message expected by the test
                     logger.info("Operation was canceled by a prior request")
                 else:
-                    # For other types of errors, keep the generic logging
                     logging.warning(f"Error closing operation handle: {e}")
             except Exception as e:
-                # For non-RequestError exceptions
                 logging.warning(f"Error closing operation handle: {e}")
             finally:
                 self.active_op_handle = None
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -1172,7 +1172,7 @@ class Cursor:
     def close(self) -> None:
         """Close cursor"""
         self.open = False
-        
+
         # Close active operation handle if it exists
         if self.active_op_handle:
             try:
@@ -1182,7 +1182,7 @@ class Cursor:
                 logging.warning(f"Error closing operation handle: {e}")
             finally:
                 self.active_op_handle = None
-                
+
         if self.active_result_set:
             self._close_and_clear_active_result_set()
 
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -315,7 +315,13 @@ class Connection:
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.close()
+        try:
+            self.close()
+        except BaseException as e:
+            logger.warning(f"Exception during connection close in __exit__: {e}")
+            if exc_type is None:
+                raise
+        return False
 
     def __del__(self):
         if self.open:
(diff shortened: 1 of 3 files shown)


gopalldb (commented):
Apart from unit tests, any other way to test this, especially E2E?

gopalldb (approved):

src/databricks/sql/client.py
```diff
@@ -456,7 +456,16 @@ def __enter__(self) -> "Cursor":
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.close()
+        try:
```
jprakash-db, line 459: The default behaviour is anyway to not suppress an error. The `False` is just redundant. The logging is useful but catching exception and then throwing it again is the default behaviour and explicit handling don't think adds anything valuable.

jprakash-db, line 459: I don't feel we are adding anything here. From the code it doesn't look like `self.close()` will ever throw an error. Every step error is handled.

src/databricks/sql/client.py
```diff
@@ -456,7 +456,16 @@ def __enter__(self) -> "Cursor":
...
+            logger.debug("Cursor context manager exiting, calling close()")
+            self.close()
+        except Exception as e:
+            logger.warning(f"Exception during cursor close in __exit__: {e}")
+            # Don't suppress the original exception if there was one
+            if exc_type is None:
+                # Only raise our new exception if there wasn't already one in progress
+                raise
+        return False
```
jprakash-db, line 468: Since `self.close()` doesn't throw an error I believe this error handling doesn't make sense, as default behaviour is to anyway throw the error within the with block

src/databricks/sql/client.py
madhav-db: It can, this handles stuff like keyboard interrupts. I've written supporting tests for the same. If for example an interrupt occurs, then the cursor close doesn't throw an exception, but it is an external interrupt, and the connection still closes since we don't throw an exception and return False.

src/databricks/sql/client.py
madhav-db: Explained above

src/databricks/sql/client.py
samikshya-db: How can this be reproduced?

madhav-db: Discussed offline, changed `Exception` to `BaseException`

jprakash-db: @madhav-db  Thanks
Had 2 doubts / suggestions
- What happens if the user repeatedly does `KeyboardInterrupt`, I believe this will fail as we are handling only one interrupt
- I don't think we should write code for `KeyboardInterrupt` as it  is a hard fail, having try catch blocks will make it confusing to read the code when there are no code exceptions being thrown

madhav-db: 1. I believe that's okay. I don't think this change here is intended for someone trying to intentionally break their use case.
2. This isn't just KeyboardInterrupt, it's for any sig interrupts, and it being a hard fail is I believe irrelevant since this is done for closing objects gracefully.

jprakash-db (approved):

PR #559, opened 2025-05-19
Support for complex params

jprakash-db:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A


src/databricks/sql/parameters/native.py
```diff
@@ -428,6 +444,99 @@ def __init__(self, value: int, name: Optional[str] = None):
...
+        """Returns a TSparkParameter object that can be passed to the DBR thrift server."""
+
+        tsp = TSparkParameter(type=self._cast_expr())
+        tsp.arguments = [val._tspark_value_arg() for val in self.value]
+
+        if named:
+            tsp.name = self.name
+            tsp.ordinal = False
+        elif not named:
+            tsp.ordinal = True
```
shivam2680, line 480: isn't ordinal an int field? Should we set it to parameter index here?

src/databricks/sql/parameters/native.py
```diff
@@ -428,6 +444,99 @@ def __init__(self, value: int, name: Optional[str] = None):
...
+            ```
+            If not None, your query should contain a named parameter marker. Like:
+            ```sql
+                SELECT * FROM table WHERE field = :my_param
+            ```
+
+            The `name` argument to this function would be `my_param`.
+        """
+        self.name = name
+        self.value = [dbsql_parameter_from_primitive(val) for val in value]
```
shivam2680, line 468: are we able to handle nested types like array<map<int,int>> this way?

src/databricks/sql/parameters/native.py
```diff
@@ -543,23 +652,26 @@ def dbsql_parameter_from_primitive(
...
+    elif isinstance(value, datetime.date):
         return DateParameter(value=value, name=name)
-    elif type(value) is bool:
-        return BooleanParameter(value=value, name=name)
-    elif type(value) is decimal.Decimal:
+    elif isinstance(value, decimal.Decimal):
         return DecimalParameter(value=value, name=name)
+    elif isinstance(value, dict):
+        return MapParameter(value=value, name=name)
+    elif isinstance(value, Sequence) and not isinstance(value, str):
```
shivam2680, line 671: `Range` is also a Sequence in python, are we allowing that also as parameter?

src/databricks/sql/utils.py
```diff
@@ -429,7 +429,7 @@ def user_friendly_error_message(self, no_retry_reason, attempt, elapsed):
 # Taken from PyHive
 class ParamEscaper:
     _DATE_FORMAT = "%Y-%m-%d"
-    _TIME_FORMAT = "%H:%M:%S.%f"
+    _TIME_FORMAT = "%H:%M:%S.%f %z"
```
shivam2680, line 432: Timezone will not be there for TIMESTAMP_NTZ param.

src/databricks/sql/utils.py
```diff
@@ -458,13 +458,22 @@ def escape_string(self, item):
...
+            (element for key, value in item.items() for element in (key, value)),
+        )
+        l = list(map(str, l))
+        return "MAP(" + ",".join(l) + ")"
 
     def escape_datetime(self, item, format, cutoff=0):
         dt_str = item.strftime(format)
         formatted = dt_str[:-cutoff] if cutoff and format.endswith(".%f") else dt_str
-        return "'{}'".format(formatted)
+        return "'{}'".format(formatted.strip())
```
shivam2680, line 476: why this change?

src/databricks/sql/parameters/native.py
jprakash-db: No, in this case we don't need to set the index. ordinal is only supported when the params have `?`, so it auto fills the variables in the order they appear

src/databricks/sql/utils.py
jprakash-db: while using striptime, we get a extra space after the text. This just removes it

src/databricks/sql/parameters/native.py
jprakash-db: yes, it handles the recursive cases. Tests are also added to support this

src/databricks/sql/parameters/native.py
jprakash-db: No, we don't want users to pass lazy types like `Range`. They are expected to pass initialized values like `list` or `tuples`

tests/e2e/test_parameterized_queries.py
```diff
@@ -166,10 +179,9 @@ def _inline_roundtrip(self, params: dict, paramstyle: ParamStyle):
         :paramstyle:
             This is a no-op but is included to make the test-code easier to read.
         """
-        target_column = self._get_inline_table_column(params.get("p"))
-        INSERT_QUERY = f"INSERT INTO pysql_e2e_inline_param_test_table (`{target_column}`) VALUES (%(p)s)"
-        SELECT_QUERY = f"SELECT {target_column} `col` FROM pysql_e2e_inline_param_test_table LIMIT 1"
-        DELETE_QUERY = "DELETE FROM pysql_e2e_inline_param_test_table"
+        INSERT_QUERY = f"INSERT INTO ___________________first.jprakash.pysql_e2e_inline_param_test_table (`{target_column}`) VALUES (%(p)s)"
```
vikrantpuppala: why are we using this ______first catalog and jprakash namespace? can you move into a team catalog and schema?

src/databricks/sql/utils.py
vikrantpuppala: +1, would like to know how we've accounted/tested for NTZ

src/databricks/sql/parameters/native.py
shivam2680: it's better to add `and not isinstance(value, Range)`

tests/e2e/test_parameterized_queries.py
gopalldb: +1

src/databricks/sql/parameters/native.py
jprakash-db: We anyways convert Range to list(Range) which expands it, so it is handled

src/databricks/sql/utils.py
jprakash-db: If it is not there it will be an empty space, there is already a test suite that inserts NTZ and none NTZ and reads back to compare whether it is equal or not

jprakash-db: @vikrantpuppala @shivam2680 There are already existing tests that insert NTZ and non NTZ values and reads back from table to ensure everything is working as expected - https://github.com/databricks/databricks-sql-python/blob/384258364d45e3088c861e99495267ee600a280e/tests/e2e/test_parameterized_queries.py#L283

gopalldb (approved):
Thanks!

PR #560, opened 2025-05-19
add variant support

shivam2680:
## Description
This pull request introduces support for detecting and handling VARIANT column types in the Databricks SQL Thrift backend, along with corresponding tests for validation. 
updated the `_col_to_description` and `_hive_schema_to_description` methods to process metadata for VARIANT types, integration of Arrow schemas
Added unit and end-to-end tests to ensure proper functionality.

### End-to-End Tests:
* Added `tests/e2e/test_variant_types.py` to validate VARIANT type detection and data retrieval. Includes tests for creating tables with VARIANT columns, inserting records, and verifying correct type handling and JSON parsing.

### Unit Tests:
* Tests cover scenarios like VARIANT type detection, handling of null or malformed metadata, and fallback behavior for missing Arrow schemas.

### DIFF
--- a/src/databricks/sql/thrift_backend.py
+++ b/src/databricks/sql/thrift_backend.py
@@ -665,7 +665,7 @@ class ThriftBackend:
         return pyarrow.schema([convert_col(col) for col in t_table_schema.columns])
 
     @staticmethod
-    def _col_to_description(col):
+    def _col_to_description(col, field):
         type_entry = col.typeDesc.types[0]
 
         if type_entry.primitiveEntry:
@@ -692,12 +692,36 @@ class ThriftBackend:
         else:
             precision, scale = None, None
 
+        # Extract variant type from field if available
+        if field is not None:
+            try:
+                # Check for variant type in metadata
+                if field.metadata and b"Spark:DataType:SqlName" in field.metadata:
+                    sql_type = field.metadata.get(b"Spark:DataType:SqlName")
+                    if sql_type == b"VARIANT":
+                        cleaned_type = "variant"
+            except Exception as e:
+                logger.debug(f"Could not extract variant type from field: {e}")
+
         return col.columnName, cleaned_type, None, None, precision, scale, None
 
     @staticmethod
-    def _hive_schema_to_description(t_table_schema):
+    def _hive_schema_to_description(t_table_schema, schema_bytes=None):
+        # Create a field lookup dictionary for efficient column access
+        field_dict = {}
+        if pyarrow and schema_bytes:
+            try:
+                arrow_schema = pyarrow.ipc.read_schema(pyarrow.py_buffer(schema_bytes))
+                # Build a dictionary mapping column names to fields
+                for field in arrow_schema:
+                    field_dict[field.name] = field
+            except Exception as e:
+                logger.debug(f"Could not parse arrow schema: {e}")
+
+        # Process each column with its corresponding Arrow field (if available)
         return [
-            ThriftBackend._col_to_description(col) for col in t_table_schema.columns
+            ThriftBackend._col_to_description(col, field_dict.get(col.columnName))
+            for col in t_table_schema.columns
         ]
 
     def _results_message_to_execute_response(self, resp, operation_state):
@@ -726,9 +750,6 @@ class ThriftBackend:
             or (not direct_results.resultSet)
             or direct_results.resultSet.hasMoreRows
         )
-        description = self._hive_schema_to_description(
-            t_result_set_metadata_resp.schema
-        )
 
         if pyarrow:
             schema_bytes = (
@@ -739,6 +760,10 @@ class ThriftBackend:
             )
         else:
             schema_bytes = None
+
+        description = self._hive_schema_to_description(
+            t_result_set_metadata_resp.schema, schema_bytes
+        )
 
         lz4_compressed = t_result_set_metadata_resp.lz4Compressed
         is_staging_operation = t_result_set_metadata_resp.isStagingOperation
@@ -793,9 +818,6 @@ class ThriftBackend:
         lz4_compressed = t_result_set_metadata_resp.lz4Compressed
         is_staging_operation = t_result_set_metadata_resp.isStagingOperation
         has_more_rows = resp.hasMoreRows
-        description = self._hive_schema_to_description(
-            t_result_set_metadata_resp.schema
-        )
 
         if pyarrow:
             schema_bytes = (
@@ -806,6 +828,10 @@ class ThriftBackend:
             )
         else:
             schema_bytes = None
+
+            description = self._hive_schema_to_description(
+                t_result_set_metadata_resp.schema, schema_bytes
+            )
 
         queue = ResultSetQueueFactory.build_queue(
             row_set_type=resp.resultSetMetadata.resultFormat,
--- a/tests/e2e/test_variant_types.py
+++ b/tests/e2e/test_variant_types.py
@@ -1,0 +1,97 @@
+import pytest
+from datetime import datetime
+import json
+try:
+    import pyarrow
+except ImportError:
+    pyarrow = None
+
+from tests.e2e.test_driver import PySQLPytestTestCase
+
+class TestVariantTypes(PySQLPytestTestCase):
+    """Tests for the proper detection and handling of VARIANT type columns"""
+
+    @pytest.fixture(scope="class")
+    def variant_table_fixture(self, connection_details):
+        self.arguments = connection_details.copy()
+        """A pytest fixture that creates a table with variant columns, inserts records, yields, and then drops the table"""
+
+        with self.cursor() as cursor:
+            # Check if VARIANT type is supported
+            try:
+                # delete the table if it exists
+                cursor.execute("DROP TABLE IF EXISTS pysql_test_variant_types_table")
+
+                # Create the table with variant columns
+                cursor.execute(
+                    """
+                    CREATE TABLE IF NOT EXISTS pysql_test_variant_types_table (
+                        id INTEGER,
+                        variant_col VARIANT,
+                        regular_string_col STRING
+                    )
+                    """
+                )
+                
+                # Insert test records with different variant values
+                cursor.execute(
+                    """
+                    INSERT INTO pysql_test_variant_types_table
+                    VALUES 
+                    (1, PARSE_JSON('{"name": "John", "age": 30}'), 'regular string'),
+                    (2, PARSE_JSON('[1, 2, 3, 4]'), 'another string')
+                    """
+                )
+                
+                variant_supported = True
+            except Exception as e:
+                # VARIANT type not supported in this environment
+                print(f"VARIANT type not supported: {e}")
+                variant_supported = False
+            
+            yield variant_supported
+            
+            # Clean up if table was created
+            if variant_supported:
+                cursor.execute("DROP TABLE IF EXISTS pysql_test_variant_types_table")
+
+    def test_variant_type_detection(self, variant_table_fixture):
+        """Test that VARIANT type columns are properly detected"""
+        if not variant_table_fixture:
+            pytest.skip("VARIANT type not supported in this environment")
+            
+        with self.cursor() as cursor:
+            cursor.execute("SELECT * FROM pysql_test_variant_types_table LIMIT 1")
+            
+            # Check that the column type is properly detected as 'variant'
+            assert cursor.description[1][1] == 'variant', "VARIANT column type not correctly identified"
+            
+            # Regular string column should still be reported as string
+            assert cursor.description[2][1] == 'string', "Regular string column type not correctly identified"
+
+    def test_variant_data_retrieval(self, variant_table_fixture):
+        """Test that VARIANT data is properly retrieved and can be accessed as JSON"""
+        if not variant_table_fixture:
+            pytest.skip("VARIANT type not supported in this environment")
+            
+        with self.cursor() as cursor:
+            cursor.execute("SELECT * FROM pysql_test_variant_types_table ORDER BY id")
+            rows = cursor.fetchall()
+            
+            # First row should have a JSON object
+            json_obj = rows[0][1]
+            assert isinstance(json_obj, str), "VARIANT column should be returned as string"
+            
+            # Parsing to verify it's valid JSON
+            parsed = json.loads(json_obj)
+            assert parsed.get('name') == 'John'
+            assert parsed.get('age') == 30
+            
+            # Second row should have a JSON array
+            json_array = rows[1][1]
+            assert isinstance(json_array, str), "VARIANT array should be returned as string"
+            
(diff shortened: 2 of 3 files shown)


src/databricks/sql/thrift_backend.py
```diff
@@ -692,12 +692,36 @@ def _col_to_description(col):
         else:
             precision, scale = None, None
 
+        # Extract variant type from field if available
```
jprakash-db: Are you sure this is correct? I tried and was getting `metadata` as null when the column type is `variant`. Also for variant the `pyarrow` schema just shows `string` in my testing, shouldn't the server return `variant` type ?

tests/e2e/test_variant_types.py
```diff
@@ -0,0 +1,174 @@
...
+                cursor.execute(
+                    """
+                    INSERT INTO pysql_test_variant_types_table
+                    VALUES 
+                    (1, PARSE_JSON('{"name": "John", "age": 30}'), 'regular string'),
+                    (2, PARSE_JSON('[1, 2, 3, 4]'), 'another string')
+                    """
+                )
+                
+                variant_supported = True
```
jprakash-db: I don't understand the point of the test. If table creation passes then `variant` is supported otherwise it is not? Table creation can fail due to many reasons, how is this a good test ?

tests/e2e/test_variant_types.py
```diff
@@ -0,0 +1,174 @@
...
+            except Exception as e:
+                # VARIANT type not supported in this environment
+                print(f"VARIANT type not supported: {e}")
+                variant_supported = False
+            
+            yield variant_supported
+            
+            # Clean up if table was created
+            if variant_supported:
```
jprakash-db: Best practice is to always use a 
`
try:
    yield
finally:
    Cleanup table
`
In your code, if code using the yield throws as error the table will not be dropped

tests/e2e/test_variant_types.py
```diff
@@ -0,0 +1,174 @@
...
+    @pytest.fixture(scope="class")
+    def variant_table_fixture(self, connection_details):
+        self.arguments = connection_details.copy()
+        """A pytest fixture that creates a table with variant columns, inserts records, yields, and then drops the table"""
+
+        with self.cursor() as cursor:
+            # Check if VARIANT type is supported
+            try:
+                # delete the table if it exists
+                cursor.execute("DROP TABLE IF EXISTS pysql_test_variant_types_table")
```
jprakash-db: Don't add so many drop checks. Once before the table creation and once after

tests/e2e/test_variant_types.py
```diff
@@ -0,0 +1,174 @@
...
+            (15, '[]', [], "Empty array"),
+            (16, '{"unicode": "✓ öäü 😀"}', {"unicode": "✓ öäü 😀"}, "Unicode characters"),
+            (17, '{"large_number": 9223372036854775807}', {"large_number": 9223372036854775807}, "Large integer"),
+            
+            # Deeply nested structure
+            (18, '{"level1": {"level2": {"level3": {"level4": {"level5": "deep value"}}}}}',
+             {"level1": {"level2": {"level3": {"level4": {"level5": "deep value"}}}}}, "Deeply nested structure"),
+             
+            # Date and time types
+            (19, '"2023-01-01"', "2023-01-01", "Date as string (ISO format)"),
```
jprakash-db: How is this test related to this PR? In these tests you are testing whether json.loads is working properly, don't see the point of testing `json.loads()` ability

tests/e2e/test_variant_types.py
```diff
@@ -0,0 +1,174 @@
...
+            (22, '"2023-01-01T12:34:56Z"', "2023-01-01T12:34:56Z", "Datetime with Z timezone (UTC)"),
+            (23, '"2023-01-01T12:34:56+02:00"', "2023-01-01T12:34:56+02:00", "Datetime with timezone offset"),
+            (24, '{"date": "2023-01-01", "time": "12:34:56"}', {"date": "2023-01-01", "time": "12:34:56"}, "Object with date and time fields"),
+            (25, '["2023-01-01", "2023-02-02", "2023-03-03"]', ["2023-01-01", "2023-02-02", "2023-03-03"], "Array of dates"),
+            (26, '{"events": [{"timestamp": "2023-01-01T12:34:56Z", "name": "event1"}, {"timestamp": "2023-02-02T12:34:56Z", "name": "event2"}]}',
+                 {"events": [{"timestamp": "2023-01-01T12:34:56Z", "name": "event1"}, {"timestamp": "2023-02-02T12:34:56Z", "name": "event2"}]},
+                 "Complex object with timestamps"),
+        ]
+    )
+    def test_variant_data_types(self, test_id, json_value, expected_result, description):
```
jprakash-db: Don't think this test should be there. json.loads() is a well tested library and there is no need to test whether it parses correctly or not

src/databricks/sql/thrift_backend.py
shivam2680: yes,
debug output:
[SHIVAM] field  pyarrow.Field<CAST(1 AS VARIANT): string>
[SHIVAM] field metadata  {b'Spark:DataType:SqlName': b'VARIANT', b'Spark:DataType:JsonType': b'"variant"'}

jprakash-db: @shivam2680 I am getting this as the arrow_schema, where metadata is null. Is this some transient behaviour ? or am I missing something 
Screenshot 2025-06-17 at 1 43 18 PM

src/databricks/sql/thrift_backend.py
```diff
@@ -702,12 +702,36 @@ def _col_to_description(col):
...
+        if field is not None:
+            try:
+                # Check for variant type in metadata
+                if field.metadata and b"Spark:DataType:SqlName" in field.metadata:
+                    sql_type = field.metadata.get(b"Spark:DataType:SqlName")
+                    if sql_type == b"VARIANT":
+                        cleaned_type = "variant"
+            except Exception as e:
+                logger.debug(f"Could not extract variant type from field: {e}")
+
```
jayantsing-db: please check with eng-sqlgateway if there is a way to get this from thrift metadata. python connector uses thrift metadata for getting metadata

jayantsing-db (changes requested):

src/databricks/sql/thrift_backend.py
jayantsing-db: is there is some documentation/contract around it or is it purely from empirical evidence?

shivam2680: these schema bytes are read from `t_result_set_metadata_resp.arrowSchema` itself. Please refer to https://sourcegraph.prod.databricks-corp.com/databricks/databricks-sql-python/-/blob/src/databricks/sql/backend/thrift_backend.py?L812

tests/e2e/test_variant_types.py
```diff
@@ -0,0 +1,80 @@
...
+import json
+try:
+    import pyarrow
+except ImportError:
+    pyarrow = None
+
+from tests.e2e.test_driver import PySQLPytestTestCase
+from tests.e2e.common.predicates import pysql_supports_arrow
+
+class TestVariantTypes(PySQLPytestTestCase):
```
jprakash-db, line 15: why don't you add skip tests if no arrow over here itself, why to add on each test.

tests/unit/test_thrift_backend.py
```diff
@@ -2356,6 +2356,149 @@ def test_execute_command_sets_complex_type_fields_correctly(
                 t_execute_statement_req.useArrowNativeTypes.intervalTypesAsArrow
             )
 
+    def test_col_to_description_with_variant_type(self):
```
jprakash-db: There is too much code duplication, like multiple tests just need different arguments. Can you use pytest fuxtures with arguments

tests/unit/test_thrift_backend.py
```diff
@@ -2356,6 +2356,149 @@ def test_execute_command_sets_complex_type_fields_correctly(
...
+        # Verify the result has string as the type (unchanged)
+        self.assertEqual(result[0], "missing_field")  # Column name
+        self.assertEqual(result[1], "string")        # Type name (should be string)
+        self.assertIsNone(result[2])                # No display size
+        self.assertIsNone(result[3])                # No internal size
+        self.assertIsNone(result[4])                # No precision
+        self.assertIsNone(result[5])                # No scale
+        self.assertIsNone(result[6])                # No null ok
+
+    def test_hive_schema_to_description_with_arrow_schema(self):
```
jprakash-db: Same comment, can you just write a single test for col_to_description with multiple arguments. Then a single test for hive_schema_to_description with multiple arguments

tests/unit/test_thrift_backend.py
```diff
@@ -2356,6 +2356,149 @@ def test_execute_command_sets_complex_type_fields_correctly(
...
+
+    def test_col_to_description_without_variant_type(self):
+        # Test normal column without variant type
+        col = ttypes.TColumnDesc(
+            columnName="normal_col",
+            typeDesc=self._make_type_desc(ttypes.TTypeId.STRING_TYPE),
+        )
+        
+        # Create a normal field without variant metadata
+        field = pyarrow.field(
```
jprakash-db: What will happen here if arrow is not installed? You are not skipping this tests

tests/unit/test_thrift_backend.py
```diff
@@ -2356,6 +2356,149 @@ def test_execute_command_sets_complex_type_fields_correctly(
...
+                columnName="variant_col",
+                typeDesc=self._make_type_desc(ttypes.TTypeId.STRING_TYPE),
+            ),
+        ]
+        t_table_schema = ttypes.TTableSchema(columns=columns)
+        
+        # Create an Arrow schema with one variant column
+        fields = [
+            pyarrow.field("regular_col", pyarrow.string()),
+            pyarrow.field(
```
jprakash-db: same comment, shouldn't these tests be skipped without arrow?

jprakash-db (commented):
can you plz incorporate `@pytest.mark.parametrize` , I feel a lot of test code duplication can be avoided

tests/unit/test_thrift_backend.py
shivam2680: done

tests/unit/test_thrift_backend.py
shivam2680: done.

jprakash-db (approved):

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -764,12 +764,39 @@ def _col_to_description(col, session_id_hex=None):
...
+                logger.debug(f"Could not parse arrow schema: {e}")
+
         return [
-            ThriftDatabricksClient._col_to_description(col, session_id_hex)
+            ThriftDatabricksClient._col_to_description(
+                col,
+                field_dict.get(col.columnName) if field_dict else None,
+                session_id_hex,
+            )
             for col in t_table_schema.columns
```
jayantsing-db, line 800: will these changes work for SEA backend as well? cc @Varun0157

Varun0157: SEA does not return schema bytes, so we can't apply the changes in the same manner. 

The SEA backend already converts `type_name`s into lowercase, so if the only change to be made is to return `variant` instead of `VARIANT` for relevant fields, then this should already be working for the SEA backend, because [SEA returns `VARIANT` in the `type_name`](https://github.com/databricks/databricks-jdbc/blob/acad168dddd1350862f6873186583f55c5b28606/src/test/resources/sqlexecapi/datatypesintegrationtests/testvarianttypes/mappings/api_2.0_sql_statements-bf32f446-2dd6-4d2f-9d76-54a1e2dda706.json#L15).

jayantsing-db: schema bytes are returned inline (embedded) to arrow cloud-fetch files i think

additionally, no variant type info in result manifest for SEA?

jayantsing-db: >The SEA backend already converts type_names into lowercase, so if the only change to be made is to return variant instead of VARIANT for relevant fields, then this should already be working for the SEA backend, because [SEA returns VARIANT in the type_name](https://github.com/databricks/databricks-jdbc/blob/acad168dddd1350862f6873186583f55c5b28606/src/test/resources/sqlexecapi/datatypesintegrationtests/testvarianttypes/mappings/api_2.0_sql_statements-bf32f446-2dd6-4d2f-9d76-54a1e2dda706.json#L15).

got it, so this is present via result manifest?

jayantsing-db (approved):

src/databricks/sql/backend/thrift_backend.py
Varun0157: > schema bytes are returned inline (embedded) to arrow cloud-fetch files i think
> 
> additionally, no variant type info in result manifest for SEA?

Yes, the schema bytes are embedded in the arrow cloud fetch files but the above changes seem to be solely affecting the `cursor.description`, which only depends on the manifest. 

The `VARIANT` type is returned as a part of the `type_name`, perhaps more info may be available in the `type_code`, but I'm not sure.

Varun0157: > > The SEA backend already converts type_names into lowercase, so if the only change to be made is to return variant instead of VARIANT for relevant fields, then this should already be working for the SEA backend, because [SEA returns VARIANT in the type_name](https://github.com/databricks/databricks-jdbc/blob/acad168dddd1350862f6873186583f55c5b28606/src/test/resources/sqlexecapi/datatypesintegrationtests/testvarianttypes/mappings/api_2.0_sql_statements-bf32f446-2dd6-4d2f-9d76-54a1e2dda706.json#L15).
> 
> got it, so this is present via result manifest?

Yes, exactly

Varun0157: > schema bytes are returned inline (embedded) to arrow cloud-fetch files i think
> 
> additionally, no variant type info in result manifest for SEA?

Yes, schema bytes will be embedded in the arrow cloud fetch files. However, the above changes seem to be targeting setting the type as `variant` in `cursor.description` - this should be handled by the current state of the code since SEA seems to return the `type_name` as `VARIANT` in the manifest itself as shown in the above link.

Varun0157: > > The SEA backend already converts type_names into lowercase, so if the only change to be made is to return variant instead of VARIANT for relevant fields, then this should already be working for the SEA backend, because [SEA returns VARIANT in the type_name](https://github.com/databricks/databricks-jdbc/blob/acad168dddd1350862f6873186583f55c5b28606/src/test/resources/sqlexecapi/datatypesintegrationtests/testvarianttypes/mappings/api_2.0_sql_statements-bf32f446-2dd6-4d2f-9d76-54a1e2dda706.json#L15).
> 
> got it, so this is present via result manifest?

Yes, that's right

PR #566, opened 2025-05-22
[PECOBLR-441] Added classes required for telemetry

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Added classes required for telemetry
## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [x] N/A

PECOBLR-441


src/databricks/sql/telemetry/DriverConnectionParameters.py
```diff
@@ -0,0 +1,24 @@
...
+class DriverConnectionParameters:
+    http_path: str
+    driver_mode: DatabricksClientType
+    host_details: HostDetails
+    auth_mech: AuthMech
+    auth_flow: AuthFlow
+    auth_scope: str
+    discovery_url: str
+    allowed_volume_ingestion_paths: str
+    enable_complex_datatype_support: bool
```
shivam2680, line 19: pls refer to Connection class in client.py file for valid params.
https://github.com/databricks/databricks-sql-python/blob/edfb283f932312e005d3749be30163c0e9982c73/src/databricks/sql/client.py#L70

src/databricks/sql/client.py
```diff
@@ -214,6 +213,12 @@ def read(self) -> Optional[OAuthToken]:
         # use_cloud_fetch
         # Enable use of cloud fetch to extract large query results in parallel via cloud storage
 
+        logger.debug(
+            "Connection.__init__(server_hostname=%s, http_path=%s)",
+            server_hostname,
+            http_path,
+        )
```
vikrantpuppala, line 220: i don't think this should be a part of this PR?

src/databricks/sql/telemetry/DriverSystemConfiguration.py
```diff
@@ -0,0 +1,37 @@
...
+        self.os_name = platform.system()
+        self.os_version = platform.version()
+        self.os_arch = platform.machine()
+        self.runtime_name = platform.python_implementation()
+        self.runtime_version = platform.python_version()
+        self.runtime_vendor = sys.implementation.name
+        self.client_app_name = "databricks-sql-python"
+        self.locale_name = locale.getdefaultlocale()[0]
+        self.driver_name = "databricks-sql-python"
+        self.char_set_encoding = "UTF-8"
```
vikrantpuppala, line 34: can you add an example of what these properties would look like as comments to this class?

vikrantpuppala: Infact let's add class comments across all classes and hopefully an example for each

src/databricks/sql/client.py
saishreeeee: yes my bad that was part of PECOBLR-86

src/databricks/sql/telemetry/DriverSystemConfiguration.py
saishreeeee: ok I will do that

PR #571, opened 2025-05-26
Separate Session related functionality from Connection class

varun-edachali-dbx:
## What type of PR is this?
- [x] Refactor

## Description
Separate the Session related functionality from the Connection class in order to further abstract the backend implementation details from the connection, to eventually give way to the introduction of SEA. 

## How is this tested?
- [x] Unit tests
    - some of the existing unit tests were slightly altered to account for the introduction of the Session class. No unit tests were removed or introduced. 
- [x] E2E Tests
- [ ] Manually
- [ ] N/A

https://databricks.atlassian.net/browse/PECOBLR-438?atlOrigin=eyJpIjoiYjBiMjlhYjBjM2RhNDdiOGE4OGU4NjgwYzU1ZmIzNzciLCJwIjoiaiJ9

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -215,6 +215,12 @@ class Connection:
         # use_cloud_fetch
         # Enable use of cloud fetch to extract large query results in parallel via cloud storage
 
+        logger.debug(
+            "Connection.__init__(server_hostname=%s, http_path=%s)",
+            server_hostname,
+            http_path,
+        )
+
         if access_token:
             access_token_kv = {"access_token": access_token}
             kwargs = {**kwargs, **access_token_kv}
@@ -744,6 +750,9 @@ class Cursor:
 
         :returns self
         """
+        logger.debug(
+            "Cursor.execute(operation=%s, parameters=%s)", operation, parameters
+        )
 
         param_approach = self._determine_parameter_approach(parameters)
         if param_approach == ParameterApproach.NONE:
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -313,6 +313,11 @@ class Connection:
         """Delegate to Session class static method"""
         return Session.get_protocol_version(openSessionResp)
 
+    @property
+    def open(self) -> bool:
+        """Return whether the connection is open by checking if the session is open."""
+        return self.session.open
+
     def cursor(
         self,
         arraysize: int = DEFAULT_ARRAY_SIZE,
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -235,10 +235,13 @@ class Connection:
             catalog,
             schema,
             _use_arrow_native_complex_types,
-            **kwargs
+            **kwargs,
         )
-        
-        logger.info("Successfully opened connection with session " + str(self.get_session_id_hex()))
+
+        logger.info(
+            "Successfully opened connection with session "
+            + str(self.get_session_id_hex())
+        )
 
         self.use_inline_params = self._set_use_inline_params_with_warning(
             kwargs.get("use_inline_params", False)
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -277,7 +277,13 @@ class Connection:
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.close()
+        try:
+            self.close()
+        except BaseException as e:
+            logger.warning(f"Exception during connection close in __exit__: {e}")
+            if exc_type is None:
+                raise
+        return False
 
     def __del__(self):
         if self.open:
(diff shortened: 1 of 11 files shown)


src/databricks/sql/session.py
```diff
@@ -0,0 +1,160 @@
...
+from databricks.sql import USER_AGENT_NAME
+from databricks.sql.thrift_backend import ThriftBackend
+
+logger = logging.getLogger(__name__)
+
+
+class Session:
+    def __init__(
+        self,
+        server_hostname: str,
```
jayantsing-db, line 18: should this be renamed to databricks_hostname, workspace_hostname?

src/databricks/sql/session.py
```diff
@@ -0,0 +1,160 @@
...
+        http_path: str,
+        http_headers: Optional[List[Tuple[str, str]]] = None,
+        session_configuration: Optional[Dict[str, Any]] = None,
+        catalog: Optional[str] = None,
+        schema: Optional[str] = None,
+        _use_arrow_native_complex_types: Optional[bool] = True,
+        **kwargs,
```
jayantsing-db, line 25: nit
- these are identical to connection params. are all of them relevant to a DBSQL session?
- should be introduce a type/namedtuple like `ConnectionParams` or a better name. The reason is that now for each added connection param, we would have to modify at two places

src/databricks/sql/session.py
```diff
@@ -0,0 +1,160 @@
...
+        self.protocol_version = self.get_protocol_version(self._open_session_resp)
+        self.is_open = True
+        logger.info("Successfully opened session " + str(self.get_id_hex()))
+
+    @staticmethod
+    def get_protocol_version(openSessionResp):
+        """
+        Since the sessionHandle will sometimes have a serverProtocolVersion, it takes
+        precedence over the serverProtocolVersion defined in the OpenSessionResponse.
+        """
```
jayantsing-db, line 102:
```suggestion
        precedence over the serverProtocolVersion defined in the OpenSessionResponse.
        
        """
```
i think there is a line gap after a multi-line pydoc. @jprakash-db do we follow any python coding guidelines like for docstring: https://peps.python.org/pep-0257/

jayantsing-db: in databricks we follow this https://databricks.atlassian.net/wiki/spaces/UN/pages/3334538555/Python+Guidelines+go+py but this could be different for OSS repo.

src/databricks/sql/client.py
```diff
@@ -342,34 +305,32 @@ def __del__(self):
...
+
+    @staticmethod
+    def get_protocol_version(openSessionResp):
+        """Delegate to Session class static method"""
+        return Session.get_protocol_version(openSessionResp)
+
+    @property
+    def open(self) -> bool:
+        """Return whether the connection is open by checking if the session is open."""
+        return self.session.is_open
```
jayantsing-db, line 333: nit: should we group together property and staticmethod? @jprakash-db any coding/lint guidelines OSS python driver follows?

src/databricks/sql/session.py
```diff
@@ -0,0 +1,160 @@
...
+                    f"Attempted to close session that was already closed: {e}"
+                )
+            else:
+                logger.warning(
+                    f"Attempt to close session raised an exception at the server: {e}"
+                )
+        except Exception as e:
+            logger.error(f"Attempt to close session raised a local exception: {e}")
+
+        self.is_open = False
```
jayantsing-db, line 160: when opening session the flag is set at the end which makes sense. for closing session call, should we be eager to unset the flag in the very beginning?

jayantsing-db (changes requested):
Hi Varun, thanks for these changes. I have added some minor comments. Please take a look.

src/databricks/sql/session.py
varun-edachali-dbx: I agree, that makes sense. 

Shouldn't we name it `SessionParams` though? These parameters will be passed through to the session constructor.

src/databricks/sql/session.py
jayantsing-db: if the client code has called on the Session class to close the session, then the client assumes that method will close the session. I think unsetting the flag right away makes more sense then. However, an interesting question is do we use this flag internally in Session class to make unsetting meaningful (i.e., when flag is false, do we give null or throw exception when getting session handle?)

src/databricks/sql/client.py
jprakash-db: There are no specific standard in python, because saying something as private etc has no meaning, we can access anything anytime. There are some general standards but nothing concrete

src/databricks/sql/session.py
varun-edachali-dbx: https://databricks.atlassian.net/browse/PECOBLR-447?atlOrigin=eyJpIjoiNDM5OWZkZjlhOGRjNGU5MWI0NGU3NTNiMjIwMTU5ZjIiLCJwIjoiaiJ9

src/databricks/sql/session.py
varun-edachali-dbx: https://databricks.atlassian.net/browse/PECOBLR-448

jayantsing-db (approved):
thanks!

PR #574, opened 2025-05-26
Implement ResultSet Abstraction (backend interfaces for fetch phase)

varun-edachali-dbx:
## What type of PR is this?
- [x] Refactor

## Description
- Create ResultSet interface and make the existing ResultSet (as ThriftResultSet) implement it. Concrete Thrift Backend layer should return ThriftResultSet from execute commands. 
- Generalise the return types of some functions in the backend interface to prevent alignment with Thrift (create new Status enum to avoid using Thrift's Status type). 

## How is this tested?
- [x] Unit tests
    - some of the existing unit tests were slightly altered to account for the introduction of the interface. No unit tests were removed or introduced. 
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

https://databricks.atlassian.net/browse/PECOBLR-445?atlOrigin=eyJpIjoiMjA3ZTljYTUzNDgxNDY4MmFhMThjYjI5MWY0Yjc0MTIiLCJwIjoiaiJ9


src/databricks/sql/backend/databricks_client.py
```diff
@@ -130,7 +136,7 @@ def cancel_command(self, command_id: CommandId) -> None:
         pass
 
     @abstractmethod
-    def close_command(self, command_id: CommandId) -> ttypes.TStatus:
+    def close_command(self, command_id: CommandId) -> None:
```
jayantsing-db, line 139: why changed to void/none? if you choose to return void, make sure the implementations throw appropriate errors for the calling code

src/databricks/sql/backend/databricks_client.py
```diff
@@ -175,7 +178,7 @@ def get_execution_result(
         self,
         command_id: CommandId,
         cursor: "Cursor",
-    ) -> ExecuteResponse:
+    ) -> "ResultSet":
```
jayantsing-db, line 181: please update/re-review the pydoc(s) **everywhere** since we are changing signatures.

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -12,6 +12,7 @@
 
 from databricks.sql.thrift_api.TCLIService.ttypes import TOperationState
 from databricks.sql.backend.types import (
+    CommandState,
     SessionId,
     CommandId,
     BackendType,
```
jayantsing-db, line 19: nit: alphabetic ordered imports; do we have that setup in the linter?

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -808,7 +810,7 @@ def _results_message_to_execute_response(self, resp, operation_state):
 
     def get_execution_result(
         self, command_id: CommandId, cursor: "Cursor"
-    ) -> ExecuteResponse:
+    ) -> "ResultSet":
```
jayantsing-db, line 814: for a follow-up PR to this, since the actual implementations of databricks client interface might be different for SEA and thrift, consider adding specific pydoc to different implementations

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -869,6 +871,15 @@ def get_execution_result(
             arrow_schema_bytes=schema_bytes,
         )
 
+        return ThriftResultSet(
+            connection=cursor.connection,
+            execute_response=execute_response,
+            thrift_client=self,
```
jayantsing-db, line 880: please be mindful of introducing circular dependencies or references. thrift_backend outputs thrift_result_set which depends on thrift_backend

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -869,6 +871,15 @@ def get_execution_result(
             arrow_schema_bytes=schema_bytes,
         )
 
+        return ThriftResultSet(
+            connection=cursor.connection,
```
jayantsing-db, line 878: why the need for connection in result set?

src/databricks/sql/backend/types.py
```diff
@@ -1,12 +1,46 @@
...
+from typing import Dict, Optional, Any
 import logging
 
 from databricks.sql.backend.utils import guid_to_hex_id
+from databricks.sql.thrift_api.TCLIService import ttypes
 
 logger = logging.getLogger(__name__)
 
 
+class CommandState(Enum):
```
jayantsing-db, line 11: pydoc for this and other types.

src/databricks/sql/backend/types.py
```diff
@@ -1,12 +1,46 @@
...
+            ttypes.TOperationState.TIMEDOUT_STATE,
+            ttypes.TOperationState.UKNOWN_STATE,
+        ):
+            return cls.FAILED
+        elif state == ttypes.TOperationState.CLOSED_STATE:
+            return cls.CLOSED
+        elif state == ttypes.TOperationState.CANCELED_STATE:
+            return cls.CANCELLED
+        else:
+            raise ValueError(f"Unknown command state: {state}")
```
jayantsing-db: where is this categorisation coming from?

src/databricks/sql/client.py
```diff
@@ -42,14 +42,15 @@
...
+from databricks.sql.result_set import ResultSet, ThriftResultSet
 from databricks.sql.types import Row, SSLOptions
 from databricks.sql.auth.auth import get_python_sql_connector_auth_provider
 from databricks.sql.experimental.oauth_persistence import OAuthPersistence
 from databricks.sql.session import Session
-from databricks.sql.backend.types import CommandId, BackendType
+from databricks.sql.backend.types import CommandId, BackendType, CommandState, SessionId
 
 from databricks.sql.thrift_api.TCLIService.ttypes import (
+    TOpenSessionResp,
```
jayantsing-db, line 53: is there a standard way of imports like ordering, line-gaps, etc.? is that covered by the manual linter? @jprakash-db could you please opine?

src/databricks/sql/client.py
```diff
@@ -783,18 +792,8 @@ def execute(
...
-            self.buffer_size_bytes,
-            self.arraysize,
-            self.connection.use_cloud_fetch,
-        )
 
-        if execute_response.is_staging_operation:
+        if self.active_result_set and self.active_result_set.is_staging_operation:
             self._handle_staging_operation(
                 staging_allowed_local_path=self.connection.staging_allowed_local_path
             )
```
jayantsing-db, line 801: @jprakash-db, is this uc volume ingestion? Apparently, in the result set, we get the pre-signed URL as the first row.

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+class ResultSet(ABC):
+    """
+    Abstract base class for result sets returned by different backend implementations.
+
+    This class defines the interface that all concrete result set implementations must follow.
+    """
+
+    def __init__(self, connection, backend, arraysize: int, buffer_size_bytes: int):
```
jayantsing-db: please have a detailed pydoc similar to the deleted Result set class. Same for other methods and APIs.

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,363 @@
...
+        """Initialize the base ResultSet with common properties."""
+        self.connection = connection
+        self.backend = backend  # Store the backend client directly
```
jayantsing-db: this seems redundant as connection already has session which has backend?

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,363 @@
...
+
+    @abstractmethod
+    def fetchmany_arrow(self, size: int) -> Any:
+        """Fetch the next set of rows as an Arrow table."""
+        pass
+
+    @abstractmethod
+    def fetchall_arrow(self) -> Any:
+        """Fetch all remaining rows as an Arrow table."""
+        pass
```
jayantsing-db: please make sure that this covers all fetch variants

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,363 @@
...
+        self._use_cloud_fetch = use_cloud_fetch
+        self._is_staging_operation = execute_response.is_staging_operation
+
+        # Initialize results queue
+        if execute_response.arrow_queue:
+            self.results = execute_response.arrow_queue
+        else:
+            self._fill_results_buffer()
+
+    def _fill_results_buffer(self):
```
jayantsing-db, line 186: please retain any comments or pydoc of the previous authors

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def merge_columnar(self, result1, result2):
+        """
+        Function to merge / combining the columnar results into a single result
+        :param result1:
+        :param result2:
+        :return:
+        """
+
+        if result1.column_names != result2.column_names:
```
jayantsing-db, line 259: nit: please be consistent with the line gaps and pydoc styles throughout the changes if there is no standard for the repo yet.

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,363 @@
...
+        df = table_renamed.to_pandas(
+            types_mapper=dtype_mapping.get,
+            date_as_object=True,
+            timestamp_as_object=True,
+        )
+
+        res = df.to_numpy(na_value=None, dtype="object")
+        return [ResultRow(*v) for v in res]
+
+    def merge_columnar(self, result1, result2):
```
jayantsing-db: no return typing? if present in original code, please retain. Same for other implementations.

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,363 @@
...
+            return self._convert_columnar_table(self.fetchmany_columnar(size))
+        else:
+            return self._convert_arrow_table(self.fetchmany_arrow(size))
+
+    def close(self) -> None:
+        """
+        Close the cursor.
+
+        If the connection has not been closed, and the cursor has not already
+        been closed on the server for some other reason, issue a request to the server to close it.
```
jayantsing-db: let's reword to result set.

tests/unit/test_client.py
```diff
@@ -91,16 +91,19 @@ def test_closing_connection_closes_commands(self, mock_result_set_class):
         # Test once with has_been_closed_server side, once without
         for closed in (True, False):
             with self.subTest(closed=closed):
-                mock_result_set_class.return_value = Mock()
                 connection = databricks.sql.connect(**self.DUMMY_CONNECTION_ARGS)
                 cursor = connection.cursor()
-                cursor.execute("SELECT 1;")
+
+                # Create a mock result set and set it as the active result set
+                mock_result_set = Mock()
```
jayantsing-db: will line 89 and test method param be redundant with line 98?

tests/unit/test_client.py
```diff
@@ -91,16 +91,19 @@ def test_closing_connection_closes_commands(self, mock_result_set_class):
...
+                mock_result_set.has_been_closed_server_side = closed
+                cursor.active_result_set = mock_result_set
```
jayantsing-db: looks like you are manually setting the property in mocked result set but the previous test didn't have to do that. this unit test now has reduced fidelity.

is it because the previous test executed the query SELECT 1?

tests/unit/test_client.py
```diff
@@ -162,17 +166,21 @@ def test_closing_result_set_hard_closes_commands(self):
...
-    @patch("%s.client.ResultSet" % PACKAGE_NAME)
+    @patch("%s.result_set.ThriftResultSet" % PACKAGE_NAME)
     def test_executing_multiple_commands_uses_the_most_recent_command(
         self, mock_result_set_class
     ):
-
         mock_result_sets = [Mock(), Mock()]
+        # Set is_staging_operation to False to avoid _handle_staging_operation being called
+        for mock_rs in mock_result_sets:
+            mock_rs.is_staging_operation = False
```
jayantsing-db, line 227: why is this needed now?

tests/unit/test_client.py
```diff
@@ -349,14 +357,19 @@ def test_execute_parameter_passthrough(self):
...
+    @patch("%s.result_set.ThriftResultSet" % PACKAGE_NAME)
     def test_executemany_parameter_passhthrough_and_uses_last_result_set(
         self, mock_result_set_class
     ):
         # Create a new mock result set each time the class is instantiated
         mock_result_set_instances = [Mock(), Mock(), Mock()]
+        # Set is_staging_operation to False to avoid _handle_staging_operation being called
+        for mock_rs in mock_result_set_instances:
+            mock_rs.is_staging_operation = False
+
```
jayantsing-db, line 415: why is this change needed in tests?

tests/unit/test_client.py
```diff
@@ -162,17 +166,21 @@ def test_closing_result_set_hard_closes_commands(self):
...
+        for mock_rs in mock_result_sets:
+            mock_rs.is_staging_operation = False
+
         mock_result_set_class.side_effect = mock_result_sets
 
-        cursor = client.Cursor(
-            connection=Mock(), backend=ThriftDatabricksClientMockFactory.new()
-        )
+        mock_backend = ThriftDatabricksClientMockFactory.new()
+        mock_backend.execute_command.side_effect = mock_result_sets
```
jayantsing-db, line 230: if we already have `mock_result_set_class.side_effect = mock_result_sets`, why is there a need `mock_backend.execute_command.side_effect = mock_result_sets`. 

the first one should ensure that the two mocked results sets are returned on respective calls to execute command?

tests/unit/test_client.py
```diff
@@ -349,14 +357,19 @@ def test_execute_parameter_passthrough(self):
...
     ):
         # Create a new mock result set each time the class is instantiated
         mock_result_set_instances = [Mock(), Mock(), Mock()]
+        # Set is_staging_operation to False to avoid _handle_staging_operation being called
+        for mock_rs in mock_result_set_instances:
+            mock_rs.is_staging_operation = False
+
         mock_result_set_class.side_effect = mock_result_set_instances
         mock_backend = ThriftDatabricksClientMockFactory.new()
+        mock_backend.execute_command.side_effect = mock_result_set_instances
```
jayantsing-db, line 417: if we already have mock_result_set_class.side_effect = mock_result_sets, why is there a need mock_backend.execute_command.side_effect = mock_result_sets.

the first one should ensure that the two mocked results sets are returned on respective calls to execute command?

tests/unit/test_thrift_backend.py
```diff
@@ -1297,6 +1312,9 @@ def test_get_columns_calls_client_and_handle_execute_response(
             table_name="table_pattern",
             column_name="column_pattern",
         )
+        # Verify the result is a ResultSet
+        self.assertIsInstance(result, ResultSet)
```
jayantsing-db, line 1316: nice. if possible, we should look into further increasing coverage with respect to refracted code in separate PRs

jayantsing-db (changes requested):
mostly looks good. Posted some suggestions and comments.

src/databricks/sql/client.py
jprakash-db: yes

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: The return type is not used by any of the references, so I chose to convert it to a `None` return type. 

Any errors in the call will be raised by `make_request`. This structure is identical to the existing `cancel_command` implementation (which returns `None` as well).

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: In this case it feels essential to me, since the `ThriftResultSet` needs to have access to a `ThriftDatabricksClient` in order to implement the fetch phase. 

The only way I can think of to avoid this anti-pattern is to create separate fetch phase and exec phase clients, but that would be a large deviation from the existing connectors. 

Any other ideas?

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: In the current implementation, it is used for: 
- checking if the connection is open (we only call `backend.close_command` if the connection is not already closed)
- to access a property `disable_pandas` currently stored in the connection instance

Passing the property is not an essential use-case, because we can just pass it as a parameter to the ResultSet if required, but I don't think we can circumvent the first use case.

src/databricks/sql/backend/types.py
varun-edachali-dbx: It comes from [the JDBC implementation](https://github.com/databricks/databricks-jdbc/blob/30688befd95060874b2dd00019dc85a449ed435a/src/main/java/com/databricks/jdbc/common/util/DatabricksThriftUtil.java#L85).

src/databricks/sql/result_set.py
varun-edachali-dbx: Yes, not sure how I missed this. I'll add them back.

src/databricks/sql/result_set.py
varun-edachali-dbx: This is mostly helpful to store the specialised backend in a unified way. We can call the parent class's constructor (i.e. this constructor) and store `thrift_client` (for example) as `backend`. 

Once we do this we can abstract away some common functionality across the backend clients. For example, the `close` method can likely be pulled up to the parent since it only depends on methods defined in the common interface.

varun-edachali-dbx: I will pull up the `close` method from the specialised implementation to abide by this.

src/databricks/sql/result_set.py
varun-edachali-dbx: There is [no return typing](https://github.com/databricks/databricks-sql-python/blob/0947b9aa194d37198fa0b05147f50e4682a5643f/src/databricks/sql/client.py#L1404) in the original code. I will add it here, though.

tests/unit/test_client.py
varun-edachali-dbx: Yes, you're right. I'll remove it.

tests/unit/test_client.py
varun-edachali-dbx: Earlier, we were patching the `ResultSet` class, which no longer works because internally a `ThriftResultSet` is created, so I'm creating a `mock_result_set` instance now and assigning it as the active result set. 

Earlier, the test was not using the `closed` attribute defined for the creation of the subTests. I added an assertion ensuring that the `mock_result_set` is closed regardless of whether or not the command has been closed on the server side, and that this state is preserved.

tests/unit/test_client.py
varun-edachali-dbx: The second call is required to override the `execute_command` of the `mock_backend` as returned by the factory. In the factory, it returns only a single `mock_result_set`.

You are right that this makes the first call redundant though, because this is a higher level mock.  I will remove the first call.

tests/unit/test_client.py
varun-edachali-dbx: Discussed [here](https://github.com/databricks/databricks-sql-python/pull/574/files?w=1#r2120620943).

tests/unit/test_client.py
varun-edachali-dbx: I will remove the first call, as discussed [here](https://github.com/databricks/databricks-sql-python/pull/574/files?w=1#r2120657642).

tests/unit/test_client.py
varun-edachali-dbx: Earlier: this value was extracted from the `ExecuteResponse` type, not from the `ResultSet` type. 

Now: the `ExecuteResponse` type is no longer returned by the backend. `is_staging_operation` is checked as an attribute on the returned `ResultSet`. In the `ThriftDatabricksClientMockFactory` the value of this attribute is set to `False`, but in the tests where we create our own mock result sets, we have to set this attribute explicitly. If we do not do this, the value extracted from the mock defaults to true and we treat the operation as a staging operation, which we do not want here.

tests/unit/test_client.py
varun-edachali-dbx: Discussed [here](https://github.com/databricks/databricks-sql-python/pull/574/files?w=1#r2122434778).

src/databricks/sql/result_set.py
varun-edachali-dbx: Sure, I will introduce a line gap after each multi-line pydoc for consistency. I will focus on the files that I have altered and ignore the sub-dirs `sql/parameters`, `sql/cloudfetch`, and `sql/experimental` for now, because the application of these rules here is not consistent. I can make a separate PR for linting across the codebase later if required.

src/databricks/sql/backend/thrift_backend.py
jayantsing-db: we already pass connection which has session which in turn has the backend? can we look into normalising this init params (in a separate PR)

jayantsing-db (approved):
thanks. a minor comment for a separate PR.

Issue #578, opened 2025-05-29
Allow passing parameters into `to_pandas` call


j-bennet:
When creating a `Connection`, currently it's possible to configure some aspects of pandas / arrow, such as `_use_arrow_native_complex_types` and `_use_arrow_native_timestamps`, but `_convert_arrow_table` uses a hardcoded  type mapper that's impossible to change:

https://github.com/databricks/databricks-sql-python/blob/0947b9aa194d37198fa0b05147f50e4682a5643f/src/databricks/sql/client.py#L1349-L1362

Similarly, there's no way to pass other parameters to the `to_pandas` call here:

https://github.com/databricks/databricks-sql-python/blob/main/src/databricks/sql/client.py#L1366-L1370

Please consider extending the connection parameters to add more flexibility.

PR #579, opened 2025-05-30
Enhance Arrow to Pandas conversion with type overrides and additional kwargs [#578]

madhav-db:
* Introduced _arrow_pandas_type_override and _arrow_to_pandas_kwargs in Connection class for customizable dtype mapping and DataFrame construction parameters.
* Updated ResultSet to utilize these new options during conversion from Arrow tables to Pandas DataFrames.
* Added unit tests to validate the new functionality, including scenarios for type overrides and additional kwargs handling.

## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -213,6 +213,11 @@ class Connection:
         # (True by default)
         # use_cloud_fetch
         # Enable use of cloud fetch to extract large query results in parallel via cloud storage
+        # _arrow_pandas_type_override
+        # Override the default pandas dtype mapping for Arrow types.
+        # This is a dictionary of Arrow types to pandas dtypes.
+        # _arrow_to_pandas_kwargs
+        # Additional or modified arguments to pass to pandas.DataFrame constructor.
 
         logger.debug(
             "Connection.__init__(server_hostname=%s, http_path=%s)",
@@ -1346,7 +1351,7 @@ class ResultSet:
         # Need to use nullable types, as otherwise type can change when there are missing values.
         # See https://arrow.apache.org/docs/python/pandas.html#nullable-types
         # NOTE: This api is epxerimental https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html
-        dtype_mapping = {
+        DEFAULT_DTYPE_MAPPING: Dict[pyarrow.DataType, pandas.api.extensions.ExtensionDtype] = {
             pyarrow.int8(): pandas.Int8Dtype(),
             pyarrow.int16(): pandas.Int16Dtype(),
             pyarrow.int32(): pandas.Int32Dtype(),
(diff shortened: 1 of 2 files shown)


src/databricks/sql/client.py
```diff
@@ -1361,13 +1370,35 @@ def _convert_arrow_table(self, table):
             pyarrow.string(): pandas.StringDtype(),
         }
 
+        arrow_pandas_type_override = self.connection._arrow_pandas_type_override
+        if not isinstance(arrow_pandas_type_override, dict):
```
jprakash-db, line 1374: This if block is not needed, let it fail here itself. Don't want the user to give something incorrect and then everything works. This is a new change and nothing to be backward compatible.
just leave it at this - `arrow_pandas_type_override = self.connection._arrow_pandas_type_override`

src/databricks/sql/client.py
```diff
@@ -1361,13 +1370,35 @@ def _convert_arrow_table(self, table):
...
+        }
+
+        to_pandas_kwargs: dict[str, Any] = {
+            "types_mapper": dtype_mapping.get,
+            "date_as_object": True,
+            "timestamp_as_object": True,
+        }
+
+        arrow_to_pandas_kwargs = self.connection._arrow_to_pandas_kwargs
+        if isinstance(arrow_to_pandas_kwargs, dict):
```
jprakash-db, line 1392: Same let it fail when the input format is incorrect. The default python interpreter error of type mismatch is enough

tests/unit/test_arrow_conversion.py
```diff
@@ -0,0 +1,184 @@
...
+        mock_connection._arrow_to_pandas_kwargs = {}
+        rs_ts_default = ResultSet(
+            mock_connection, mock_raw_execute_response, mock_thrift_backend
+        )
+        result_default = rs_ts_default.fetchall()
+        self.assertEqual(len(result_default), 1)
+        self.assertIsInstance(result_default[0].col_ts, datetime.datetime)
+
+
+if __name__ == "__main__":
```
jprakash-db, line 183: this is not needed, as we run tests using pytest. Also can you move everything to pytest and remove unittest

tests/unit/test_arrow_conversion.py
```diff
@@ -0,0 +1,184 @@
...
+import unittest
+from unittest.mock import MagicMock
+
+from databricks.sql.client import ResultSet, Connection, ExecuteResponse
+from databricks.sql.types import Row
+from databricks.sql.utils import ArrowQueue
+
+@pytest.mark.skipif(pa is None, reason="PyArrow is not installed")
+class ArrowConversionTests(unittest.TestCase):
+    @staticmethod
```
jprakash-db, line 18: Move this to use fixtures

tests/unit/test_arrow_conversion.py
```diff
@@ -0,0 +1,184 @@
...
+    def mock_connection_static():
+        conn = MagicMock(spec=Connection)
+        conn.disable_pandas = False
+        conn._arrow_pandas_type_override = {}
+        conn._arrow_to_pandas_kwargs = {}
+        return conn
+
+    @staticmethod
```
jprakash-db, line 26: Use fixtures or just normal functions, don't need static methods

tests/unit/test_arrow_conversion.py
```diff
@@ -0,0 +1,184 @@
...
+            sample_arrow_table, sample_arrow_table.num_rows
+        )
+        rs = ResultSet(mock_connection, mock_raw_execute_response, mock_thrift_backend)
+        result = rs.fetchall()
+        self.assertEqual(len(result), 3)
+        self.assertIsInstance(result[0].col_int, float)
+        self.assertEqual(result[0].col_int, 1.0)
+        self.assertEqual(result[0].col_str, "a")
+
+    def test_convert_arrow_table_to_pandas_kwargs(self):
```
jprakash-db, line 130: Too much code duplication. Can you create a parameterized test, where in this `mock_connection._arrow_to_pandas_kwargs = {"timestamp_as_object": False}` and the assertIsInstance values are parameterized. Otherwise the checking part looks the same and is copied repeatedly

tests/unit/test_arrow_conversion.py
```diff
@@ -0,0 +1,184 @@
...
+        er.arrow_queue = None
+        er.has_more_rows = False
+        er.lz4_compressed = False
+        er.command_handle = MagicMock()
+        er.status = MagicMock()
+        er.has_been_closed_server_side = False
+        er.is_staging_operation = False
+        return er
+
+    def test_convert_arrow_table_default(self):
```
jprakash-db, line 61: The test_convert_arrow_table_deafult, test_convert_arrow_table_disable_pandas and test_convert_arrow_table_type_override are essentially the same test flow just with different arguments. Plz use pytest's parameterized tests for such tests where only arguments change

PR #581, opened 2025-06-03
E2E POC for python telemetry for connect + latency + error logs

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

PECOBLR-457


src/databricks/sql/client.py
```diff
@@ -294,6 +302,25 @@ def read(self) -> Optional[OAuthToken]:
...
+            )
+
+            telemetry_client.export_initial_telemetry_log(
+                http_path,
+                self.port,
+                kwargs.get("_socket_timeout", None),
+                self.get_session_id_hex(),
+            )
+        else:
+            self.telemetry_client = NoopTelemetryClient()
```
vikrantpuppala: should we just pass telemetry_enabled to the initialize call and just use the Noop client there? that way the logic of which client is being used is abstracted from the rest of the codebase

src/databricks/sql/client.py
```diff
@@ -294,6 +302,25 @@ def read(self) -> Optional[OAuthToken]:
             kwargs.get("use_inline_params", False)
         )
 
+        if self.telemetry_enabled:
+            telemetry_client.initialize(
+                host=self.host,
+                connection_uuid=self.get_session_id_hex(),
```
vikrantpuppala: why is connection_uuid = session_id_hex?

src/databricks/sql/client.py
```diff
@@ -487,7 +517,10 @@ def __iter__(self):
             for row in self.active_result_set:
                 yield row
         else:
-            raise Error("There is no active result set")
+            raise Error(
+                "There is no active result set",
+                connection_uuid=self.connection.get_session_id_hex(),
```
vikrantpuppala: why pass this here? i don't think the connection_uuid is being used?

src/databricks/sql/client.py
```diff
@@ -713,7 +753,13 @@ def _handle_staging_put(
...
 
         if local_file is None:
-            raise Error("Cannot perform PUT without specifying a local_file")
+            raise Error(
+                "Cannot perform PUT without specifying a local_file",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
+
+        self.volume_operation_type = DriverVolumeOperationType.PUT
+        self.volume_path = local_file
```
vikrantpuppala: why do we need to set these? are there any side effects to doing this?

src/databricks/sql/client.py
```diff
@@ -748,25 +796,38 @@ def _handle_staging_get(
...
+    @log_latency()
     def _handle_staging_remove(
         self, presigned_url: str, headers: Optional[dict] = None
     ):
         """Make an HTTP DELETE request to the presigned_url"""
 
+        self.volume_operation_type = DriverVolumeOperationType.DELETE
+        self.volume_path = (
+            presigned_url  # Using presigned URL as path since there's no local file
+        )
```
vikrantpuppala: please don't log presigned urls, keep this empty if not required

src/databricks/sql/client.py
```diff
@@ -748,25 +796,38 @@ def _handle_staging_get(
...
 
         if local_file is None:
-            raise Error("Cannot perform GET without specifying a local_file")
+            raise Error(
+                "Cannot perform GET without specifying a local_file",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
+
+        self.volume_operation_type = DriverVolumeOperationType.GET
+        self.volume_path = local_file
```
vikrantpuppala: i don't think we should log local_file directory, why are we doing this?

src/databricks/sql/exc.py
```diff
@@ -11,10 +12,22 @@ class Error(Exception):
...
+        self.connection_uuid = connection_uuid
+
+        try:
+            error_name = self.__class__.__name__
+            error_message = self.message_with_context()
+            telemetry_client.export_failure_log(
+                error_name, error_message, self.connection_uuid
+            )
+        except Exception as e:
+            logger.warning(f"Failed to log error to telemetry: {str(e)}")
```
vikrantpuppala: users of telemetry client should not be responsible for error handling, it should be upto telemetry client to handle errors without throwing back to user

src/databricks/sql/client.py
saishreeeee: session id is specific to each connection, that way we make a separate telemetry client per connection

src/databricks/sql/client.py
saishreeeee: For error logging, I added connection_uuid in the args of error init function, so that telemetry manager knows which telemetry client to send the error logs to

src/databricks/sql/client.py
saishreeeee: this was for logging latency for volume operations but I will shift this to latency_logger as it makes sense for this to be set there

src/databricks/sql/client.py
saishreeeee: yeah I will correct this, I have to set uc volume path not the local path.

src/databricks/sql/client.py
```diff
@@ -294,6 +298,23 @@ def read(self) -> Optional[OAuthToken]:
             kwargs.get("use_inline_params", False)
         )
 
+        if self.telemetry_enabled:
```
vikrantpuppala: instead of checking everywhere if telemetry_enabled == true, I think it is better to delegate this responsibility to the telemetry client i.e. just pass this flag to the client and then it should do whatever it needs to in that case, that way, we also do not need to depend on if the telemetry_client exists to know if telemetry is enabled?

vikrantpuppala: think of it in these terms: the application using the telemetry library should not need to understand the internals.. it should know the APIs and only focus on params to call it. the library should do what it needs to do based on the API params

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,272 @@
...
+        # for key, value in response.headers.items():
+        #     print(f"  {key}: {value}", flush=True)
+
+        # print("\nResponse Body:", flush=True)
+        # try:
+        #     response_json = response.json()
+        #     print(json.dumps(response_json, indent=2), flush=True)
+        # except json.JSONDecodeError:
+        #     print(response.text, flush=True)
+        # sys.stdout.flush()
```
vikrantpuppala: remove commented code

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,272 @@
...
+        self.connection_uuid = connection_uuid
+        self.auth_provider = auth_provider
+        self.is_authenticated = is_authenticated
+        self.batch_size = batch_size
+        self.user_agent = user_agent
+        self.events_batch = []
+        self.lock = threading.Lock()
+        self.executor = ThreadPoolExecutor(
+            max_workers=10  # TODO: Decide on max workers
+        )  # Thread pool for async operations
```
vikrantpuppala: we would create a new thread pool for each client? that's not very good

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,272 @@
...
+        # except json.JSONDecodeError:
+        #     print(response.text, flush=True)
+        # sys.stdout.flush()
+
+    def close(self):
+        """Flush remaining events and shut down executor"""
+        self.flush()
+        self.executor.shutdown(wait=True)
+
+    def export_initial_telemetry_log(self, http_path, port, socket_timeout):
```
vikrantpuppala: instead of just taking socket_timeout as input, i would recommend taking kwargs and then getting whatever you want here, it'd be easier to add more things later on that way

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def export_sql_latency_log(
+        self, latency_ms, sql_execution_event, sql_statement_id=None
+    ):
+        """Export telemetry for sql execution"""
+        pass
+
+    def export_volume_latency_log(self, latency_ms, volume_operation):
+        """Export telemetry for volume operation"""
+        pass
```
vikrantpuppala: can we add these in next PR please

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,272 @@
...
+    def close(self, connection_uuid):
+        """Close telemetry client(s)"""
+        if connection_uuid:
+            if connection_uuid in self._clients:
+                self._clients[connection_uuid].close()
+                del self._clients[connection_uuid]
+
+
+# Create a global instance
+telemetry_client = TelemetryManager()
```
vikrantpuppala: hmm, telemetry_client is actually a TelemetryManager and TelemetryManager is just a wrapper on top of TelemetryClient, not particularly a fan of this design, is there a reason why we chose to do it this way? how is it better than a factory pattern?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Yes, I will shift send_telemetry to TelemetryManger

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: I did not have TelemetryManager initially. During error and latency logging, it was easier to just pass connection_uuid rather than the telemetry client, so I made a singleton class to handle all the telemetry clients. The naming here does look confusing. I will change things up.

src/databricks/sql/client.py
```diff
@@ -294,6 +298,26 @@ def read(self) -> Optional[OAuthToken]:
...
+            telemetry_enabled=self.telemetry_enabled,
+            batch_size=telemetry_batch_size,
+            connection_uuid=self.get_session_id_hex(),
+            **telemetry_kwargs,
+        )
+
+        intial_telmetry_kwargs = {
+            "http_path": http_path,
+            "port": self.port,
+            "socket_timeout": kwargs.get("_socket_timeout", None),
```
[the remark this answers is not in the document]
saishreeeee: In the comment about internal arguments in **kwargs, it is mentioned as _socket_timeout

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,218 @@
...
+
+        if is_authenticated and auth_provider:
+            auth_provider.add_headers(headers)
+
+        self.executor.submit(
+            requests.post, url, data=json.dumps(request), headers=headers, timeout=10
+        )
+
+    @classmethod
+    def getDriverSystemConfiguration(cls) -> DriverSystemConfiguration:
```
vikrantpuppala: let's move util functions to a helper class

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,218 @@
...
+                    telemetry_enabled=telemetry_enabled,
+                    batch_size=batch_size,
+                    connection_uuid=connection_uuid,
+                    **kwargs,
+                )
+            return self._clients[connection_uuid]
+        else:
+            return NoopTelemetryClient()
+
+    def _send_telemetry(self, events, host_url, is_authenticated, auth_provider):
```
vikrantpuppala: send_telemetry should ideally be in telemetry client, let's create a single thread pool and pass that to the telemetry client or do it using the helper class

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,218 @@
...
+
+        self._clients = {}  # Map of connection_uuid -> TelemetryClient
+        self.executor = ThreadPoolExecutor(
+            max_workers=10
+        )  # Thread pool for async operations TODO: Decide on max workers
+        self._initialized = True
+
+    def get_telemetry_client(
+        self, telemetry_enabled, batch_size, connection_uuid, **kwargs
+    ):
```
vikrantpuppala, line 298: the user of this function would not know what to pass in kwargs, so let's try to avoid using tyhat

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,218 @@
...
+
+        if events_to_flush:
+            telemetry_client_factory._send_telemetry(
+                events_to_flush,
+                self.host_url,
+                self.is_authenticated,
+                self.auth_provider,
+            )
+
+    def export_initial_telemetry_log(self, **kwargs):
```
vikrantpuppala: similar comment on kwargs

src/databricks/sql/telemetry/models/event.py
```diff
@@ -9,6 +9,14 @@
...
 )
 from typing import Optional
+from enum import Enum
+
+
+class EnumEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, Enum):
+            return obj.value
+        return super().default(obj)
```
vikrantpuppala: why is this needed? let's add comments to document why

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,218 @@
...
+)
+from databricks.sql.telemetry.models.enums import DatabricksClientType
+import sys
+import platform
+import uuid
+import locale
+from abc import ABC, abstractmethod
+
+
+class BaseTelemetryClient(ABC):
```
vikrantpuppala, line 95: let's add docstring across

examples/persistent_oauth.py
```diff
@@ -60,4 +60,4 @@ def read(self, hostname: str) -> Optional[OAuthToken]:
             print(row)
         cursor.close()
 
-    connection.close()
+    connection.close()
```
vikrantpuppala: let's remove this change

src/databricks/sql/client.py
```diff
@@ -238,6 +247,9 @@ def read(self) -> Optional[OAuthToken]:
         self.telemetry_enabled = (
             self.client_telemetry_enabled and self.server_telemetry_enabled
         )
+        telemetry_batch_size = kwargs.get(
+            "telemetry_batch_size", 100
+        )  # TODO: Decide on batch size
```
vikrantpuppala: default should be define in telemetry_client, we should only pass this value if provided, infact to start with let's not make this configurable at all

src/databricks/sql/client.py
```diff
@@ -294,6 +306,26 @@ def read(self) -> Optional[OAuthToken]:
...
+            discovery_url=TelemetryHelper.get_discovery_url(auth_provider),
+            socket_timeout=kwargs.get("_socket_timeout", None),
+        )
+        self.telemetry_client = telemetry_client_factory.get_telemetry_client(
+            telemetry_enabled=self.telemetry_enabled,
+            batch_size=telemetry_batch_size,
+            connection_uuid=self.get_session_id_hex(),
+            auth_provider=auth_provider,
+            user_agent=useragent_header,
+            driver_connection_params=driver_connection_params,
```
vikrantpuppala, line 327: not sure if we should have driver_connection_params here, we would want to call get_telemetry_client at classes where we would not have all info to create a driver_connection_params, infact, other than basic params like telemetry_enabled, connection_uuid, we should have a very strong reason to have other params while calling this function. essentially, please ensure that we're able to get a telemetry_client anywhere without passing heavy objects

src/databricks/sql/telemetry/models/enums.py
```diff
@@ -3,15 +3,14 @@
...
 
 
 class AuthMech(Enum):
-    OTHER = "other"
-    PAT = "pat"
-    OAUTH = "oauth"
+    OTHER = "OTHER"
+    PAT = "PAT"
+    OAUTH = "OAUTH"
+    EXTERNAL = "EXTERNAL"
```
vikrantpuppala: can you add comments for what external means, or have a more descriptive enum name?

vikrantpuppala: for that matter what does OTHER mean too?

src/databricks/sql/telemetry/models/event.py
```diff
@@ -9,6 +9,20 @@
...
 )
 from typing import Optional
+from enum import Enum
+
+
+class EnumEncoder(json.JSONEncoder):
+    """
+    Custom JSON encoder to handle Enum values.
+    This is used to convert Enum values to their string representations.
+    Default JSON encoder does not handle Enum values.
```
vikrantpuppala: what happens in default json encoder?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,286 @@
...
+                runtime_name=f"Python {sys.version.split()[0]}",
+                runtime_vendor=platform.python_implementation(),
+                runtime_version=platform.python_version(),
+                os_name=platform.system(),
+                os_version=platform.release(),
+                os_arch=platform.machine(),
+                client_app_name=None,  # TODO: Add client app name
+                locale_name=locale.getlocale()[0] or locale.getdefaultlocale()[0],
+                char_set_encoding=sys.getdefaultencoding(),
+            )
```
vikrantpuppala: can we not define this directly at line 32 to not have this re-initialization check?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,286 @@
...
+        """Get the auth flow for the auth provider."""
+        if not auth_provider:
+            return None
+
+        if isinstance(auth_provider, DatabricksOAuthProvider):
+            if (
+                hasattr(auth_provider, "_refresh_token")
+                and auth_provider._refresh_token
+            ):
+                return AuthFlow.TOKEN_PASSTHROUGH
```
vikrantpuppala: why is the existence of refersh_token a signal for TOKEN_PASSTHROUGH?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,286 @@
...
+    @staticmethod
+    def get_discovery_url(auth_provider):
+        """Get the discovery URL for the auth provider."""
+        if not auth_provider:
+            return None
+
+        if isinstance(auth_provider, DatabricksOAuthProvider):
+            return auth_provider.oauth_manager.idp_endpoint.get_openid_config_url(
+                auth_provider.hostname
+            )
```
vikrantpuppala: discovery_url should only be populated when provided directly, not by constructing the OIDC endpoint

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,286 @@
...
+    def _send_telemetry(self, events):
+        """Send telemetry events to the server"""
+        try:
+            request = {
+                "uploadTime": int(time.time() * 1000),
+                "items": [],
+                "protoLogs": [event.to_json() for event in events],
+            }
+        except Exception as e:
+            print(f"[DEBUG] Error creating telemetry request: {e}", flush=True)
```
vikrantpuppala: lets not use print and use log instead, please add log statements across the code you're adding

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,286 @@
...
+    It uses a thread pool to handle asynchronous operations.
+    """
+
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = super(TelemetryClientFactory, cls).__new__(cls)
+            cls._instance._initialized = False
+        return cls._instance
```
vikrantpuppala: why do we make this singleton and not a static class?

src/databricks/sql/client.py
saishreeeee: made two functions in TelemetryClientFactory, intitialize_telemetry_client (send driver connection params and the other fields here) and get_telemetry_client (send only connection_uuid and retrieve telemetry client)

src/databricks/sql/telemetry/models/enums.py
saishreeeee: added comments and renamed to make it more understandable

src/databricks/sql/telemetry/models/event.py
saishreeeee: raises TypeError (updated the comment to include this)

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: In DatabricksOAuthProvider, both _access_token and _refresh_token are required for token passthrough (have updated the same)

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: So the user sets it in connection params? Should I set it from **kwargs?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: I had faced a circular import issue earlier, hence had changed it.

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Making it static, would make the executor thread and the client map global. I felt that wasn't appropriate initially (I was initializing telemetry manager in connection class if telemetry was enabled in one of my previous implementations). But telemetry_client_factory is anyway initialized globally now so that doesn't make a difference. I can change it to a static class.

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,315 @@
...
+        if not auth_provider:
+            return None
+        if isinstance(auth_provider, AccessTokenAuthProvider):
+            return AuthMech.PAT  # Personal Access Token authentication
+        elif isinstance(auth_provider, DatabricksOAuthProvider):
+            return AuthMech.OAUTH  # Databricks-managed OAuth flow
+        elif isinstance(auth_provider, ExternalAuthProvider):
+            return (
+                AuthMech.EXTERNAL
+            )  # External identity provider (AWS IAM, Azure AD, etc.)
```
vikrantpuppala: I think you changed this to EXTERNAL_AUTH? please add comments where the enum is defined

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,315 @@
...
+        return AuthMech.OTHER  # Custom or unknown authentication provider
```
vikrantpuppala: when can this happen?

src/databricks/sql/telemetry/models/event.py
```diff
@@ -50,16 +64,16 @@ class DriverConnectionParameters:
...
-    auth_flow: AuthFlow
-    auth_scope: str
-    discovery_url: str
-    allowed_volume_ingestion_paths: str
-    azure_tenant_id: str
-    socket_timeout: int
+    auth_mech: Optional[AuthMech] = None
+    auth_flow: Optional[AuthFlow] = None
+    auth_scope: Optional[str] = None
+    discovery_url: Optional[str] = None
```
vikrantpuppala: let's remove discovery_url from here as well, please truncate the dataclasses to only keep parameters that we are actually using

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,315 @@
...
+                        f"Creating new TelemetryClient for connection {connection_uuid}"
+                    )
+                    TelemetryClientFactory._clients[connection_uuid] = TelemetryClient(
+                        telemetry_enabled=telemetry_enabled,
+                        connection_uuid=connection_uuid,
+                        auth_provider=auth_provider,
+                        host_url=host_url,
+                        executor=TelemetryClientFactory._executor,
+                    )
+                return TelemetryClientFactory._clients[connection_uuid]
```
vikrantpuppala: let's not return the telemetry client here, the function says initialize so we should only initialize, library users should call initialize in one place and then call get_telemetry_client as needed

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,315 @@
...
+        else:
+            return NoopTelemetryClient()
+
+    @staticmethod
+    def get_telemetry_client(connection_uuid):
+        """Get the telemetry client for a specific connection"""
+        if connection_uuid in TelemetryClientFactory._clients:
+            return TelemetryClientFactory._clients[connection_uuid]
+        else:
+            return NoopTelemetryClient()
```
vikrantpuppala, line 329: i think we should also print an error log here notifying users that they must initialize the telemetry client first and then call this function

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,324 @@
...
+        self.user_agent = None
+        self.events_batch = []
+        self.lock = threading.Lock()
+        self.driver_connection_params = None
+        self.host_url = host_url
+        self.executor = executor
+
+    def export_event(self, event):
+        """Add an event to the batch queue and flush if batch is full"""
+        logger.debug(f"Exporting event for connection {self.connection_uuid}")
```
jprakash-db: use lazy logging

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,324 @@
...
+        with self.lock:
+            self.events_batch.append(event)
+        if len(self.events_batch) >= self.batch_size:
+            logger.debug(
+                f"Batch size limit reached ({self.batch_size}), flushing events"
```
jprakash-db: Use lazy logging - same comment across all the logging places.

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,324 @@
...
+    def _send_telemetry(self, events):
+        """Send telemetry events to the server"""
+
+        request = {
+            "uploadTime": int(time.time() * 1000),
+            "items": [],
+            "protoLogs": [event.to_json() for event in events],
+        }
+
+        path = "/telemetry-ext" if self.auth_provider else "/telemetry-unauth"
```
jprakash-db: can you define this as a constant in the class and then use it

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,324 @@
...
+                    driver_connection_params=self.driver_connection_params,
+                )
+            ),
+        )
+
+        self.export_event(telemetry_frontend_log)
+
+    def close(self):
+        """Flush remaining events before closing"""
+        logger.info(f"Closing TelemetryClient for connection {self.connection_uuid}")
```
jprakash-db: Any general logging should be done using `debug` and not `info`

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,324 @@
...
+        url = f"https://{self.host_url}{path}"
+
+        headers = {"Accept": "application/json", "Content-Type": "application/json"}
+
+        if self.auth_provider:
+            self.auth_provider.add_headers(headers)
+
+        try:
+            logger.debug("Submitting telemetry request to thread pool")
+            self.executor.submit(
```
jprakash-db: Can we improve this pattern with a callback function? Currently we have no idea if the task completed successfully or not after submition. Add a callback to the future that just logs to have proper idea of failures

src/databricks/sql/client.py
```diff
@@ -294,6 +303,31 @@ def read(self) -> Optional[OAuthToken]:
...
         )
 
+        TelemetryClientFactory.initialize_telemetry_client(
+            telemetry_enabled=self.telemetry_enabled,
+            connection_uuid=self.get_session_id_hex(),
+            auth_provider=auth_provider,
+            host_url=self.host,
+        )
+
+        self.telemetry_client = TelemetryClientFactory.get_telemetry_client(
```
jprakash-db: nit: any variable with strictly internal use (private) should be defined with a underscore - `_telemetry_client`

src/databricks/sql/telemetry/models/event.py
```diff
@@ -9,6 +9,20 @@
     ExecutionResultFormat,
 )
 from typing import Optional
+from enum import Enum
+
+
+class EnumEncoder(json.JSONEncoder):
```
jprakash-db: Don't think `EnumEncoder` belongs to the event class or even be a part of models

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,324 @@
...
+    def __init__(
+        self,
+        telemetry_enabled,
+        connection_uuid,
+        auth_provider,
+        host_url,
+        executor,
+    ):
+        logger.info(f"Initializing TelemetryClient for connection: {connection_uuid}")
+        self.telemetry_enabled = telemetry_enabled
```
jprakash-db: nit: All the variables with private access should be named with a _ -> Ex  `self._lock = threading.Lock()`

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,324 @@
...
+            self.flush()
+
+    def flush(self):
+        """Flush the current batch of events to the server"""
+        with self.lock:
+            events_to_flush = self.events_batch.copy()
+            self.events_batch = []
+
+        if events_to_flush:
+            logger.info(f"Flushing {len(events_to_flush)} telemetry events to server")
```
jprakash-db: We follow the general logs to be  `debug` logs instead of `info`

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,324 @@
...
+    @staticmethod
+    def close(connection_uuid):
+        """Close and remove the telemetry client for a specific connection"""
+
+        with TelemetryClientFactory._lock:
+            if connection_uuid in TelemetryClientFactory._clients:
+                logger.debug(
+                    f"Removing telemetry client for connection {connection_uuid}"
+                )
+                del TelemetryClientFactory._clients[connection_uuid]
```
jprakash-db: nit: can you use pop over here. Precaution because we don't want telemetry to break anything related to actual flow

tests/unit/test_telemetry.py
```diff
@@ -0,0 +1,287 @@
+import unittest
+import uuid
+import requests
+from unittest.mock import patch, MagicMock, call
```
jprakash-db, line 4: We are planning to move away from unittest. Can you write all tests in pytest

tests/unit/test_telemetry.py
```diff
@@ -0,0 +1,287 @@
...
+        self.assertTrue(TelemetryClientFactory._initialized)
+        self.assertIsNotNone(TelemetryClientFactory._executor)
+        self.assertEqual(TelemetryClientFactory._executor, mock_executor2)
+        self.assertIn(connection_uuid2, TelemetryClientFactory._clients)
+        self.assertEqual(TelemetryClientFactory._clients[connection_uuid2], mock_client2)
+        
+        # Verify new ThreadPoolExecutor was created
+        self.assertEqual(mock_executor_class.call_count, 1)
+
+if __name__ == "__main__":
```
jprakash-db: This is not needed, we don't tun tests as main scripts. It is run using pytest. 
[Prev comment, move all tests to pytest]

tests/unit/test_telemetry.py
```diff
@@ -0,0 +1,287 @@
...
+from databricks.sql.auth.authenticators import (
+    AccessTokenAuthProvider,
+)
+
+class TestNoopTelemetryClient(unittest.TestCase):
+    """Tests for the NoopTelemetryClient class."""
+
+    def test_singleton(self):
+        """Test that NoopTelemetryClient is a singleton."""
+        client1 = NoopTelemetryClient()
```
jprakash-db, line 76: Creation of `NoopTelemetryClient` can be a pytest fixture scoped to before every function call

tests/unit/test_telemetry.py
```diff
@@ -0,0 +1,287 @@
...
+    def test_close(self):
+        """Test that close does nothing."""
+        client = NoopTelemetryClient()
+        client.close()
+
+
+class TestTelemetryClient(unittest.TestCase):
+    """Tests for the TelemetryClient class."""
+
+    def setUp(self):
```
jprakash-db: Once you move to pytest, replace the setUp to pytest fixtures. Check examples from other test files

jprakash-db:
nit: You can add the PECOBLR-457 in the PR heading instead of the

src/databricks/sql/telemetry/models/event.py
saishreeeee: shifted it to telemetry/utils.py

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Added _telemetry_request_callback function to handle this

jprakash-db (approved):
LGTM. Thanks for making the changes.

PR #584, opened 2025-06-04
fix `test_closing_connection_closes_commands`

varun-edachali-dbx:
## What type of PR is this?
- [x] Bug Fix

## Description
Currently, `test_closing_connection_closes_commands` is incorrect because: 
- it iterates over values of `closed` but does not use them distinctly in the `subTest`s. The intention is likely to set them as the initial state of `has_been_closed_server_side`, based on the comment. 
- the entire result set is `Mock`ed, so when we check the final value of `has_been_closed_server_side`, it may extract an attribute that does not exist. This assertion will pass even if we do not set `has_been_closed_server_side` to `True` in our `close()` in the `ResultSet`. This is because the attribute is created as a new mock on being accessed, and mocks are by-default truthy. In order to test this, even if we change the assertion to `self.assertTrue(mock_result_set_class.return_value.random_attr)`, the test will pass, despite `random_attr` not existing on the type. 

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

N/A

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -13,6 +13,7 @@ from databricks.sql.thrift_api.TCLIService.ttypes import (
     TExecuteStatementResp,
     TOperationHandle,
     THandleIdentifier,
+    TOperationState,
     TOperationType,
 )
 from databricks.sql.thrift_backend import ThriftBackend
@@ -23,6 +24,7 @@ from databricks.sql import InterfaceError, DatabaseError, Error, NotSupportedErr
 from databricks.sql.exc import RequestError, CursorAlreadyClosedError
 from databricks.sql.types import Row
 
+from databricks.sql.utils import ExecuteResponse
 from tests.unit.test_fetches import FetchTests
 from tests.unit.test_thrift_backend import ThriftBackendTestSuite
 from tests.unit.test_arrow_queue import ArrowQueueSuite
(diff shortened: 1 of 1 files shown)


tests/unit/test_client.py
```diff
@@ -168,22 +170,69 @@ def test_useragent_header(self, mock_client_class):
...
-    @patch("%s.client.ThriftBackend" % PACKAGE_NAME, ThriftBackendMockFactory.new())
-    @patch("%s.client.ResultSet" % PACKAGE_NAME)
-    def test_closing_connection_closes_commands(self, mock_result_set_class):
+    @patch("%s.client.ThriftBackend" % PACKAGE_NAME)
+    def test_closing_connection_closes_commands(self, mock_thrift_client_class):
         # Test once with has_been_closed_server side, once without
         for closed in (True, False):
             with self.subTest(closed=closed):
-                mock_result_set_class.return_value = Mock()
+                initial_state = (
```
jprakash-db, line 188: Since you are changing it, can you change the code to use pytest with parameterized instead of subtest

tests/unit/test_client.py
```diff
@@ -168,22 +170,69 @@ def test_useragent_header(self, mock_client_class):
...
+                    TOperationState.FINISHED_STATE
+                    if not closed
+                    else TOperationState.CLOSED_STATE
+                )
+
+                # Mock the execute response with controlled state
+                mock_execute_response = Mock(spec=ExecuteResponse)
+                mock_execute_response.status = initial_state
```
jprakash-db, line 196: The general rule for writing tests is to try to follow this as much as possible
- All initializations
-  All the function calls
- All the assertions

tests/unit/test_client.py
```diff
@@ -168,22 +170,69 @@ def test_useragent_header(self, mock_client_class):
...
+                mock_backend = Mock(spec=ThriftBackend)
+                mock_thrift_client_class.return_value = mock_backend
+
+                # Create connection and cursor
                 connection = databricks.sql.connect(**self.DUMMY_CONNECTION_ARGS)
                 cursor = connection.cursor()
-                cursor.execute("SELECT 1;")
+
+                # Verify initial state
+                self.assertEqual(
```
jprakash-db: Why is there any need to verify initial state, it is just an initialization. There is no need to verify this

tests/unit/test_client.py
```diff
@@ -168,22 +170,69 @@ def test_useragent_header(self, mock_client_class):
...
+                self.assertEqual(active_result_set.has_been_closed_server_side, closed)
+
+                # Close the connection 
                 connection.close()
 
-                self.assertTrue(
-                    mock_result_set_class.return_value.has_been_closed_server_side
+                # Verify the close logic worked:
+                # 1. has_been_closed_server_side should always be True after close()
+                self.assertTrue(active_result_set.has_been_closed_server_side)
```
jprakash-db: Same feedback, once you move to pytest don't use self.assertTrue or anything core to unittest

tests/unit/test_client.py
varun-edachali-dbx: I have to move the function outside the `ClientTestSuite` class then, right? Since `ClientTestSuite` is a `unittest.TestCase`.

jprakash-db: I don't think so. unittest and pytest are compatible with each other in most cases. Can you run the tests and check whether pytest is working within unittest class

varun-edachali-dbx: I moved the function into the `ClientTestSuite` and the test did run, but the parametrization did not kick in (it only ran for the first value of `closed`, not the second). 

This [StackOverflow discussion](https://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests) talks about how `parametrize` does not work in `unittest.TestCase` classes. 

There are [separate libraries](https://github.com/adamchainz/unittest-parametrize) to help parametrize tests within `unittest.TestCase` classes, but I do not think that is an option for us.

jprakash-db: Hmm, ok then go ahead with the subTest syntax for now

PR #587, opened 2025-06-09
Normalise Execution Response (clean backend interfaces)

varun-edachali-dbx:
## What type of PR is this?

- [x] Refactor

## Description
The overall goal is to clean up the backend interfaces to facilitate the introduction of the SEA client with minimal overhead. 
- Earlier, `ExecuteResponse` was used as a Thrift aligned object for the return type of execution commands. Now, it is used as a normalised object store for common state that is passed as a parameter to the `ResultSet` type to prevent too many primitive types. 
- More common functionality and attributes have been pulled up to the common base `ResultSet` class. 
- Move `Queue` initialisation away from the execution phase, instead doing it directly in the `ResultSet` constructor, to better decouple the exec and fetch phases. 

```
               # Before                                       # After
                                                
ExecuteResponse = namedtuple(                   @dataclass
  "ExecuteResponse",                            class ExecuteResponse:
  "command_handle                                 command_id: CommandId
   status                                         status: CommandState
   description                                    description: Optional[List[Tuple]] = None
   has_more_rows                                 
   arrow_queue                                    
   has_been_closed_server_side                    has_been_closed_server_side: bool = False
   lz4_compressed                                 lz4_compressed: bool = True
   is_staging_operation                           is_staging_operation: bool = False
   arrow_schema_bytes"                            arrow_schema_bytes: Optional[bytes] = None
)                                                 result_format
```

## How is this tested?
- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A


src/databricks/sql/backend/thrift_backend.py
```diff
@@ -5,20 +5,20 @@
...
     from databricks.sql.client import Cursor
-    from databricks.sql.result_set import ResultSet, ThriftResultSet
 
 from databricks.sql.thrift_api.TCLIService.ttypes import TOperationState
 from databricks.sql.backend.types import (
     CommandState,
     SessionId,
     CommandId,
     BackendType,
+    guid_to_hex_id,
```
jayantsing-db: why is this moved to types?

jayantsing-db (changes requested):
Could you please add more details in description, like what is the new structure? Without the follow-up PRs, it is difficult to judge the need to do this. What the pros of this refactor?

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: Thanks for the catch! Not sure why the type checkers didn't catch the invalid import. I've changed the import back.

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -797,23 +797,27 @@ def _results_message_to_execute_response(self, resp, operation_state):
...
+                command_id=command_id,
+                status=status,
+                description=description,
+                has_more_rows=has_more_rows,
+                results_queue=arrow_queue_opt,
+                has_been_closed_server_side=has_been_closed_server_side,
+                lz4_compressed=lz4_compressed,
+                is_staging_operation=is_staging_operation,
+            ),
+            schema_bytes,
```
jayantsing-db: any reason to return schema bytes separately?

varun-edachali-dbx: I thought it was only relevant to Thrift, but I now realise it is relevant to SEA too for the creation of the arrow table during the initialisation of the `CloudFetchQueue`. I'll move it back in and update the tests accordingly.

jayantsing-db:
thank you. i think internal types like this are subject to multiple iterations because of the scope of this migration. Please feel free to add the changes of the result-set using this internal type for better context.

src/databricks/sql/backend/thrift_backend.py
jayantsing-db: >Not sure why the type checkers didn't catch the invalid import

could you check this? i think the imports in types might be causing this?

samikshya-db:
> Should I add more detail to the PR description?

Yes, on testing - was this tested end to end using the Thrift flow? Is there any change that the end user might see?

src/databricks/sql/backend/types.py
```diff
@@ -80,6 +81,26 @@ def from_thrift_state(
         else:
             return None
 
+    @classmethod
+    def from_sea_state(cls, state: str) -> Optional["CommandState"]:
```
samikshya-db, line 85: Can we separate utils away from types.py? Let's restrict this file to contain common classes across SEA and Thrift.

src/databricks/sql/result_set.py
```diff
@@ -138,7 +152,7 @@ class ThriftResultSet(ResultSet):
     def __init__(
         self,
         connection: "Connection",
-        execute_response: ExecuteResponse,
+        execute_response: "ExecuteResponse",
```
samikshya-db, line 155: Food for thought : Have you explored `from __future__ import annotations` ? Given that our driver supports py v3.8+, this would bring cleaner code.

src/databricks/sql/backend/types.py
varun-edachali-dbx: The `CommandState` is common across services right? I think it is more appropriate in `types` since it is used as a normalised representation of command state by both the SEA and Thrift backends. 

Is there a reason it is more appropriate in utils?

src/databricks/sql/result_set.py
varun-edachali-dbx: I wasn't aware of this, thanks! I definitely agree that it could help clean up some code. Should I raise a ticket for now and integrate it in a future cleanup, or should I add it in the current set of PRs? 

It does seem to be in use in one module (`src/databricks/sql/utils.py`) in the codebase already.

varun-edachali-dbx:
> > Should I add more detail to the PR description?
> 
> Yes, on testing - was this tested end to end using the Thrift flow? Is there any change that the end user might see?

I did try to ensure I maintain all existing APIs, but did not try anything manually. This should be handled by the e2e tests running automatically in this PR right? I did not change any e2e tests.

src/databricks/sql/result_set.py
samikshya-db: Yes, I am good with you using it in a later PR ✅

tests/e2e/test_driver.py
```diff
@@ -933,12 +933,12 @@ def test_result_set_close(self):
...
                 assert result_set is not None
 
-                initial_op_state = result_set.op_state
+                initial_op_state = result_set.status
 
                 result_set.close()
 
-                assert result_set.op_state == CommandState.CLOSED
-                assert result_set.op_state != initial_op_state
+                assert result_set.status == CommandState.CLOSED
```
samikshya-db, line 940: Is this a user facing change?

samikshya-db:
> This should be handled by the e2e tests running automatically in this PR right? 

Although I did not check the reason for E2E failure. But as long as you add that e2e fails are not because of the refactor in the PR description, I am good.

src/databricks/sql/backend/types.py
samikshya-db: `from_sea_state ` is a util function, what am I missing?

jayantsing-db: +1 on utils vs types

varun-edachali-dbx: Ah, that is a (static) member function of `CommandState` that is used to create the `CommandState` type, so we can use `CommandState.from_sea_state(sea state)` to return a normalised version of the state. 

I agree that this is a utility function but it's functionality is constrained to this type and I think it makes sense for it to be coupled to this type. Do you think it would be better to put this function in a separate file?

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -778,42 +777,27 @@ def _results_message_to_execute_response(self, resp, operation_state):
...
-            arrow_queue=arrow_queue_opt,
-            status=CommandState.from_thrift_state(operation_state),
-            has_been_closed_server_side=has_been_closed_server_side,
-            has_more_rows=has_more_rows,
-            lz4_compressed=lz4_compressed,
-            is_staging_operation=is_staging_operation,
             command_id=command_id,
+            status=status,
             description=description,
+            has_more_rows=has_more_rows,
```
jayantsing-db: has_more_rows is scoped to a fetch request which is scoped to result-set/fetch. if queue is removed, by extension this should be removed too. For direct results, i think we don't care about has_more_rows

src/databricks/sql/backend/types.py
jayantsing-db: for this factory method, seems fine but in general i feel the distinction would be helpful

tests/e2e/test_driver.py
varun-edachali-dbx: No, `ResultSet` is not a [spec object](https://peps.python.org/pep-0249/) and users should not be accessing them directly.

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: Yes I agree, will make that change. Thanks!

varun-edachali-dbx:
> > This should be handled by the e2e tests running automatically in this PR right?
> 
> Although I did not check the reason for E2E failure. But as long as you add that e2e fails are not because of the refactor in the PR description, I am good.

Seems like the tests are flaky. I just re-ran the tests and they all passed. 

I'll ensure they're passing before I merge anything.

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: Should `has_been_closed_server_side` and `is_staging_operation` also be moved out of `ExecuteResponse`? 

The former does not really have an equivalent in SEA, and I remember discussing that the latter is not relevant to SEA (for now) either because it represents volume operations.

samikshya-db: let's keep it for now, and refactor later if need be. (given the size of this PR already)

src/databricks/sql/backend/types.py
samikshya-db: Ideally it is good to keep the util/transformation logic and type definitions separately. But this can be an exception for now (given the bound util). 👍

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -1209,6 +1224,8 @@ def fetch_results(
                 )
             )
 
+        from databricks.sql.utils import ResultSetQueueFactory
```
samikshya-db: duplicate import?

samikshya-db:
`arrow_schema_bytes ` - this is also specific to thrift flow, isn't it?

varun-edachali-dbx:
> `arrow_schema_bytes ` - this is also specific to thrift flow, isn't it?

My understanding is that it will be used by SEA for the construction of the arrow table for the `EXTERNAL_LINKS` disposition. Is that right?

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: The type checker likely did not complain because since `guid_to_hex_id` is imported in `types`, `mypy` is able to dynamically trace it back to `utils`. 
- if I do `from databricks.sql.backend.types import non_existant_type`, `mypy` successfully throws an error
- if I do `from databricks.sql.backend.types import guid_to_hex_id`, `mypy` does not throw an error, it says everything is fine. 

Python likely allows such indirect imports, because this is how `__init__.py` exports work as well, we simply import what we want the module to expose. So the code is not technically wrong, but it is definitely more clear to import directly from the source file.

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -778,42 +779,28 @@ def _results_message_to_execute_response(self, resp, operation_state):
...
+            status=status,
             description=description,
+            has_been_closed_server_side=has_been_closed_server_side,
+            lz4_compressed=lz4_compressed,
+            is_staging_operation=t_result_set_metadata_resp.isStagingOperation,
             arrow_schema_bytes=schema_bytes,
+            result_format=t_result_set_metadata_resp.resultFormat,
         )
 
+        return execute_response, has_more_rows
```
jayantsing-db: has_more_rows here is slightly confusing and not used in the standard way. it is just used to indicate whether the execution result is direct results or not. instead i would recommend, please add a boolean field in ExecutionResponse `is_direct_results` to achieve the same

src/databricks/sql/backend/types.py
```diff
@@ -394,3 +415,17 @@ def to_hex_guid(self) -> str:
             return guid_to_hex_id(self.guid)
         else:
             return str(self.guid)
+
+
+@dataclass
+class ExecuteResponse:
+    """Response from executing a SQL command."""
```
jayantsing-db, line 422: please rephrase as this seems to be over-the-wire proto model but instead this is an internal type

src/databricks/sql/result_set.py
```diff
@@ -36,30 +36,45 @@ def __init__(
...
+            backend: The backend client
+            arraysize: The max number of rows to fetch at a time (PEP-249)
+            buffer_size_bytes: The size (in bytes) of the internal buffer + max fetch
+            command_id: The command ID
+            status: The command status
+            has_been_closed_server_side: Whether the command has been closed on the server
+            has_more_rows: Whether the command has more rows
+            results_queue: The results queue
+            description: column description of the results
+            is_staging_operation: Whether the command is a staging operation
```
jayantsing-db: is this the correct pydoc format as opposed to Args?

src/databricks/sql/result_set.py
```diff
@@ -389,11 +422,6 @@ def fetchmany(self, size: int) -> List[Row]:
         else:
             return self._convert_arrow_table(self.fetchmany_arrow(size))
 
-    @property
```
jayantsing-db, line 392: is this moved to parent class?

jayantsing-db (approved):
minor comments inline. rest lgtm.

src/databricks/sql/result_set.py
varun-edachali-dbx: Yes, this is handled by the parent now.

samikshya-db:
> My understanding is that it will be used by SEA for the construction of the arrow table for the EXTERNAL_LINKS disposition. Is that right?

This is not correct. But given the fact that we have discussed the mental model with J on this offline, i am good. Approving.

samikshya-db (approved):

src/databricks/sql/result_set.py
varun-edachali-dbx: Changed it to align with the `Connection` class.

PR #589, opened 2025-06-09
Introduce manual SEA test scripts for Exec Phase 

varun-edachali-dbx:
## What type of PR is this?
- [x] Other - test scripts for eventual feature

## Description
For the SEA implementation, before we introduce e2e tests, we introduce manual testing scripts which allow us to execute queries against a real warehouse as an implementation sanity check. 

The tests are structured as follows: 
```
sea_connector_test.py
│
├─ test_sea_session.py - tests basic session opening and closure 
│
├─ test_sea_sync_query.py - tests synchronous query execution 
│  ├─ test_sea_sync_query_with_cloud_fetch()    [use_cloud_fetch=True]
│  └─ test_sea_sync_query_without_cloud_fetch() [use_cloud_fetch=False]
│
├─ test_sea_async_query.py - tests async query execution 
│  ├─ test_sea_async_query_with_cloud_fetch()    [use_cloud_fetch=True]
│  └─ test_sea_async_query_without_cloud_fetch() [use_cloud_fetch=False]
│
└─ test_sea_metadata.py - tests metadata commands
```
This allows us to experiment with metadata queries, session opening and closure, synchronous queries (`INLINE` as well as `EXTERNAL_LINKS`) and async queries (`INLINE` as well as `EXTERNAL_LINKS`). 

## How is this tested?
- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [x] N/A

## Related Tickets & Documents
https://docs.google.com/document/d/1Y-eXLhNqqhrMVGnOlG8sdFrCxBTN1GdQvuKG4IfHmo0/edit?usp=sharing

PR #590, opened 2025-06-09
Implement SeaDatabricksClient (Complete Execution Spec) 

varun-edachali-dbx:
## What type of PR is this?
- [x] Feature

## Description
- Implement execution relevant methods in `SeaDatabricksClient` as defined in the `DatabricksClient` interface. 
- Does not implement metadata relevant methods. 

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [x] Manually - using the test scripts to be introduced in #589 
- [ ] N/A

The coverage of `SeaDatabricksClient` by `test_sea_backend.py`'s unit tests are as below. 
| Module | Statements | Missing | Coverage | Notes |
| `backend.py` (SeaDatabricksClient class) | 167 | 2 | 99% | Missing lines are TYPE_CHECKING imports (lines 16-17) which are not meant to be covered in runtime tests |


tests/unit/test_sea_backend.py
```diff
@@ -41,8 +51,43 @@ def sea_client(self, mock_http_client):
...
+    def mock_cursor(self):
+        """Create a mock cursor."""
+        cursor = Mock()
+        cursor.active_command_id = None
+        cursor.buffer_size_bytes = 1000
+        cursor.arraysize = 100
+        return cursor
+
+    @pytest.fixture
+    def thrift_session_id(self):
```
samikshya-db, line 79: why is this present in sea_backend test?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -274,41 +365,234 @@ def execute_command(
...
+                sea_parameters.append(
+                    StatementParameter(
+                        name=param.name,
+                        value=param.value,
+                        type=param.type if hasattr(param, "type") else None,
+                    )
+                )
+
+        format = "ARROW_STREAM" if use_cloud_fetch else "JSON_ARRAY"
+        disposition = "EXTERNAL_LINKS" if use_cloud_fetch else "INLINE"
```
samikshya-db: nit : can we add a todo to add hybrid results in this disposition so that we don't miss it later.

src/databricks/sql/backend/sea/backend.py
```diff
@@ -274,41 +365,234 @@ def execute_command(
...
+        result_compression = "LZ4_FRAME" if lz4_compression else None
+
+        request = ExecuteStatementRequest(
+            warehouse_id=self.warehouse_id,
+            session_id=sea_session_id,
+            statement=operation,
+            disposition=disposition,
+            format=format,
+            wait_timeout="0s" if async_op else "10s",
```
samikshya-db: can we house these values as constants? Check this https://medium.com/@zeid.zandi/how-to-manage-constants-in-python-best-practices-and-advanced-techniques-50fa1591d517

src/databricks/sql/backend/sea/models/responses.py
```diff
@@ -138,9 +138,9 @@ def from_dict(cls, data: Dict[str, Any]) -> "GetStatementResponse":
         """Create a GetStatementResponse from a dictionary."""
         return cls(
             statement_id=data.get("statement_id", ""),
-            status=_parse_status(data),
-            manifest=_parse_manifest(data),
-            result=_parse_result(data),
+            status=parse_status(data),
+            manifest=parse_manifest(data),
```
samikshya-db: Can you explore if `@dataclass` makes this JSON deserialising cleaner?

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Thanks for the reference! I'll house all these values in `enum`s now.

tests/unit/test_sea_backend.py
varun-edachali-dbx: It's used to test how some methods respond to invalid session id's (i.e., ones constructed from thrift handles instead of sea).

src/databricks/sql/backend/sea/models/responses.py
varun-edachali-dbx: Based on some exploration, there seem to be ways to serialise and deserialise `dataclass`es into JSON less explicitly, but I don't think it's a good idea for us: 
1. If we're looking for direct conversion (i.e., just `ClassName(**json)`): we cannot directly unpack nested JSON into nested dataclasses, which is a common use case for us (`Chunk`s in `ManifestData`) - [[StackOverflow discussion](https://stackoverflow.com/questions/69059290/how-to-unpack-nested-json-into-python-dataclass)]. 
2. Sometimes we want to convert the values into specific forms (with non-trivial conversion methods) before we store them in our types, eg: the state needs to be stored as `CommandState` in our models. We would have to define custom methods for this serialisation and de-serialisation, which [can get quite verbose](https://tomaugspurger.net/posts/serializing-dataclasses/). 

There are ways to help mitigate both of the above complexities, mainly by using third-party libraries, but I think the code in it's current form (with explicit conversion) is more readable.

src/databricks/sql/backend/sea/backend.py
```diff
@@ -262,8 +286,79 @@ def get_allowed_session_configurations() -> List[str]:
         """
         return list(ALLOWED_SESSION_CONF_TO_DEFAULT_VALUES_MAP.keys())
 
-    # == Not Implemented Operations ==
-    # These methods will be implemented in future iterations
+    def _extract_description_from_manifest(self, manifest_obj) -> Optional[List]:
```
jayantsing-db: can we introduce a type for manifest_obj: ResultManifest?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -262,8 +286,79 @@ def get_allowed_session_configurations() -> List[str]:
...
+        schema_data = manifest_obj.schema
+        columns_data = schema_data.get("columns", [])
+
+        if not columns_data:
+            return None
+
+        columns = []
+        for col_data in columns_data:
+            if not isinstance(col_data, dict):
+                continue
```
jayantsing-db: let's add debug logging if we are skipping some col

src/databricks/sql/backend/sea/backend.py
```diff
@@ -262,8 +286,79 @@ def get_allowed_session_configurations() -> List[str]:
...
+
+            # Format: (name, type_code, display_size, internal_size, precision, scale, null_ok)
+            columns.append(
+                (
+                    col_data.get("name", ""),  # name
+                    col_data.get("type_name", ""),  # type_code
+                    None,  # display_size (not provided by SEA)
+                    None,  # internal_size (not provided by SEA)
```
jayantsing-db, line 316: display size and internal size are usually hardcoded for different column types in SEA implementation of JDBC. let's evaluate doing something similar here.

cc @shivam2680 do you know more about this?

src/databricks/sql/backend/sea/backend.py
jayantsing-db: qq: why do we need description from manifest? specifically what public APIs this is exposed?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -262,8 +286,79 @@ def get_allowed_session_configurations() -> List[str]:
...
+                    col_data.get("precision"),  # precision
+                    col_data.get("scale"),  # scale
+                    col_data.get("nullable", True),  # null_ok
+                )
+            )
+
+        return columns if columns else None
+
+    def _results_message_to_execute_response(self, sea_response, command_id):
```
jayantsing-db: let's add data class type for the params or is this the plain JSON dict?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -274,41 +369,242 @@ def execute_command(
...
+                    StatementParameter(
+                        name=param.name,
+                        value=param.value,
+                        type=param.type if hasattr(param, "type") else None,
+                    )
+                )
+
+        format = (
+            ResultFormat.ARROW_STREAM if use_cloud_fetch else ResultFormat.JSON_ARRAY
+        ).value
```
jayantsing-db, line 447: why `.value`?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -274,41 +369,242 @@ def execute_command(
...
+            )
+
+        command_id = CommandId.from_sea_statement_id(statement_id)
+
+        # Store the command ID in the cursor
+        cursor.active_command_id = command_id
+
+        # If async operation, return and let the client poll for results
+        if async_op:
+            return None
```
jayantsing-db, line 491: is this inline with current thrift implementation? are the clients using async mode aware of None return type for execute?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -274,41 +369,242 @@ def execute_command(
...
+
+        # For synchronous operation, wait for the statement to complete
+        status = response.status
+        state = status.state
+
+        # Keep polling until we reach a terminal state
+        while state in [CommandState.PENDING, CommandState.RUNNING]:
+            time.sleep(0.5)  # add a small delay to avoid excessive API calls
```
jayantsing-db: let's make this delay as a constant and then later on maybe a connection prop

src/databricks/sql/backend/sea/backend.py
```diff
@@ -274,41 +369,242 @@ def execute_command(
...
+            state = self.get_query_state(command_id)
+
+        if state != CommandState.SUCCEEDED:
```
jayantsing-db: do we need to handle any other states gracefully? is it possible to reuse the state handling of thrift since i think CommandState is used in thrift client as well.

src/databricks/sql/backend/sea/models/responses.py
jayantsing-db: +1 @varun-edachali-dbx could you please look into this.

jayantsing-db (changes requested):
Added few suggestions and questions

src/databricks/sql/backend/sea/models/responses.py
varun-edachali-dbx: I'll add [this in a future PR](https://databricks.atlassian.net/browse/PECOBLR-562?atlOrigin=eyJpIjoiNDAwMzg4NTc3ZThiNGM2YzhjOGYyNzJiYzVkZDQyZGIiLCJwIjoiaiJ9). 

I was avoiding it for now because I assumed we were against adding more external libraries.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: I'm removing that check entirely - I added it initially because I was having issues with my models, but the check seems un-necessary now.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: The `Cursor` [has a property that exposes the description](https://github.com/databricks/databricks-sql-python/blob/c123af3603c959c8b30296398c1830a8573ea140/src/databricks/sql/client.py#L1219), which is this description of columns. 

It is also used to construct the response to each fetch call, by helping incorporate the column names.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: I was initially passing the raw JSON, but this was a mistake. I changed it to use the response model and now we don't have the parsing code in the backend anymore. Thanks for the catch!

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: `.value` extracts the value of the `enum`, which is what I want to use in our case.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Yes, in the current thrift implementation this is less explicit, but [nothing is returned is the async case](https://github.com/databricks/databricks-sql-python/blob/c123af3603c959c8b30296398c1830a8573ea140/src/databricks/sql/thrift_backend.py#L940), which becomes `None` in Python.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: I gave the code another read, and looks like the Thrift implementation was explicitly checking for command closure and failure. I'll add this to the code.

I do not think we should reuse the state handling of Thrift just yet - in the current Python connector implementation it checks for some specific state that is not *uniquely* accounted for by `CommandState`. eg: [here](https://github.com/databricks/databricks-sql-python/blob/c123af3603c959c8b30296398c1830a8573ea140/src/databricks/sql/thrift_backend.py#L580), the Python connector explicitly checks for `ERROR` state, while in the `CommandState` enum `ERROR`, `TIMEDOUT` and `UNKNOWN` are all normalised to `FAILED`.  Note that we got [this characterisation from the JDBC repo](https://github.com/databricks/databricks-jdbc/blob/a47ab1a85e2d4d3a574cb6fd8378cc55256fc260/src/main/java/com/databricks/jdbc/common/util/DatabricksThriftUtil.java#L86).

We should normalise the Exceptions raised before we abstract the state handling.  Should we do this? 

For now, a quick solution would be to create our own characterisation by making `ERROR_STATE` separate from `TIMEDOUT` and `UNKNOWN`, and this would likely allow us to create some common state handling while maintaining the exact functionality of existing Exceptions. The only issue is that this characterisation would not have any precedent since we're abandoning the JDBC one.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Should I replicate [getDisplaySize](https://github.com/databricks/databricks-jdbc/blob/a47ab1a85e2d4d3a574cb6fd8378cc55256fc260/src/main/java/com/databricks/jdbc/common/util/DatabricksTypeUtil.java#L222)? 

I'm not sure what the equivalent of `internal_size` is in the JDBC repo. 

Note that in the Python connector `None` is returned for `display_size`, `internal_size` and `null_ok` by the Thrift backend.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: I removed it entirely for now because the Thrift implementation does not seem to have a polling interval. Should I add it back? I will add it as a constant if it should be present.

src/databricks/sql/backend/sea/backend.py
jayantsing-db: alright fine to align with existing thrift implementation. where is this format `name, type_code, display_size, internal_size, precision, scale, null_ok` coming from? is this part of spec? please document in docstring.

src/databricks/sql/backend/sea/backend.py
```diff
@@ -262,8 +282,111 @@ def get_allowed_session_configurations() -> List[str]:
...
+                },
+            )
+
+    def _wait_until_command_done(
+        self, response: ExecuteStatementResponse
+    ) -> CommandState:
+        """
+        Wait until a command is done.
+        """
+
```
jayantsing-db, line 384: polling delay removed?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -274,41 +397,221 @@ def execute_command(
         lz4_compression: bool,
         cursor: "Cursor",
         use_cloud_fetch: bool,
-        parameters: List[ttypes.TSparkParameter],
+        parameters: List,
```
jayantsing-db: earlier type removed?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -274,41 +397,221 @@ def execute_command(
...
+
     def get_execution_result(
         self,
         command_id: CommandId,
         cursor: "Cursor",
-    ):
-        """Not implemented yet."""
-        raise NotImplementedError(
-            "get_execution_result is not yet implemented for SEA backend"
+    ) -> "ResultSet":
```
jayantsing-db, line 576: this is for async flow?

tests/unit/test_sea_backend.py
```diff
@@ -1,11 +1,26 @@
+"""
+Tests for the SEA (Statement Execution API) backend implementation.
+
+This module contains tests for the SeaDatabricksClient class, which implements
+the Databricks SQL connector's SEA backend functionality.
+"""
+
 import pytest
-from unittest.mock import patch, MagicMock
+from unittest.mock import patch, MagicMock, Mock
```
jayantsing-db, line 9: is it possible to get the coverage of sea_backend through these set of tests?

jayantsing-db (approved):
some minor questions/suggestions. Rest LGTM

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: It does come [from the spec](https://peps.python.org/pep-0249/#description), yes. I'll add it to the docstring.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: I removed it because the Thrift implementation does not seem to have one. Should I add it back as a constant? 

Update: adding a 200 ms polling delay as a constant

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: `ttypes.TSparkParameter` is strongly aligned with Thrift right? We do not expect this type in the SEA backend. I will add explicit `Dict` typing for now.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Yes

tests/unit/test_sea_backend.py
varun-edachali-dbx: Yes, I've already added it to the description of the PR.

PR #591, opened 2025-06-10
Added functionality for export of failure logs

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Added error logging to the init function of Error class
Added export_failure_log to TelemetryClient

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [x] Manually
        Queried from a non-existent table
          Request Body Summary:
            uploadTime: 1749620674701
            items: 0 items
            protoLogs: 2 logs
          
          Proto Log #1: Initial connection log
          Proto Log #2: Error log
          {
            "frontend_log_event_id": "b251bea7-c9a8-42e4-b00c-5ea4011d7d17",
            "context": {
              "client_context": {
                "timestamp_millis": 1749716213881,
                "user_agent": "PyDatabricksSqlConnector/4.0.3"
              }
            },
            "entry": {
              "sql_driver_log": {
                "session_id": "<REDACTED>",
                "system_configuration": {
                  "driver_version": "4.0.3",
                  "os_name": "Darwin",
                  "os_version": "24.5.0",
                  "os_arch": "arm64",
                  "runtime_version": "3.13.3",
                  "runtime_vendor": "CPython",
                  "char_set_encoding": "utf-8",
                  "locale_name": "en_US"
                },
                "driver_connection_params": {
                  "http_path": "<REDACTED>",
                  "mode": "THRIFT",
                  "host_info": {
                    "host_url": "<REDACTED>",
                    "port": 443
                  },
                  "auth_mech": "PAT"
                },
                "error_info": {
                  "error_name": "ServerOperationError",
                  "stack_trace": "[TABLE_OR_VIEW_NOT_FOUND] The table or view `non_existent_table` cannot be found. Verify the spelling and correctness of the schema and catalog.\nIf you did not qualify the name with a schema, verify the current_schema() output, or qualify the name with the correct schema and catalog.\nTo tolerate the error on drop use DROP VIEW IF EXISTS or DROP TABLE IF EXISTS. SQLSTATE: 42P01; line 2 pos 33"
                }
              }
            }
          }          
          === Response Details ===
          Status Code: 200
          Response Body:
          {
            "errors": [],
            "numSuccess": 0,
            "numProtoSuccess": 2,
            "numRealtimeSuccess": 0
          }
- [ ] N/A

The coverage of `telemetry_client.py` by `test_telemetry.py`'s unit tests are as below. 
| Module | Statements | Miss | Cover | Missing |
| `telemetry_client.py` | 186 | 15 | 92% | 107, 111, 115, 331-339, 364-367, 380-382 |

PECOBLR-524

### DIFF
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -61,7 +61,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       - name: Run tests
-        run: poetry run python -m pytest tests/unit 
+        run: poetry run python -m pytest tests/unit
   run-unit-tests-with-arrow:
     runs-on: ubuntu-latest
     strategy:
@@ -75,7 +75,7 @@ jobs:
           uses: actions/checkout@v2
       -   name: Set up python ${{ matrix.python-version }}
           id: setup-python
-          uses: actions/setup-python@v5
+          uses: actions/setup-python@v2
           with:
             python-version: ${{ matrix.python-version }}
       #----------------------------------------------
@@ -117,7 +117,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13t"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
       #----------------------------------------------
       #       check-out repo and set-up python
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -117,7 +117,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13t", "3.14"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13t"]
     steps:
       #----------------------------------------------
       #       check-out repo and set-up python
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -61,7 +61,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       - name: Run tests
-        run: LOG_LEVEL=DEBUG poetry run python -m pytest tests/unit -v
+        run: poetry run python -m pytest tests/unit -v -s --log-cli-level=DEBUG
   run-unit-tests-with-arrow:
     runs-on: ubuntu-latest
     strategy:
@@ -112,7 +112,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       -   name: Run tests
-          run: LOG_LEVEL=DEBUG poetry run python -m pytest tests/unit -v
+          run: poetry run python -m pytest tests/unit -v -s --log-cli-level=DEBUG
   check-linting:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -61,7 +61,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       - name: Run tests
-        run: poetry run python -m pytest tests/unit
+        run: poetry run python -m pytest tests/unit -v
   run-unit-tests-with-arrow:
     runs-on: ubuntu-latest
     strategy:
@@ -112,7 +112,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       -   name: Run tests
-          run: poetry run python -m pytest tests/unit
+          run: poetry run python -m pytest tests/unit -v
   check-linting:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -117,7 +117,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13t", "3.14"]
     steps:
       #----------------------------------------------
       #       check-out repo and set-up python
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -75,7 +75,7 @@ jobs:
           uses: actions/checkout@v2
       -   name: Set up python ${{ matrix.python-version }}
           id: setup-python
-          uses: actions/setup-python@v2
+          uses: actions/setup-python@v5
           with:
             python-version: ${{ matrix.python-version }}
       #----------------------------------------------
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -61,7 +61,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       - name: Run tests
-        run: poetry run python -m pytest tests/unit -v
+        run: LOG_LEVEL=DEBUG poetry run python -m pytest tests/unit -v
   run-unit-tests-with-arrow:
     runs-on: ubuntu-latest
     strategy:
@@ -112,7 +112,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       -   name: Run tests
-          run: poetry run python -m pytest tests/unit -v
+          run: LOG_LEVEL=DEBUG poetry run python -m pytest tests/unit -v
   check-linting:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -112,7 +112,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       -   name: Run tests
-          run: poetry run python -m pytest tests/unit -v -s 
+          run: poetry run python -m pytest tests/unit
   check-linting:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -61,7 +61,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       - name: Run tests
-        run: poetry run python -m pytest tests/unit -v -s --log-cli-level=DEBUG
+        run: poetry run python -m pytest tests/unit 
   run-unit-tests-with-arrow:
     runs-on: ubuntu-latest
     strategy:
@@ -112,7 +112,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       -   name: Run tests
-          run: poetry run python -m pytest tests/unit -v -s --log-cli-level=DEBUG
+          run: poetry run python -m pytest tests/unit -v -s 
   check-linting:
     runs-on: ubuntu-latest
     strategy:
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -470,7 +470,7 @@ class Connection:
 
         self.open = False
 
-        TelemetryClientFactory.close(self.get_session_id_hex())
+        self._telemetry_client.close()
 
     def commit(self):
         """No-op because Databricks does not support transactions"""
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -53,8 +53,9 @@ from databricks.sql.thrift_api.TCLIService.ttypes import (
     TOperationState,
 )
 from databricks.sql.telemetry.telemetry_client import (
-    TelemetryClientFactory,
     TelemetryHelper,
+    initialize_telemetry_client,
+    get_telemetry_client,
 )
 from databricks.sql.telemetry.models.enums import DatabricksClientType
 from databricks.sql.telemetry.models.event import (
@@ -306,14 +307,14 @@ class Connection:
             kwargs.get("use_inline_params", False)
         )
 
-        TelemetryClientFactory.initialize_telemetry_client(
+        initialize_telemetry_client(
             telemetry_enabled=self.telemetry_enabled,
             connection_uuid=self.get_session_id_hex(),
             auth_provider=auth_provider,
             host_url=self.host,
         )
 
-        self._telemetry_client = TelemetryClientFactory.get_telemetry_client(
+        self._telemetry_client = get_telemetry_client(
             connection_uuid=self.get_session_id_hex()
         )
 
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -421,7 +421,10 @@ class Connection:
         Will throw an Error if the connection has been closed.
         """
         if not self.open:
-            raise Error("Cannot create cursor from closed connection")
+            raise Error(
+                "Cannot create cursor from closed connection",
+                connection_uuid=self.get_session_id_hex(),
+            )
 
         cursor = Cursor(
             self,
@@ -471,7 +474,10 @@ class Connection:
         pass
 
     def rollback(self):
-        raise NotSupportedError("Transactions are not supported on Databricks")
+        raise NotSupportedError(
+            "Transactions are not supported on Databricks",
+            connection_uuid=self.get_session_id_hex(),
+        )
 
 
 class Cursor:
@@ -523,7 +529,10 @@ class Cursor:
             for row in self.active_result_set:
                 yield row
         else:
-            raise Error("There is no active result set")
+            raise Error(
+                "There is no active result set",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
     def _determine_parameter_approach(
         self, params: Optional[TParameterCollection]
@@ -660,7 +669,10 @@ class Cursor:
 
     def _check_not_closed(self):
         if not self.open:
-            raise Error("Attempting operation on closed cursor")
+            raise Error(
+                "Attempting operation on closed cursor",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
     def _handle_staging_operation(
         self, staging_allowed_local_path: Union[None, str, List[str]]
@@ -678,7 +690,8 @@ class Cursor:
             _staging_allowed_local_paths = staging_allowed_local_path
         else:
             raise Error(
-                "You must provide at least one staging_allowed_local_path when initialising a connection to perform ingestion commands"
+                "You must provide at least one staging_allowed_local_path when initialising a connection to perform ingestion commands",
+                connection_uuid=self.connection.get_session_id_hex(),
             )
 
         abs_staging_allowed_local_paths = [
@@ -707,7 +720,8 @@ class Cursor:
                     continue
             if not allow_operation:
                 raise Error(
-                    "Local file operations are restricted to paths within the configured staging_allowed_local_path"
+                    "Local file operations are restricted to paths within the configured staging_allowed_local_path",
+                    connection_uuid=self.connection.get_session_id_hex(),
                 )
 
         # May be real headers, or could be json string
@@ -737,7 +751,8 @@ class Cursor:
         else:
             raise Error(
                 f"Operation {row.operation} is not supported. "
-                + "Supported operations are GET, PUT, and REMOVE"
+                + "Supported operations are GET, PUT, and REMOVE",
+                connection_uuid=self.connection.get_session_id_hex(),
             )
 
     def _handle_staging_put(
@@ -749,7 +764,10 @@ class Cursor:
         """
 
         if local_file is None:
-            raise Error("Cannot perform PUT without specifying a local_file")
+            raise Error(
+                "Cannot perform PUT without specifying a local_file",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
         with open(local_file, "rb") as fh:
             r = requests.put(url=presigned_url, data=fh, headers=headers)
@@ -766,7 +784,8 @@ class Cursor:
 
         if r.status_code not in [OK, CREATED, NO_CONTENT, ACCEPTED]:
             raise Error(
-                f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}"
+                f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}",
+                connection_uuid=self.connection.get_session_id_hex(),
             )
 
         if r.status_code == ACCEPTED:
@@ -784,7 +803,10 @@ class Cursor:
         """
 
         if local_file is None:
-            raise Error("Cannot perform GET without specifying a local_file")
+            raise Error(
+                "Cannot perform GET without specifying a local_file",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
         r = requests.get(url=presigned_url, headers=headers)
 
@@ -792,7 +814,8 @@ class Cursor:
         # Any 2xx or 3xx will evaluate r.ok == True
         if not r.ok:
             raise Error(
-                f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}"
+                f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}",
+                connection_uuid=self.connection.get_session_id_hex(),
             )
 
         with open(local_file, "wb") as fp:
@@ -807,7 +830,8 @@ class Cursor:
 
         if not r.ok:
             raise Error(
-                f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}"
+                f"Staging operation over HTTP was unsuccessful: {r.status_code}-{r.text}",
+                connection_uuid=self.connection.get_session_id_hex(),
             )
 
     def execute(
@@ -1006,7 +1030,8 @@ class Cursor:
             return self
         else:
             raise Error(
-                f"get_execution_result failed with Operation status {operation_state}"
+                f"get_execution_result failed with Operation status {operation_state}",
+                connection_uuid=self.connection.get_session_id_hex(),
             )
 
     def executemany(self, operation, seq_of_parameters):
@@ -1156,7 +1181,10 @@ class Cursor:
         if self.active_result_set:
             return self.active_result_set.fetchall()
         else:
-            raise Error("There is no active result set")
+            raise Error(
+                "There is no active result set",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
     def fetchone(self) -> Optional[Row]:
         """
@@ -1170,7 +1198,10 @@ class Cursor:
         if self.active_result_set:
             return self.active_result_set.fetchone()
         else:
-            raise Error("There is no active result set")
+            raise Error(
+                "There is no active result set",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
     def fetchmany(self, size: int) -> List[Row]:
         """
@@ -1192,21 +1223,30 @@ class Cursor:
         if self.active_result_set:
             return self.active_result_set.fetchmany(size)
         else:
-            raise Error("There is no active result set")
+            raise Error(
+                "There is no active result set",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
     def fetchall_arrow(self) -> "pyarrow.Table":
         self._check_not_closed()
         if self.active_result_set:
             return self.active_result_set.fetchall_arrow()
         else:
-            raise Error("There is no active result set")
+            raise Error(
+                "There is no active result set",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
     def fetchmany_arrow(self, size) -> "pyarrow.Table":
         self._check_not_closed()
         if self.active_result_set:
             return self.active_result_set.fetchmany_arrow(size)
         else:
-            raise Error("There is no active result set")
+            raise Error(
+                "There is no active result set",
+                connection_uuid=self.connection.get_session_id_hex(),
+            )
 
     def cancel(self) -> None:
         """
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -309,13 +309,13 @@ class Connection:
 
         initialize_telemetry_client(
             telemetry_enabled=self.telemetry_enabled,
-            connection_uuid=self.get_session_id_hex(),
+            session_id_hex=self.get_session_id_hex(),
             auth_provider=auth_provider,
             host_url=self.host,
         )
 
         self._telemetry_client = get_telemetry_client(
-            connection_uuid=self.get_session_id_hex()
+            session_id_hex=self.get_session_id_hex()
         )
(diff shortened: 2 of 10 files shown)


src/databricks/sql/client.py
```diff
@@ -421,7 +421,10 @@ def cursor(
         Will throw an Error if the connection has been closed.
         """
         if not self.open:
-            raise Error("Cannot create cursor from closed connection")
+            raise Error(
+                "Cannot create cursor from closed connection",
+                connection_uuid=self.get_session_id_hex(),
```
samikshya-db: are there plans to include the statement ID too across this PR?

samikshya-db:
Let's elaborate on testing details, maybe include a redacted log JSON ?

src/databricks/sql/exc.py
```diff
@@ -11,10 +13,23 @@ class Error(Exception):
...
+
+        error_name = self.__class__.__name__
+        if self.connection_uuid:
+            try:
+                telemetry_client = TelemetryClientFactory.get_telemetry_client(
+                    self.connection_uuid
+                )
+                telemetry_client.export_failure_log(error_name, self.message)
+            except Exception as telemetry_error:
+                logger.error(f"Failed to send error to telemetry: {telemetry_error}")
```
samikshya-db: let's change it to a slightly more silent log (could be debug or trace)

samikshya-db:
From your JSON, can we skip null fields while building the telemetry request?

src/databricks/sql/exc.py
```diff
@@ -11,10 +13,23 @@ class Error(Exception):
...
+
+        error_name = self.__class__.__name__
+        if self.connection_uuid:
+            try:
+                telemetry_client = TelemetryClientFactory.get_telemetry_client(
+                    self.connection_uuid
+                )
+                telemetry_client.export_failure_log(error_name, self.message)
+            except Exception as telemetry_error:
+                logger.debug(f"Failed to send error to telemetry: {telemetry_error}")
```
vikrantpuppala: can we move these try-catch into the telemetry functions so that we do not need to do this everywhere we use these functions?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -26,30 +27,33 @@
...
-        char_set_encoding=sys.getdefaultencoding(),
-    )
+    _DRIVER_SYSTEM_CONFIGURATION = None
 
     @classmethod
     def getDriverSystemConfiguration(cls) -> DriverSystemConfiguration:
+        if cls._DRIVER_SYSTEM_CONFIGURATION is None:
+            from databricks.sql import __version__
+
+            cls._DRIVER_SYSTEM_CONFIGURATION = DriverSystemConfiguration(
```
vikrantpuppala, line 44: why did we need to do this?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -255,10 +266,33 @@ def export_initial_telemetry_log(self, driver_connection_params, user_agent):
...
+                client_context=TelemetryClientContext(
+                    timestamp_millis=int(time.time() * 1000),
+                    user_agent=self._user_agent,
+                )
+            ),
+            entry=FrontendLogEntry(
+                sql_driver_log=TelemetryEvent(
+                    session_id=self._connection_uuid,
+                    system_configuration=TelemetryHelper.getDriverSystemConfiguration(),
+                    driver_connection_params=self._driver_connection_params,
```
vikrantpuppala: do we need to send system_configuration and driver_connection_params with every telemetry log? cc: @samikshya-db

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -284,11 +336,58 @@ def _initialize(cls):
...
+                "Flushing pending telemetry and waiting for thread pool completion..."
+            )
+            for uuid, client in cls._clients.items():
+                if hasattr(client, "flush"):
+                    try:
+                        client.flush()  # Submit any pending events
+                    except Exception as e:
+                        logger.debug(
+                            "Failed to flush telemetry for connection %s: %s", uuid, e
+                        )
```
vikrantpuppala: too many levels of try-catches, what's the idea here? should we do it on all internal functions, all external functions, let's build an understanding first

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -284,11 +336,58 @@ def _initialize(cls):
...
+
+        try:
+            # Flush existing thread pool work and wait for completion
+            logger.debug(
+                "Flushing pending telemetry and waiting for thread pool completion..."
+            )
+            for uuid, client in cls._clients.items():
+                if hasattr(client, "flush"):
+                    try:
+                        client.flush()  # Submit any pending events
```
vikrantpuppala: why not call close here?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -284,11 +336,58 @@ def _initialize(cls):
...
+                try:
+                    cls._executor.shutdown(
+                        wait=True
+                    )  # This waits for all submitted work to complete
+                    logger.debug("Thread pool shutdown completed successfully")
+                except Exception as e:
+                    logger.debug("Thread pool shutdown failed: %s", e)
+
+        except Exception as e:
+            logger.debug("Exception in excepthook telemetry handler: %s", e)
```
vikrantpuppala: this looks to be duplicate to what we do in close function, can we just call that here?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: circular import issue: databricks.sql.__init__.py imports exc.py which imports telemetry_client which imports databricks.sql

src/databricks/sql/telemetry/telemetry_client.py
samikshya-db: Yes, that is correct.

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Removed outer try/catch block and kept the inner one. If Client 1's flush() fails and we only had outer try/catch, the exception would stop the entire loop. So Client 2, Client 3, and executor.shutdown() would never execute.

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: The thread pool only shuts down when the dict becomes empty (last client removed). If Client 1's close() fails, it stays in the dict. When Clients 2 & 3 succeed, the dict still isn't empty (Client 1 still there), so executor.shutdown(wait=True) never gets called.

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -26,30 +27,33 @@
...
-        os_version=platform.release(),
-        os_arch=platform.machine(),
-        client_app_name=None,  # TODO: Add client app name
-        locale_name=locale.getlocale()[0] or locale.getdefaultlocale()[0],
-        char_set_encoding=sys.getdefaultencoding(),
-    )
+    _DRIVER_SYSTEM_CONFIGURATION = None
 
     @classmethod
     def getDriverSystemConfiguration(cls) -> DriverSystemConfiguration:
```
vikrantpuppala: probably would want to have a python style function name

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -284,11 +330,54 @@ def _initialize(cls):
...
                 cls._initialized = True
                 logger.debug(
                     "TelemetryClientFactory initialized with thread pool (max_workers=10)"
                 )
 
+    @classmethod
+    def _install_exception_hook(cls):
+        """Install global exception handler for unhandled exceptions"""
+        if not cls._excepthook_installed:
+            import sys
```
vikrantpuppala: why do we add this import here and not at the top?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -284,11 +330,54 @@ def _initialize(cls):
...
+    def _handle_unhandled_exception(cls, exc_type, exc_value, exc_traceback):
+        """Handle unhandled exceptions by sending telemetry and flushing thread pool"""
+        logger.debug("Handling unhandled exception: %s", exc_type.__name__)
+
+        # Flush existing thread pool work and wait for completion
+        logger.debug(
+            "Flushing pending telemetry and waiting for thread pool completion..."
+        )
+        for uuid, client in cls._clients.items():
+            if hasattr(client, "flush"):
```
vikrantpuppala: why do we need this if condition

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -297,35 +386,48 @@ def initialize_telemetry_client(
...
-                )
-                if telemetry_enabled:
-                    TelemetryClientFactory._clients[connection_uuid] = TelemetryClient(
-                        telemetry_enabled=telemetry_enabled,
-                        connection_uuid=connection_uuid,
-                        auth_provider=auth_provider,
-                        host_url=host_url,
-                        executor=TelemetryClientFactory._executor,
+            with TelemetryClientFactory._lock:
+                if connection_uuid not in TelemetryClientFactory._clients:
```
vikrantpuppala: this is actually not the connection_uuid but the hex right?

src/databricks/sql/telemetry/utils.py
```diff
@@ -13,3 +14,23 @@ def default(self, obj):
...
+    else:
+        return data
+
+
+def to_json_compact(dataclass_obj):
+    """
+    Convert a dataclass to JSON string, excluding None values.
+    """
+    data_dict = asdict(dataclass_obj)
+    filtered_dict = filter_none_values(data_dict)
```
vikrantpuppala: i think you can pass above as lambda to asdict:
```
    return asdict(dataclass_obj, dict_factory=lambda data: {k: v for k, v in data if v is not None})
```

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: The client could be TelemetryClient or NoopTelemetryClient. NoopTelemetryClient does not have a flush function. Should I add a flush function?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: That's my bad, this is a duplicate import (already imported sys once at the top)

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: It is the session id of ThriftBackend converted to a hex string. Every connection has only one session, so I used it as a connection identifier.

src/databricks/sql/telemetry/telemetry_client.py
vikrantpuppala: yes, all public functions should be defined at interface level and be implemented by all subclasses

src/databricks/sql/telemetry/telemetry_client.py
vikrantpuppala: yes however let's not use the variable name `connection_uuid` since that is misleading, we should name it exactly what it is

tests/unit/test_telemetry.py
```diff
@@ -83,6 +83,12 @@ def test_export_initial_telemetry_log(self, noop_telemetry_client):
             driver_connection_params=MagicMock(), user_agent="test"
         )
 
+    def test_export_failure_log(self, noop_telemetry_client):
```
vikrantpuppala, line 93: this test and the test at line 141 have the same name, let's dedup

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -297,35 +386,48 @@ def initialize_telemetry_client(
...
+                        connection_uuid,
                     )
-                else:
-                    TelemetryClientFactory._clients[
-                        connection_uuid
-                    ] = NoopTelemetryClient()
+                    if telemetry_enabled:
+                        TelemetryClientFactory._clients[
+                            connection_uuid
+                        ] = TelemetryClient(
```
vikrantpuppala, line 381: do we need so many line breaks here?

tests/unit/test_telemetry.py
saishreeeee: I thought it's alright since they are in different classes? Have done the same thing for test_export_initial_telemetry_log and test_close. Should I change these too?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: happens due to black formatter

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -284,11 +330,54 @@ def _initialize(cls):
...
+                    logger.debug(
+                        "Failed to flush telemetry for connection %s: %s", uuid, e
+                    )
+
+        if cls._executor:
+            try:
+                cls._executor.shutdown(
+                    wait=True
+                )  # This waits for all submitted work to complete
+                logger.debug("Thread pool shutdown completed successfully")
```
samikshya-db: misleading log message, let's add logs appropriately.

src/databricks/sql/telemetry/telemetry_client.py
samikshya-db: cc @varun-edachali-dbx  - something to take a note of.

src/databricks/sql/client.py
```diff
@@ -1156,7 +1181,10 @@ def fetchall(self) -> List[Row]:
         if self.active_result_set:
             return self.active_result_set.fetchall()
         else:
-            raise Error("There is no active result set")
+            raise Error(
```
samikshya-db: do we plan to have custom tags for each message?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: I'm sorry I don't quite understand. Do you mean the submitted work make have completed but with a failure so I shouldn't add "successfully" in the log message?

src/databricks/sql/client.py
```diff
@@ -421,7 +421,10 @@ def cursor(
         Will throw an Error if the connection has been closed.
         """
         if not self.open:
-            raise Error("Cannot create cursor from closed connection")
+            raise Error(
```
jprakash-db: Can this be a more specific error? Like ConnectionError

src/databricks/sql/exc.py
```diff
@@ -11,10 +13,20 @@ class Error(Exception):
...
     """
 
-    def __init__(self, message=None, context=None, *args, **kwargs):
+    def __init__(
+        self, message=None, context=None, connection_uuid=None, *args, **kwargs
+    ):
         super().__init__(message, *args, **kwargs)
         self.message = message
         self.context = context or {}
+        self.connection_uuid = connection_uuid
```
jprakash-db: What is the need to store `connection_uuid` in a class variable, it looks like a temp variable?

src/databricks/sql/client.py
```diff
@@ -707,7 +720,8 @@ def _handle_staging_operation(
                     continue
             if not allow_operation:
                 raise Error(
-                    "Local file operations are restricted to paths within the configured staging_allowed_local_path"
+                    "Local file operations are restricted to paths within the configured staging_allowed_local_path",
```
jprakash-db, line 726: Same comment across all the places we are throwing an error. Since the error class name matters in the telemetry logs can we throw the correct error and not a blanket `Error()` . Be specific like `ConnectionError`, etc

src/databricks/sql/thrift_backend.py
```diff
@@ -255,12 +256,14 @@ def _initialize_retry_args(self, kwargs):
             )
 
     @staticmethod
-    def _check_response_for_error(response):
+    def _check_response_for_error(response, connection_uuid=None):
```
samikshya-db: probably a dumb q : why is there a need to pass the connection uuid if we already have self._connection_uuid ?

src/databricks/sql/telemetry/telemetry_client.py
samikshya-db: There are 2 things here : 
1. We can mention that telemetry client has been shut down successfully. 
2. Telemetry logs should not be too verbose; debug/error should be avoided as much as possible. 

Another thing that I noticed across this PR is : let's not try to 1:1 replicate what we have in JDBC telemetry. We can follow python best practices instead. For example : 

- you can explore using a mixin base class or a utility function to make JSON serialization reusable in all the telemetry models that you have - that will make the code prettier. 
- Another thing is : `TelemetryClientFactory ` is overly Java specific, we could do `factory functions` instead

Another thing that is missing in this PR is coverage. Can we add coverage on the PR description?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -233,33 +244,66 @@ def export_initial_telemetry_log(self, driver_connection_params, user_agent):
...
-            entry=FrontendLogEntry(
-                sql_driver_log=TelemetryEvent(
-                    session_id=self._connection_uuid,
-                    system_configuration=TelemetryHelper.getDriverSystemConfiguration(),
-                    driver_connection_params=self._driver_connection_params,
-                )
-            ),
-        )
+            self.export_event(telemetry_frontend_log)
+        except Exception as e:
```
jprakash-db, line 272: Why is there a try except block here. From what I can see all the errors can stem from the flush() function and there anyway we are logging and cathching the exception. It does't make sense to have try blocks where there is no error thrown

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -233,33 +244,66 @@ def export_initial_telemetry_log(self, driver_connection_params, user_agent):
...
+            logger.debug("Failed to export initial telemetry log: %s", e)
 
-        self.export_event(telemetry_frontend_log)
+    def export_failure_log(self, error_name, error_message):
+        logger.debug("Exporting failure log for connection %s", self._connection_uuid)
+        try:
```
jprakash-db, line 277: Same comment on error handling. In telemetry the only error is the network call and that is handled in flush error handling. I don't feel random try blocks are needed cc @vikrantpuppala

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -233,33 +244,66 @@ def export_initial_telemetry_log(self, driver_connection_params, user_agent):
...
 
     def close(self):
         """Flush remaining events before closing"""
         logger.debug("Closing TelemetryClient for connection %s", self._connection_uuid)
-        self.flush()
-        TelemetryClientFactory.close(self._connection_uuid)
+        try:
+            self.flush()
+            TelemetryClientFactory.close(self._connection_uuid)
+        except Exception as e:
```
jprakash-db: What error is expected from flush that will be thrown? For `TelemetryClientFactory.close` the error handling should be within the close function . @vikrantpuppala can chime in , but my view is add try block only on the places you expect an error. Randomly encompassing everything in try blocks doesn't feel right. Either re throw the error again or don't have try block here

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -284,11 +330,52 @@ def _initialize(cls):
...
+        logger.debug("Handling unhandled exception: %s", exc_type.__name__)
+
+        # Flush existing thread pool work and wait for completion
+        logger.debug(
+            "Flushing pending telemetry and waiting for thread pool completion..."
+        )
+        for uuid, client in cls._clients.items():
+            if hasattr(client, "flush"):
+                try:
+                    client.flush()  # Submit any pending events
```
jprakash-db: again why try, flush() internally handles all error and nothing is re thrown

src/databricks/sql/thrift_backend.py
```diff
@@ -737,13 +763,15 @@ def _results_message_to_execute_response(self, resp, operation_state):
             or direct_results.resultSet.hasMoreRows
         )
         description = self._hive_schema_to_description(
-            t_result_set_metadata_resp.schema
+            t_result_set_metadata_resp.schema, self._connection_uuid
```
jprakash-db: Why is connection_uuid an argument to this function `_hive_schema_to_description`  is it just for telemetry logging ? Can't we use thread local because we are polluting random functions for logging cc @vikrantpuppala

jprakash-db:
3 broad suggestions
- Can we not throw blanket `Error` , try to use specific errors such as `ConnectionError`,etc as telemetry uses the error name
- Add try except blocks only where you expect an error, or atleast re throw the error from the child functions. Random try blocks don't feel makes sense
- Passing arguments like connection_uuid to functions that don't need it and  just for telemetry doesn't feel. Either store this like a class variable during class initialisation or can we explore thread locals

src/databricks/sql/thrift_backend.py
saishreeeee: static methods don't receive self parameter

src/databricks/sql/telemetry/telemetry_client.py
varun-edachali-dbx: Thanks! Noted.

src/databricks/sql/client.py
saishreeeee: I'll add in a follow up PR

src/databricks/sql/client.py
saishreeeee: will add error codes in a follow-up PR

src/databricks/sql/exc.py
```diff
@@ -1,8 +1,10 @@
 import json
 import logging
+import traceback
 
-logger = logging.getLogger(__name__)
+from databricks.sql.telemetry.telemetry_client import get_telemetry_client
 
+logger = logging.getLogger(__name__)
 
 ### PEP-249 Mandated ###
```
vikrantpuppala, line 8: can you add this link here: https://peps.python.org/pep-0249/#exceptions

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -99,34 +103,30 @@ class BaseTelemetryClient(ABC):
...
+# A single instance of the no-op client that can be reused
+NOOP_TELEMETRY_CLIENT = type(
+    "NoopTelemetryClient",
+    (BaseTelemetryClient,),
+    {
+        "export_initial_telemetry_log": lambda self, *args, **kwargs: None,
+        "export_failure_log": lambda self, *args, **kwargs: None,
+        "close": lambda self: None,
+    },
+)()
```
vikrantpuppala: i don't think i prefer this over the previous way, this is harder to read/understand, what is the motivation behind this change?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -157,18 +157,18 @@ def __init__(
         self._host_url = host_url
         self._executor = executor
 
-    def export_event(self, event):
+    def _export_event(self, event):
         """Add an event to the batch queue and flush if batch is full"""
-        logger.debug("Exporting event for connection %s", self._connection_uuid)
+        logger.debug("Exporting event for connection %s", self._session_id_hex)
```
vikrantpuppala, line 175: basic question - why do we rely on session_id_hex and not on the session_id itself, or are they the same thing?

saishreeeee: session_id_hex is human-readable while session_id is in raw bytes format.

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Changed it to avoid the explicit singleton pattern (checking instance existence during every creation) but yes it is less readable now. So planning on going back to previous implementation without the singletion check,  instead will add NOOP_TELEMETRY_CLIENT = NoopTelemetryClient() and use it.

src/databricks/sql/telemetry/telemetry_client.py
samikshya-db: why is there need to store the raw bytes format?

vikrantpuppala: no, we don't, i had the wrong impression, ok with this

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -113,20 +123,20 @@ class NoopTelemetryClient(BaseTelemetryClient):
...
         pass
 
+    def export_failure_log(self, error_name, error_message):
+        pass
+
     def close(self):
         pass
 
 
+# A single instance of the no-op client that can be reused
```
jprakash-db: Why is this not a singleton instance? We are creating an object even if NoopTelemetryClient is never used

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -230,120 +240,171 @@ def _telemetry_request_callback(self, future):
...
-            entry=FrontendLogEntry(
-                sql_driver_log=TelemetryEvent(
-                    session_id=self._connection_uuid,
-                    system_configuration=TelemetryHelper.getDriverSystemConfiguration(),
-                    driver_connection_params=self._driver_connection_params,
-                )
-            ),
-        )
+        except Exception as e:
+            logger.debug("Failed to export initial telemetry log: %s", e)
```
jprakash-db, line 273: Why is there a try catch block here? _export_event does not throw any error ?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -230,120 +240,171 @@ def _telemetry_request_callback(self, future):
...
+                entry=FrontendLogEntry(
+                    sql_driver_log=TelemetryEvent(
+                        session_id=self._session_id_hex,
+                        system_configuration=TelemetryHelper.get_driver_system_configuration(),
+                        driver_connection_params=self._driver_connection_params,
+                        error_info=error_info,
+                    )
+                ),
+            )
+            self._export_event(telemetry_frontend_log)
```
jprakash-db, line 298: Same here, is this try catch for _export_event or something else? Can you atleast re throw the error within _flush if you want to catch it here. Otherwise what is the point ?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -230,120 +240,171 @@ def _telemetry_request_callback(self, future):
...
+            logger.debug("Removing telemetry client for connection %s", session_id_hex)
+            telemetry_client = _clients.pop(session_id_hex, None)
+            telemetry_client.close()
+
+        # Shutdown executor if no more clients
+        if not _clients and _executor:
+            logger.debug(
+                "No more telemetry clients, shutting down thread pool executor"
+            )
+            _executor.shutdown(wait=True)
```
jprakash-db: _executor.shutdown can throw an error, can you wrap this in a try catch

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -230,120 +240,171 @@ def _telemetry_request_callback(self, future):
...
     def close(self):
         """Flush remaining events before closing"""
-        logger.debug("Closing TelemetryClient for connection %s", self._connection_uuid)
-        self.flush()
-        TelemetryClientFactory.close(self._connection_uuid)
+        logger.debug("Closing TelemetryClient for connection %s", self._session_id_hex)
+        self._flush()
+
+
+# Module-level state
```
jprakash-db: Why are we moving away from a class based approach? Why are all the functions split without the `TelemetryClientFactory` class. ?

src/databricks/sql/telemetry/utils.py
```diff
@@ -1,5 +1,29 @@
 import json
 from enum import Enum
+from dataclasses import asdict, is_dataclass
+from abc import ABC
+
+
+class JsonSerializableMixin(ABC):
+    """Mixin class to provide JSON serialization capabilities to dataclasses."""
+
+    def to_json(self) -> str:
```
jprakash-db, line 9: Why is this a ABC, when there is no abstract methods ?

.github/workflows/code-quality-checks.yml
```diff
@@ -112,7 +112,7 @@ jobs:
       #              run test suite
       #----------------------------------------------
       -   name: Run tests
-          run: poetry run python -m pytest tests/unit
+          run: poetry run python -m pytest tests/unit -v -s --log-cli-level=DEBUG
```
jprakash-db: What is the motivation for having logs in github actions ?

saishreeeee: Previous: connection instance -> TelemetryClient.close -> self.flush and TelemetryClientFactory.close -> pop client from map and shutdown executor if no clients
Current: connection instance -> TelemetryClientFactory.close -> calls TelemetryClient.close, pop client from map and shutdown executor if no clients

when I made this change: a code quality kept running for an unusually long time, hence am trying to figure out the issue with debug logs

src/databricks/sql/telemetry/utils.py
saishreeeee: I will remove it

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -230,120 +243,173 @@ def _telemetry_request_callback(self, future):
...
+                    _clients[session_id_hex] = TelemetryClient(
                         telemetry_enabled=telemetry_enabled,
-                        connection_uuid=connection_uuid,
+                        session_id_hex=session_id_hex,
                         auth_provider=auth_provider,
                         host_url=host_url,
-                        executor=TelemetryClientFactory._executor,
+                        executor=_executor,
                     )
+                    print("i have initialized the telemetry client yes")
```
samikshya-db: Nice debugging though 👯, but we need to remove them

tests/unit/test_client.py
```diff
@@ -789,40 +790,41 @@ def test_cursor_context_manager_handles_exit_exception(self):
 
         cursor.close.assert_called_once()
 
-    def test_connection_close_handles_cursor_close_exception(self):
-        """Test that _close handles exceptions from cursor.close() properly."""
-        cursors_closed = []
+    # def test_connection_close_handles_cursor_close_exception(self):
+    #     """Test that _close handles exceptions from cursor.close() properly."""
```
samikshya-db: can we clean up all of these too?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Though Telemetry Log object creation, Driver System Config object creation, thread lock sync in export_event are unlikely to fail, I think having a safety net is good since telemetry failures should not affect the connector.

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Rethrowing the error in flush would cause issues since other functions like close call flush and that would lead to exceptions.

samikshya-db (approved):
LGTM. thanks for getting this through 🎉

PR #593, opened 2025-06-11
add metadata commands

varun-edachali-dbx:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## Description
- metadata commands for SEA (exec phase) 

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/src/databricks/sql/backend/filters.py
+++ b/src/databricks/sql/backend/filters.py
@@ -1,0 +1,167 @@
+"""
+Client-side filtering utilities for Databricks SQL connector.
+
+This module provides filtering capabilities for result sets returned by different backends.
+"""
+
+import logging
+from typing import (
+    List,
+    Optional,
+    Any,
+    Dict,
+    Callable,
+    TypeVar,
+    Generic,
+    cast,
+    TYPE_CHECKING,
+)
+
+from databricks.sql.utils import JsonQueue, SeaResultSetQueueFactory
+from databricks.sql.backend.types import ExecuteResponse, CommandId
+from databricks.sql.backend.sea.models.base import ResultData
+from databricks.sql.backend.sea.backend import SeaDatabricksClient
+
+if TYPE_CHECKING:
+    from databricks.sql.result_set import ResultSet, SeaResultSet
+
+logger = logging.getLogger(__name__)
+
+
+class ResultSetFilter:
+    """
+    A general-purpose filter for result sets that can be applied to any backend.
+
+    This class provides methods to filter result sets based on various criteria,
+    similar to the client-side filtering in the JDBC connector.
+    """
+
+    @staticmethod
+    def _filter_sea_result_set(
+        result_set: "SeaResultSet", filter_func: Callable[[List[Any]], bool]
+    ) -> "SeaResultSet":
+        """
+        Filter a SEA result set using the provided filter function.
+
+        Args:
+            result_set: The SEA result set to filter
+            filter_func: Function that takes a row and returns True if the row should be included
+
+        Returns:
+            A filtered SEA result set
+        """
+        # Get all remaining rows
+        all_rows = result_set.results.remaining_rows()
+
+        # Filter rows
+        filtered_rows = [row for row in all_rows if filter_func(row)]
+
+        # Import SeaResultSet here to avoid circular imports
+        from databricks.sql.result_set import SeaResultSet
+
+        # Reuse the command_id from the original result set
+        command_id = result_set.command_id
+
+        # Create an ExecuteResponse with the filtered data
+        execute_response = ExecuteResponse(
+            command_id=command_id,
+            status=result_set.status,
+            description=result_set.description,
+            has_been_closed_server_side=result_set.has_been_closed_server_side,
+            lz4_compressed=result_set.lz4_compressed,
+            arrow_schema_bytes=result_set._arrow_schema_bytes,
+            is_staging_operation=False,
+        )
+
+        # Create a new ResultData object with filtered data
+        from databricks.sql.backend.sea.models.base import ResultData
+
+        result_data = ResultData(data=filtered_rows, external_links=None)
+
+        # Create a new SeaResultSet with the filtered data
+        filtered_result_set = SeaResultSet(
+            connection=result_set.connection,
+            execute_response=execute_response,
+            sea_client=cast(SeaDatabricksClient, result_set.backend),
+            buffer_size_bytes=result_set.buffer_size_bytes,
+            arraysize=result_set.arraysize,
+            result_data=result_data,
+        )
+
+        return filtered_result_set
+
+    @staticmethod
+    def filter_by_column_values(
+        result_set: "ResultSet",
+        column_index: int,
+        allowed_values: List[str],
+        case_sensitive: bool = False,
+    ) -> "ResultSet":
+        """
+        Filter a result set by values in a specific column.
+
+        Args:
+            result_set: The result set to filter
+            column_index: The index of the column to filter on
+            allowed_values: List of allowed values for the column
+            case_sensitive: Whether to perform case-sensitive comparison
+
+        Returns:
+            A filtered result set
+        """
+        # Convert to uppercase for case-insensitive comparison if needed
+        if not case_sensitive:
+            allowed_values = [v.upper() for v in allowed_values]
+
+        # Determine the type of result set and apply appropriate filtering
+        from databricks.sql.result_set import SeaResultSet
+
+        if isinstance(result_set, SeaResultSet):
+            return ResultSetFilter._filter_sea_result_set(
+                result_set,
+                lambda row: (
+                    len(row) > column_index
+                    and isinstance(row[column_index], str)
+                    and (
+                        row[column_index].upper()
+                        if not case_sensitive
+                        else row[column_index]
+                    )
+                    in allowed_values
+                ),
+            )
+
+        # For other result set types, return the original (should be handled by specific implementations)
+        logger.warning(
+            f"Filtering not implemented for result set type: {type(result_set).__name__}"
+        )
+        return result_set
+
+    @staticmethod
+    def filter_tables_by_type(
+        result_set: "ResultSet", table_types: Optional[List[str]] = None
+    ) -> "ResultSet":
+        """
+        Filter a result set of tables by the specified table types.
+
+        This is a client-side filter that processes the result set after it has been
+        retrieved from the server. It filters out tables whose type does not match
+        any of the types in the table_types list.
+
+        Args:
+            result_set: The original result set containing tables
+            table_types: List of table types to include (e.g., ["TABLE", "VIEW"])
+
+        Returns:
+            A filtered result set containing only tables of the specified types
+        """
+        # Default table types if none specified
+        DEFAULT_TABLE_TYPES = ["TABLE", "VIEW", "SYSTEM TABLE"]
+        valid_types = (
+            table_types if table_types and len(table_types) > 0 else DEFAULT_TABLE_TYPES
+        )
+
+        # Table type is the 6th column (index 5)
+        return ResultSetFilter.filter_by_column_values(
+            result_set, 5, valid_types, case_sensitive=True
+        )
--- a/src/databricks/sql/backend/sea/backend.py
+++ b/src/databricks/sql/backend/sea/backend.py
@@ -724,7 +724,20 @@ class SeaDatabricksClient(DatabricksClient):
         cursor: "Cursor",
     ) -> "ResultSet":
         """Get available catalogs by executing 'SHOW CATALOGS'."""
-        raise NotImplementedError("get_catalogs is not implemented for SEA backend")
+        result = self.execute_command(
+            operation="SHOW CATALOGS",
+            session_id=session_id,
+            max_rows=max_rows,
+            max_bytes=max_bytes,
+            lz4_compression=False,
+            cursor=cursor,
+            use_cloud_fetch=False,
+            parameters=[],
+            async_op=False,
+            enforce_embedded_schema_correctness=False,
+        )
+        assert result is not None, "execute_command returned None in synchronous mode"
+        return result
 
     def get_schemas(
         self,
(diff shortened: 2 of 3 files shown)


src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+    Any,
+    Callable,
+    cast,
+    TYPE_CHECKING,
+)
+
+from databricks.sql.backend.sea.backend import SeaDatabricksClient
+from databricks.sql.backend.types import ExecuteResponse
+
+if TYPE_CHECKING:
```
jayantsing-db: let's remove this

src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+"""
+
+import logging
+from typing import (
+    List,
+    Optional,
+    Any,
+    Callable,
+    cast,
+    TYPE_CHECKING,
```
jayantsing-db: we can remove this

src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+
+if TYPE_CHECKING:
+    from databricks.sql.result_set import ResultSet, SeaResultSet
+
+logger = logging.getLogger(__name__)
+
+
+class ResultSetFilter:
+    """
+    A general-purpose filter for result sets that can be applied to any backend.
```
jayantsing-db:
```suggestion
    A general-purpose filter for result sets.
```

src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+
+    This class provides methods to filter result sets based on various criteria,
+    similar to the client-side filtering in the JDBC connector.
```
jayantsing-db: irrelevant info. no need to mention JDBC or any other connector.

src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+
+    @staticmethod
+    def _filter_sea_result_set(
```
jayantsing-db: this is specific to SEA result set and can't be used for a generic result set class? let's try to make it generic for a result set

src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+        result_data = ResultData(data=filtered_rows, external_links=None)
+
+        # Create a new SeaResultSet with the filtered data
+        filtered_result_set = SeaResultSet(
+            connection=result_set.connection,
+            execute_response=execute_response,
+            sea_client=cast(SeaDatabricksClient, result_set.backend),
+            buffer_size_bytes=result_set.buffer_size_bytes,
+            arraysize=result_set.arraysize,
+            result_data=result_data,
```
jayantsing-db: could you remind me what is the significance of this result_data param in result set? is this present in the base class? Is this an optional param and is used to create a result set with hard-coded rows?

src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+        )
```
jayantsing-db: I think the whole implementation can be improved. you are essentially first downloading the complete result set and then initialising a new one. a filter method should ideally just take the object to be filtered and return true/false on it.

src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+        Filter a result set by values in a specific column.
+
+        Args:
+            result_set: The result set to filter
+            column_index: The index of the column to filter on
+            allowed_values: List of allowed values for the column
+            case_sensitive: Whether to perform case-sensitive comparison
+
+        Returns:
+            A filtered result set
```
jayantsing-db: same as above

src/databricks/sql/backend/filters.py
```diff
@@ -0,0 +1,165 @@
...
+        # Default table types if none specified
+        DEFAULT_TABLE_TYPES = ["TABLE", "VIEW", "SYSTEM TABLE"]
+        valid_types = (
+            table_types if table_types and len(table_types) > 0 else DEFAULT_TABLE_TYPES
+        )
+
+        # Table type is the 6th column (index 5)
+        return ResultSetFilter.filter_by_column_values(
+            result_set, 5, valid_types, case_sensitive=True
+        )
```
jayantsing-db: same as above

src/databricks/sql/backend/sea/backend.py
```diff
@@ -627,9 +632,22 @@ def get_catalogs(
         max_rows: int,
         max_bytes: int,
         cursor: "Cursor",
-    ):
-        """Not implemented yet."""
-        raise NotImplementedError("get_catalogs is not yet implemented for SEA backend")
+    ) -> "ResultSet":
+        """Get available catalogs by executing 'SHOW CATALOGS'."""
+        result = self.execute_command(
+            operation="SHOW CATALOGS",
```
jayantsing-db: please add the SQL command as some constants in a module

jayantsing-db (changes requested):
Added some comments

src/databricks/sql/backend/filters.py
varun-edachali-dbx: You're right, I missed this. I'll remove it.

src/databricks/sql/backend/filters.py
varun-edachali-dbx: Introducing a filter method that is utilised in the fetch phase would lead to a lot of specialised code for the table queries during the fetch phase. 

Currently, all that the execution relevant methods (`execute_command` and the metadata methods like `get_tables`, `get_schemas`, etc.) do is return a `ResultSet` that is set as the active result set of the `Cursor`. 

The fetch phase from this step on is completely invariant of the kind of query that took place. If we want to use a separate `filter` method, then we would have to add custom logic during the fetch (`if table metadata then filter result; return result;`). Maintaining the generality of the fetch phase is likely worth the tradeoff involved in creating a new copy.

src/databricks/sql/backend/filters.py
varun-edachali-dbx: It is not present in the base class, it is an instance of a `ResultData` model which represents the `results` returned during SEA execution. In our case, we set the filtered rows in the data array of this `ResultData` to effectively create a filtered `SeaResultSet`.

src/databricks/sql/backend/filters.py
varun-edachali-dbx: I think we need some service specific methods at some point during the filtering process to know what kind of result set to return, since our concrete instances are service specific. I tried to keep the root methods invoked (`filter by table type`) general, following which they invoke the service specific builders based on the type of the instance passed to them.

src/databricks/sql/backend/sea/backend.py
```diff
@@ -627,9 +628,22 @@ def get_catalogs(
...
-        """Not implemented yet."""
-        raise NotImplementedError("get_catalogs is not yet implemented for SEA backend")
+    ) -> "ResultSet":
+        """Get available catalogs by executing 'SHOW CATALOGS'."""
+        result = self.execute_command(
+            operation=MetadataCommands.SHOW_CATALOGS.value,
+            session_id=session_id,
+            max_rows=max_rows,
+            max_bytes=max_bytes,
+            lz4_compression=False,
```
jayantsing-db, line 640: not using compression for metadata?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -627,9 +628,22 @@ def get_catalogs(
...
+            cursor=cursor,
+            use_cloud_fetch=False,
+            parameters=[],
+            async_op=False,
+            enforce_embedded_schema_correctness=False,
```
jayantsing-db, line 645: this is a thrift-specific param?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -653,9 +688,48 @@ def get_tables(
...
+        )
+        assert result is not None, "execute_command returned None in synchronous mode"
+
+        from databricks.sql.result_set import SeaResultSet
+
+        assert isinstance(
+            result, SeaResultSet
+        ), "execute_command returned a non-SeaResultSet"
```
jayantsing-db: this assertion seems misplaced here. i think this is relevant to filter method and should be moved there

src/databricks/sql/backend/sea/backend.py
```diff
@@ -667,6 +741,33 @@ def get_columns(
         schema_name: Optional[str] = None,
         table_name: Optional[str] = None,
         column_name: Optional[str] = None,
-    ):
-        """Not implemented yet."""
-        raise NotImplementedError("get_columns is not yet implemented for SEA backend")
+    ) -> "ResultSet":
+        """Get columns by executing 'SHOW COLUMNS IN CATALOG catalog [SCHEMA LIKE pattern] [TABLE LIKE pattern] [LIKE pattern]'."""
+        if not catalog_name:
+            raise ValueError("Catalog name is required for get_columns")
```
jayantsing-db: so for the caller (client) code, it will appear as ValueError. is it okay? should we throw something else as per spec? how does the other backend throws error in first-class APIs like these?

src/databricks/sql/backend/sea/utils/filters.py
```diff
@@ -0,0 +1,153 @@
...
+    List,
+    Optional,
+    Any,
+    Callable,
+    cast,
+    TYPE_CHECKING,
+)
+
+if TYPE_CHECKING:
+    from databricks.sql.result_set import SeaResultSet
```
jayantsing-db, line 20: could you please remind me of the purpose?

src/databricks/sql/backend/sea/utils/filters.py
```diff
@@ -0,0 +1,153 @@
...
+            has_been_closed_server_side=result_set.has_been_closed_server_side,
+            lz4_compressed=result_set.lz4_compressed,
+            arrow_schema_bytes=result_set._arrow_schema_bytes,
+            is_staging_operation=False,
+        )
+
+        # Create a new ResultData object with filtered data
+        from databricks.sql.backend.sea.models.base import ResultData
+
+        result_data = ResultData(data=filtered_rows, external_links=None)
```
jayantsing-db, line 70: so when you manually set the result data in result set, what will happen in the fetch phase? will it use the execute-response to re-fetch the data? are there any other examples in the codebase, where we manually set the result data or is this the first instance?

src/databricks/sql/backend/sea/utils/filters.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Convert to uppercase for case-insensitive comparison if needed
+        if not case_sensitive:
+            allowed_values = [v.upper() for v in allowed_values]
+
+        return ResultSetFilter._filter_sea_result_set(
+            result_set,
+            lambda row: (
+                len(row) > column_index
+                and isinstance(row[column_index], str)
```
jayantsing-db: is it possible to have column values as non-str here?

jayantsing-db (approved):
LGTM. Added some questions.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: This is a side effect of setting `use_cloud_fetch=False`: compression is not supported for `INLINE` + `JSON` in SEA.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Yes, but it is a param passed to `Cursor`'s `execute` method, so I don't see a way to not include it in our `execute_command`, if it were passed as a connection level property then we may have been able to access it that way. Should I try to avoid passing it to the SEA backend?

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Then I would have to allow an arbitrary `ResultSet` to be passed to the top level filter methods, which counters our reasoning to move it over to the SEA specific utils, right? 

I just thought of a way to remove the assertion entirely, though: I changed the backend SEA methods to return a `SeaResultSet` specifically, instead of a `ResultSet`. This allows for more constrained typing while satisfying the interface. I think this is a cleaner solution.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Thanks for the catch, I changed it to raise `DatabaseError` which seems the most appropriate, I also changed a few other methods to raise `ProgrammingError` instead of `ValueError`.

src/databricks/sql/backend/sea/utils/filters.py
varun-edachali-dbx: `TYPE_CHECKING` tells the type checker that the given dependencies are not required at run-time, just for type checking. If we don't place it in this block, then this module will be imported at run-time, and fail to work because of circular imports. 

Usually, along with using this `TYPE_CHECKING` block, we have to place the types within double quotes (eg: "SeaResultSet"), but using `from __future__ import annotations` allows us to avoid doing this.

src/databricks/sql/backend/sea/utils/filters.py
varun-edachali-dbx: Since we construct the `ResultData` by setting the `data` field and not `external_links`, the `SeaResultSet` is effectively instantiated as it would be during `INLINE` + `JSON_ARRAY` querying. 

We also set it during the instantiation of the `ResultSet` in the SEA backend, as seen [here](https://github.com/databricks/databricks-sql-python/blob/a4d5bdb726aee53bfa27b60e1b7baf78c01a67d3/src/databricks/sql/backend/sea/backend.py#L619).

src/databricks/sql/backend/sea/utils/filters.py
varun-edachali-dbx: You're right, this is an un-necessary check, I removed it.

PR #594, opened 2025-06-11
Complete Fetch Phase (for `INLINE` disposition and `JSON_ARRAY` format)

varun-edachali-dbx:
## What type of PR is this?
- [x] Feature

## Description
Complete the Fetch phase of the SEA implementation for `INLINE` disposition and `JSON_ARRAY` format. Introduce a new `JsonQueue`. 

Note that this is not yet feature complete, with some known failures in the following domains (for the SEA backend only): 
- retries
- native, complex params
- metadata queries 

These will be resolved in future PRs. 

## How is this tested?
- [x] Unit tests
- [ ] E2E Tests
- [x] Manually - using the test scripts invoked in `examples/experimental/sea_connector_test.py`. 
- [ ] N/A

The coverage of the key classes by the unit tests are as below: 
| Module | Statements | Missing | Coverage | Notes |
| `utils.py` (JsonQueue class) | 15 | 0 | 100% | Completely covered |
| `utils.py` (SeaResultSetQueueFactory class) | 20 | 0 | 100% | Completely covered |
| `result_set.py` (SeaResultSet class) | 155 | 0 | 100% | Completely covered |
| `conversion.py` | 43 | 0 | 100% | Completely covered |

[PECOBLR-484](https://databricks.atlassian.net/browse/PECOBLR-484?atlOrigin=eyJpIjoiZjJlMTg1MDg1MDA4NGMwZDkxODE1MDAwYWQzNjk1YWYiLCJwIjoiaiJ9)

[PECOBLR-484]: https://databricks.atlassian.net/browse/PECOBLR-484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/databricks/sql/backend/sea/conversion.py
```diff
@@ -0,0 +1,135 @@
+"""
+Type conversion utilities for the Databricks SQL Connector.
```
jayantsing-db: if it is an util, can this be moved to a common util module for SEA?

src/databricks/sql/backend/sea/conversion.py
```diff
@@ -0,0 +1,135 @@
...
+    Based on the types supported by the Databricks SDK.
+    """
+
+    # SQL type to conversion function mapping
+    # TODO: complex types
+    TYPE_MAPPING: Dict[str, Callable] = {
+        # Numeric types
+        SqlType.BYTE: lambda v: int(v),
+        SqlType.SHORT: lambda v: int(v),
+        SqlType.INT: lambda v: int(v),
```
jayantsing-db: can we mention the source of truth for this mapping in docstring/comment?

src/databricks/sql/backend/sea/conversion.py
```diff
@@ -0,0 +1,135 @@
...
+        SqlType.LONG: lambda v: int(v),
+        SqlType.FLOAT: lambda v: float(v),
+        SqlType.DOUBLE: lambda v: float(v),
+        SqlType.DECIMAL: lambda v, p=None, s=None: (
+            decimal.Decimal(v).quantize(
+                decimal.Decimal(f'0.{"0" * s}'), context=decimal.Context(prec=p)
```
jayantsing-db: please add comments and any contract/SOT

src/databricks/sql/backend/sea/conversion.py
```diff
@@ -0,0 +1,135 @@
...
+        SqlType.BINARY: lambda v: bytes.fromhex(v),
+        # Other types
+        SqlType.NULL: lambda v: None,
+        # Complex types and user-defined types return as-is
+        SqlType.USER_DEFINED_TYPE: lambda v: v,
+    }
+
+    @staticmethod
+    def convert_value(
+        value: Any,
```
jayantsing-db: should this be str?

src/databricks/sql/backend/sea/conversion.py
```diff
@@ -0,0 +1,135 @@
...
+        sql_type: str,
+        precision: Optional[int] = None,
+        scale: Optional[int] = None,
```
jayantsing-db: these params are only valid for certain types? can we have kwargs instead?

src/databricks/sql/backend/sea/conversion.py
```diff
@@ -0,0 +1,135 @@
...
+    ) -> Any:
```
jayantsing-db: can this be a better type like object?

src/databricks/sql/result_set.py
```diff
@@ -97,12 +142,6 @@ def is_staging_operation(self) -> bool:
         """Whether this result set represents a staging operation."""
         return self._is_staging_operation
 
-    # Define abstract methods that concrete implementations must implement
```
jayantsing-db, line 100: not needed anymore?

src/databricks/sql/result_set.py
```diff
@@ -451,13 +452,13 @@ class SeaResultSet(ResultSet):
...
     def __init__(
         self,
-        connection: "Connection",
-        execute_response: "ExecuteResponse",
-        sea_client: "SeaDatabricksClient",
+        connection: Connection,
+        execute_response: ExecuteResponse,
+        sea_client: SeaDatabricksClient,
+        result_data: ResultData,
+        manifest: Optional[ResultManifest] = None,
```
jayantsing-db: why is manifest optional? we always have a manifest and if not the query should fail

src/databricks/sql/result_set.py
```diff
@@ -468,10 +469,21 @@ def __init__(
...
-            result_data: Result data from SEA response (optional)
-            manifest: Manifest from SEA response (optional)
+            result_data: Result data from SEA response
+            manifest: Manifest from SEA response
         """
 
+        results_queue = SeaResultSetQueueFactory.build_queue(
+            result_data,
+            manifest,
+            str(execute_response.command_id.to_sea_statement_id()),
```
jayantsing-db: nit: why is str passed specifically?

src/databricks/sql/result_set.py
```diff
@@ -468,10 +469,21 @@ def __init__(
...
+            description=execute_response.description,
```
jayantsing-db: what is the use of description in queue? i thought it was used by result set apis like fetch etc.?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
             command_id=execute_response.command_id,
             status=execute_response.status,
             has_been_closed_server_side=execute_response.has_been_closed_server_side,
+            results_queue=results_queue,
             description=execute_response.description,
             is_staging_operation=execute_response.is_staging_operation,
             lz4_compressed=execute_response.lz4_compressed,
-            arrow_schema_bytes=execute_response.arrow_schema_bytes,
+            arrow_schema_bytes=execute_response.arrow_schema_bytes or b"",
```
jayantsing-db: what is the use of arrow schema bytes? could you please remind me where is this coming for SEA and thrift? if i recall correctly, thrift uses result set metadata response for metadata and SEA uses manifest.

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+        """
+        Convert raw data rows to Arrow table.
+        """
+        if not rows:
+            return pyarrow.Table.from_pydict({})
+
+        columns = []
+        num_cols = len(rows[0])
+        for i in range(num_cols):
+            columns.append([row[i] for row in rows])
```
jayantsing-db: will this O(n^2) complexity okay for max 22 MB data?

src/databricks/sql/result_set.py
```diff
@@ -468,10 +469,21 @@ def __init__(
             sea_client: The SeaDatabricksClient instance for direct access
             buffer_size_bytes: Buffer size for fetching results
             arraysize: Default number of rows to fetch
-            result_data: Result data from SEA response (optional)
-            manifest: Manifest from SEA response (optional)
+            result_data: Result data from SEA response
+            manifest: Manifest from SEA response
         """
 
+        results_queue = SeaResultSetQueueFactory.build_queue(
```
jayantsing-db: in the case of inline, what role does the queue play? i think you don't have to do any fetch

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
-            arrow_schema_bytes=execute_response.arrow_schema_bytes,
+            arrow_schema_bytes=execute_response.arrow_schema_bytes or b"",
         )
 
-    def _fill_results_buffer(self):
-        """Fill the results buffer from the backend."""
-        raise NotImplementedError(
-            "_fill_results_buffer is not implemented for SEA backend"
-        )
+    def _convert_json_to_arrow(self, rows: List) -> "pyarrow.Table":
```
jayantsing-db: can the type be more strict like List<List<str>>?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+        columns = []
+        num_cols = len(rows[0])
+        for i in range(num_cols):
+            columns.append([row[i] for row in rows])
+        names = [col[0] for col in self.description]
+        return pyarrow.Table.from_arrays(columns, names=names)
+
+    def _convert_json_types(self, rows: List) -> List:
+        """
+        Convert raw data rows to Row objects with named columns based on description.
```
jayantsing-db: please elaborate on `Convert raw data rows to Row objects` and the need?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+                try:
+                    converted_value = SqlTypeConverter.convert_value(
+                        value, column_type, precision=precision, scale=scale
+                    )
+                    converted_row.append(converted_value)
+                except Exception as e:
+                    logger.warning(
+                        f"Error converting value '{value}' to {column_type}: {e}"
+                    )
+                    converted_row.append(value)
```
jayantsing-db: we seem to be swallowing the exception and doing nothing. is it okay? will it have any unintended side-effects?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+            List of Row objects with named columns and converted values
+        """
+
+        if not self.description or not rows:
+            return rows
+
+        ResultRow = Row(*[col[0] for col in self.description])
+        rows = self._convert_json_types(rows)
+
+        return [ResultRow(*row) for row in rows]
```
jayantsing-db: what is the need for this?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+
+    def fetchall_json(self) -> List:
+        """
+        Fetch all remaining rows as a columnar table.
+
+        Returns:
+            Columnar table containing all remaining rows
+        """
+
+        results = self.results.remaining_rows()
```
jayantsing-db: what is self.results? queue?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+            raise ValueError(f"size argument for fetchmany is {size} but must be >= 0")
+
+        if not isinstance(self.results, JsonQueue):
+            raise NotImplementedError("fetchmany_arrow only supported for JSON data")
+
+        rows = self._convert_json_types(self.results.next_n_rows(size))
+        results = self._convert_json_to_arrow(rows)
+        self._next_row_index += results.num_rows
+
+        return results
```
jayantsing-db: this looks clean. thanks

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+            ValueError: If size is negative
+        """
+
+        if size < 0:
+            raise ValueError(f"size argument for fetchmany is {size} but must be >= 0")
+
+        if not isinstance(self.results, JsonQueue):
+            raise NotImplementedError("fetchmany_arrow only supported for JSON data")
+
+        rows = self._convert_json_types(self.results.next_n_rows(size))
```
jayantsing-db: i am a bit worried about repetitive processing of the table data. could you please analyse that for this?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+
+        return results
+
+    def fetchall_arrow(self) -> "pyarrow.Table":
+        """
+        Fetch all remaining rows as an Arrow table.
+        """
+
+        if not isinstance(self.results, JsonQueue):
+            raise NotImplementedError("fetchall_arrow only supported for JSON data")
```
jayantsing-db: i believe this will be removed post cloud fetch implementation?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+
+        return results
 
     def fetchone(self) -> Optional[Row]:
         """
         Fetch the next row of a query result set, returning a single sequence,
         or None when no more data is available.
+
+        Returns:
+            A single Row object or None if no more rows are available
```
jayantsing-db: is the Row object relevant for arrow? this is JSON APIs specific?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+
+        if not self.description or not rows:
+            return rows
+
+        ResultRow = Row(*[col[0] for col in self.description])
+        rows = self._convert_json_types(rows)
+
+        return [ResultRow(*row) for row in rows]
+
+    def fetchmany_json(self, size: int) -> List:
```
jayantsing-db: why is this called _json when the return type is python type?

src/databricks/sql/result_set.py
jayantsing-db: you have APIs which read like fetch_json, etc whose return type is List. And then you have this method too where return type is again List but it is supposed to have Row objects. However, the doc strings of fetch_json seem to create a distinction between json, Row, Arrow. This seems ambiguous

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+
+        if size < 0:
+            raise ValueError(f"size argument for fetchmany is {size} but must be >= 0")
+
+        results = self.results.next_n_rows(size)
+        self._next_row_index += len(results)
+
+        return results
+
+    def fetchall_json(self) -> List:
```
jayantsing-db: why is this called _json?

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
             raise AssertionError("Row set type is not valid")
 
 
+class SeaResultSetQueueFactory(ABC):
```
jayantsing-db: i think having SEA code encapsulated in one package would be better?

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+    @staticmethod
+    def build_queue(
+        sea_result_data: ResultData,
+        manifest: Optional[ResultManifest],
```
jayantsing-db: why is manifest optional?

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+        statement_id: str,
+        description: Optional[List[Tuple[Any, ...]]] = None,
```
jayantsing-db: what is the use of description in the queue? and why optional?

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+        schema_bytes: Optional[bytes] = None,
```
jayantsing-db: what is the use of arrow schema here? I think this is for cloud-fetch arrow?

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+
+        if sea_result_data.data is not None:
+            # INLINE disposition with JSON_ARRAY format
+            return JsonQueue(sea_result_data.data)
+        elif sea_result_data.external_links is not None:
+            # EXTERNAL_LINKS disposition
+            raise NotImplementedError(
+                "EXTERNAL_LINKS disposition is not implemented for SEA backend"
+            )
+        return JsonQueue([])
```
jayantsing-db: i think better to throw an error here otherwise could go unnoticed

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+        length = min(num_rows, self.n_valid_rows - self.cur_row_index)
+        slice = self.data_array[self.cur_row_index : self.cur_row_index + length]
+        self.cur_row_index += length
+        return slice
+
+    def remaining_rows(self):
+        """Get all remaining rows from the data array."""
+        slice = self.data_array[self.cur_row_index :]
+        self.cur_row_index += len(slice)
+        return slice
```
jayantsing-db: nice

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+
+
+class JsonQueue(ResultSetQueue):
+    """Queue implementation for JSON_ARRAY format data."""
+
+    def __init__(self, data_array):
+        """Initialize with JSON array data."""
+        self.data_array = data_array
+        self.cur_row_index = 0
+        self.n_valid_rows = len(data_array)
```
jayantsing-db: why is this called n_valid_rows?

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+
+    def next_n_rows(self, num_rows):
+        """Get the next n rows from the data array."""
+        length = min(num_rows, self.n_valid_rows - self.cur_row_index)
```
jayantsing-db: is this correct? the index is 0 based and the n_valid_rows is 1 based

if not correct, how are unit tests passing?

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+        slice = self.data_array[self.cur_row_index : self.cur_row_index + length]
```
jayantsing-db: for each such op, the complexity is of the order of size of the table? can we do better?

src/databricks/sql/result_set.py
```diff
@@ -480,46 +492,201 @@ def __init__(
...
+        """
+        Fetch all remaining rows as a columnar table.
+
+        Returns:
+            Columnar table containing all remaining rows
+        """
+
+        results = self.results.remaining_rows()
+        self._next_row_index += len(results)
```
jayantsing-db: what is the use of this _next_row_index? right now it doesn't seem to have a purpose. the queue seems to have its own row index

src/databricks/sql/utils.py
```diff
@@ -107,6 +111,70 @@ def build_queue(
...
+        return JsonQueue([])
+
+
+class JsonQueue(ResultSetQueue):
+    """Queue implementation for JSON_ARRAY format data."""
+
+    def __init__(self, data_array):
+        """Initialize with JSON array data."""
+        self.data_array = data_array
+        self.cur_row_index = 0
```
jayantsing-db: should there be some assertions/logging using the cur_row_index? for example, when this reaches end of table and there is a read operation on the queue? Does result set already handle that?

jayantsing-db (changes requested):
Thanks. Added some comments.

src/databricks/sql/backend/sea/conversion.py
varun-edachali-dbx: There is no real source of truth for the mapping to the Python types that I referred to. However, I linked the source of the various types we have to account for in the `SqlType` docstring. 

Note that I did not include any complex types in the `TYPE_MAPPING` data structure since we're not accounting for them at the moment.

src/databricks/sql/backend/sea/conversion.py
varun-edachali-dbx: You're right, that's much cleaner.

src/databricks/sql/result_set.py
varun-edachali-dbx: Yes, I realised that `_fill_results_buffer` is used to re-initialise a `Queue` after it is consumed if `has_more_rows` is `True`, it is only relevant to Thrift and not SEA so it need not be defined in the interface.

src/databricks/sql/result_set.py
varun-edachali-dbx: Initially, I was passing it because the return type of `to_sea_statement_id` was `Any` because it was returning the raw `guid` which could be of different types depending on the backend that set it. I now realise that it is cleaner to make `to_sea_statement_id` return a `str` by validating this in the `CommandId` class. I have made this change.

src/databricks/sql/result_set.py
varun-edachali-dbx: The description is used in the [convert_decimals_in_arrow_table](https://github.com/databricks/databricks-sql-python/blob/76963164ef3c0da091de9215d130521fde852812/src/databricks/sql/utils.py#L624) method, which is used in the creation of an arrow table from a downloaded arrow file, which will be of use in the SEA cloud fetch implementation. 

It is also used by the result set APIs you mentioned.

src/databricks/sql/result_set.py
varun-edachali-dbx: You're right, SEA doesn't use arrow schema bytes and currently it is hardcoded as `None` in the `ExecuteResponse` instance created by the SEA backend. We should eventually aim to remove this service specific state from `ExecuteResponse` (`arrow_schema_bytes` and `has_been_closed_server_side`). I will do that as part of a later cleanup. 

https://databricks.atlassian.net/browse/PECOBLR-603?atlOrigin=eyJpIjoiNzg5ZTA5MjRjMWE5NDMxZWEwYzZhNzZjNjhlNTY0ZTkiLCJwIjoiaiJ9

src/databricks/sql/result_set.py
varun-edachali-dbx: That docstring was incorrect because I had extracted the method from another one. I've updated the docstring.

varun-edachali-dbx: Yes, this docstring was incorrect, I've updated it.

src/databricks/sql/utils.py
varun-edachali-dbx: The use has been discussed [here](https://github.com/databricks/databricks-sql-python/pull/594/files?w=1#r2176268498). 

It is defined as Optional for the Thrift backend, but I'm not sure why either. I am changing it to be mandatory. Note that this is changing the Thrift code paths slightly.

src/databricks/sql/result_set.py
varun-edachali-dbx: All it helps us do is get the specified number of rows from the result without having to worry about the form of the result data. Defining a Queue for inline just lets us remain consistent with the all of the remaining fetch methods.

src/databricks/sql/result_set.py
varun-edachali-dbx: I don't see any un-intended side effects. 

In case a conversion fails, we retain the string representation of the value and continue. The conversion should not fail unless there is an issue with the metadata given to us, since our conversion methods are rather simple. This code is rather defensive, we can probably do without the try-catch, but I think this could be helpful during debugging until our conversion is validated completely.

src/databricks/sql/result_set.py
varun-edachali-dbx: Yes, that's right.

src/databricks/sql/result_set.py
varun-edachali-dbx: There is a `rownumber` property exposed by the `ResultSet` that returns the value of `_next_row_index` which can be invoked by the `Cursor`, so we should maintain it.

In Thrift, it is also used to denote an "expected row offset" which is used during the `fetch_results` call, but in SEA its only purpose is `rownumber`.

src/databricks/sql/result_set.py
varun-edachali-dbx: Yes, that's right. 

We will still require some specific processing for the `JsonQueue`, though, since we have to explicitly convert what it returns into an arrow table.

src/databricks/sql/result_set.py
varun-edachali-dbx: No, the `Row` object is [defined by us](https://github.com/databricks/databricks-sql-python/blob/76963164ef3c0da091de9215d130521fde852812/src/databricks/sql/types.py#L72). `List[Row]` is the return type for `fetchone()`, `fetchmany()`, and `fetchall()` regardless of backend.

src/databricks/sql/utils.py
varun-edachali-dbx: Arrow schema bytes is not actually relevant to SEA, as [stated here](https://github.com/databricks/databricks-sql-python/pull/594/files#r2176272982). I was not passing it as an argument to `build_queue` but it looks like I left it as a parameter, I will remove it.

src/databricks/sql/utils.py
varun-edachali-dbx: I was doing that initially but there are a few queries that do not return any data (neither data array nor external links get populated). For example, below is some data populated on running [test_create_table_will_return_empty_result_set](https://github.com/databricks/databricks-sql-python/blob/76963164ef3c0da091de9215d130521fde852812/tests/e2e/test_driver.py#L330) for SEA: 

Result Data: 
```
ResultData(data=None, external_links=None, byte_count=None, chunk_index=None, next_chunk_index=None, next_chunk_internal_link=None, row_count=None, row_offset=None, attachment=None)
```

Manifest: 
```
ResultManifest(format='JSON_ARRAY', schema={'column_count': 2, 'columns': [{'name': 'num_affected_rows', 'type_text': 'BIGINT', 'type_name': 'LONG', 'position': 0}, {'name': 'num_inserted_rows', 'type_text': 'BIGINT', 'type_name': 'LONG', 'position': 1}]}, total_row_count=0, total_byte_count=0, total_chunk_count=0, truncated=False, chunks=None, result_compression=None, is_volume_operation=False)
```

Notice that both data and external links are `None` in the result data. I am currently exploring the alternative of using the `format` key of the manifest instead of the result data to instantiate the Queue. Should I include that in the current PR?

src/databricks/sql/utils.py
varun-edachali-dbx: Looks correct to me. If we request `x > n` rows and we have not done any retrievals yet, then we want `n - 0 = n` rows in the current call.

src/databricks/sql/utils.py
varun-edachali-dbx: There are no such assertions made in the current version of the codebase (i.e., across Thrift and SEA). Should I add that? 

Currently, if we reach the end of the table and do a read operation, an empty list is returned.

src/databricks/sql/result_set.py
varun-edachali-dbx: It is a helper method to convert the raw output from the Queue into `List[Row]` as expected by the `Cursor` that invokes it. This structure is consistent with the `ThriftResultSet`.

src/databricks/sql/result_set.py
varun-edachali-dbx: This is a helper method that invokes the `JsonQueue`. It returns the raw output from the `JsonQueue`, which is just a list of rows (each of which is a list of string values). I have updated the type to be the stronger `List[List[str]]` for clarity.

src/databricks/sql/utils.py
jayantsing-db: please update the discussed approach here.

src/databricks/sql/utils.py
jayantsing-db: thanks. fine to do this separately.

src/databricks/sql/backend/sea/backend.py
```diff
@@ -251,7 +251,7 @@ def close_session(self, session_id: SessionId) -> None:
         logger.debug("SeaDatabricksClient.close_session(session_id=%s)", session_id)
 
         if session_id.backend_type != BackendType.SEA:
-            raise ProgrammingError("Not a valid SEA session ID")
+            raise ValueError("Not a valid SEA session ID")
```
jayantsing-db, line 254: i think we should convert these to spec errors as well separately. let's log some ticket to track this

src/databricks/sql/backend/sea/backend.py
```diff
@@ -429,7 +426,7 @@ def execute_command(
         """
 
         if session_id.backend_type != BackendType.SEA:
-            raise ProgrammingError("Not a valid SEA session ID")
+            raise ValueError("Not a valid SEA session ID")
```
jayantsing-db, line 429: same for all value errors.

src/databricks/sql/backend/sea/queue.py
```diff
@@ -0,0 +1,71 @@
...
+
+        if manifest.format == ResultFormat.JSON_ARRAY.value:
+            # INLINE disposition with JSON_ARRAY format
+            return JsonQueue(sea_result_data.data)
+        elif manifest.format == ResultFormat.ARROW_STREAM.value:
+            # EXTERNAL_LINKS disposition
+            raise NotImplementedError(
+                "EXTERNAL_LINKS disposition is not implemented for SEA backend"
+            )
+        raise ProgrammingError("Invalid result format")
```
jayantsing-db, line 48: should this be a spec error? is ProgrammingError a custom error?

src/databricks/sql/backend/sea/result_set.py
```diff
@@ -0,0 +1,266 @@
...
+from databricks.sql.backend.sea.models.base import ResultData, ResultManifest
+from databricks.sql.backend.sea.utils.conversion import SqlTypeConverter
+
+try:
+    import pyarrow
+except ImportError:
+    pyarrow = None
+
+if TYPE_CHECKING:
+    from databricks.sql.client import Connection
```
jayantsing-db, line 17: could you please remind me of the need for this?

src/databricks/sql/backend/sea/result_set.py
```diff
@@ -0,0 +1,266 @@
...
+            status=execute_response.status,
+            has_been_closed_server_side=execute_response.has_been_closed_server_side,
+            results_queue=results_queue,
+            description=execute_response.description,
+            is_staging_operation=execute_response.is_staging_operation,
+            lz4_compressed=execute_response.lz4_compressed,
+            arrow_schema_bytes=execute_response.arrow_schema_bytes,
+        )
+
+    def _convert_json_types(self, row: List) -> List:
```
jayantsing-db: can the input type be `List<str>` or something more specific? Same for output type.

src/databricks/sql/backend/sea/result_set.py
```diff
@@ -0,0 +1,266 @@
...
+                converted_row.append(converted_value)
+            except Exception as e:
+                logger.warning(
+                    f"Error converting value '{value}' to {column_type}: {e}"
+                )
+                converted_row.append(value)
+
+        return converted_row
+
+    def _convert_json_to_arrow_table(self, rows: List[List]) -> "pyarrow.Table":
```
jayantsing-db: is the input type `List<List<str>>`?

src/databricks/sql/backend/sea/result_set.py
```diff
@@ -0,0 +1,266 @@
...
+            return pyarrow.Table.from_pydict({})
+
+        # create a generator for row conversion
+        converted_rows_iter = (self._convert_json_types(row) for row in rows)
+        cols = list(map(list, zip(*converted_rows_iter)))
+
+        names = [col[0] for col in self.description]
+        return pyarrow.Table.from_arrays(cols, names=names)
+
+    def _create_json_table(self, rows: List[List[str]]) -> List[Row]:
```
jayantsing-db, line 133: types are helpful. thanks!

src/databricks/sql/backend/sea/result_set.py
```diff
@@ -0,0 +1,266 @@
...
+        Raises:
+            ImportError: If PyArrow is not installed
+            ValueError: If size is negative
+        """
+
+        if size < 0:
+            raise ValueError(f"size argument for fetchmany is {size} but must be >= 0")
+
+        if not isinstance(self.results, JsonQueue):
+            raise NotImplementedError("fetchmany_arrow only supported for JSON data")
```
jayantsing-db, line 200: thanks these all are much more clear.

src/databricks/sql/utils.py
```diff
@@ -13,12 +13,16 @@
 
 import lz4.frame
 
+from databricks.sql.backend.sea.backend import SeaDatabricksClient
```
jayantsing-db, line 16: this could be in separate PR. But i think better to have sea package above backend, i.e. databricks.sql.sea and databricks.sql.backend.sea?

jayantsing-db (approved):
Some minor comments. Rest LGTM.

src/databricks/sql/backend/sea/queue.py
varun-edachali-dbx: `ProgrammingError` is a spec error

src/databricks/sql/backend/sea/result_set.py
varun-edachali-dbx: The `TYPE_CHECKING` block allows us to make circular _references_, we're basically stating that the imports within that block are only required for type checking, and not during run-time. Usually, this is coupled with placing this reference in quotes, but we can avoid doing this by using `from __future__ import annotations`. 

A reference: https://vickiboykis.com/2023/12/11/why-if-type_checking/

src/databricks/sql/utils.py
varun-edachali-dbx: https://databricks.atlassian.net/browse/PECOBLR-606?atlOrigin=eyJpIjoiMzc4MWIzODMxZmU2NDk3YTk4NjQ2YjEzMzA0ZTliMWIiLCJwIjoiaiJ9

PR #595, opened 2025-06-11
Introduce models for `SeaDatabricksClient`

varun-edachali-dbx:
## What type of PR is this?
- [x] Feature

## Description
Introduce models for the requests and responses for communication with the Statement Execution API. 

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [x] Manually - through the execution scripts on a complete prototype of the SEA client. 
- [ ] N/A


src/databricks/sql/backend/sea/models/base.py
```diff
@@ -0,0 +1,87 @@
...
+    """Manifest information for a result set."""
+
+    format: str
+    schema: Dict[str, Any]  # Will contain column information
+    total_row_count: int
+    total_byte_count: int
+    total_chunk_count: int
+    truncated: bool = False
+    chunks: Optional[List[ChunkInfo]] = None
+    result_compression: Optional[str] = None
```
samikshya-db, line 94: have we not included all the fields from the SQL execution proto? 

Because I don't see `is_volume_operation`.

varun-edachali-dbx: Yes, you're right. I removed `is_volume_operation` in a recent commit because I was not using it in my implementations, but it should be there for model completeness. I'll add it back. Thanks!

samikshya-db (approved):
LGTM! Thanks for breaking the larger PR into smaller ones!

PR #598, opened 2025-06-16
Complete Fetch Phase (`EXTERNAL_LINKS` disposition and `ARROW` format)

varun-edachali-dbx:
…ionality

## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/src/databricks/sql/backend/sea/backend.py
+++ b/src/databricks/sql/backend/sea/backend.py
@@ -1,7 +1,8 @@
 import logging
+import uuid
 import time
 import re
-from typing import Dict, Tuple, List, Optional, Union, TYPE_CHECKING, Set
+from typing import Dict, Tuple, List, Optional, Any, Union, TYPE_CHECKING, Set
 
 from databricks.sql.backend.sea.utils.constants import (
     ALLOWED_SESSION_CONF_TO_DEFAULT_VALUES_MAP,
@@ -22,7 +23,9 @@ from databricks.sql.backend.types import (
 )
 from databricks.sql.exc import Error, NotSupportedError, ServerOperationError
 from databricks.sql.backend.sea.utils.http_client import SeaHttpClient
+from databricks.sql.thrift_api.TCLIService import ttypes
 from databricks.sql.types import SSLOptions
+from databricks.sql.utils import SeaResultSetQueueFactory
 from databricks.sql.backend.sea.models.base import (
     ResultData,
     ExternalLink,
@@ -301,6 +304,28 @@ class SeaDatabricksClient(DatabricksClient):
             List of allowed session configuration parameter names
         """
         return list(ALLOWED_SESSION_CONF_TO_DEFAULT_VALUES_MAP.keys())
+
+    def get_chunk_links(
+        self, statement_id: str, chunk_index: int
+    ) -> "GetChunksResponse":
+        """
+        Get links for chunks starting from the specified index.
+
+        Args:
+            statement_id: The statement ID
+            chunk_index: The starting chunk index
+
+        Returns:
+            GetChunksResponse: Response containing external links
+        """
+        from databricks.sql.backend.sea.models.responses import GetChunksResponse
+
+        response_data = self.http_client._make_request(
+            method="GET",
+            path=self.CHUNK_PATH_WITH_ID_AND_INDEX.format(statement_id, chunk_index),
+        )
+
+        return GetChunksResponse.from_dict(response_data)
 
     def _get_schema_bytes(self, sea_response) -> Optional[bytes]:
         """
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -1,8 +1,8 @@
-from __future__ import annotations
+from typing import Any, Dict, List, Optional, Union, TYPE_CHECKING
 
-from dateutil import parser
-import datetime
-import decimal
+if TYPE_CHECKING:
+    from databricks.sql.backend.sea.backend import SeaDatabricksClient
+
 from abc import ABC, abstractmethod
 from collections import OrderedDict, namedtuple
 from collections.abc import Iterable
@@ -10,11 +10,11 @@ from decimal import Decimal
 from enum import Enum
 from typing import Any, Dict, List, Optional, Tuple, Union
 import re
+import datetime
+import decimal
+from dateutil import parser
 
 import lz4.frame
-
-from databricks.sql.backend.sea.backend import SeaDatabricksClient
-from databricks.sql.backend.sea.models.base import ResultData, ResultManifest
 
 try:
     import pyarrow
@@ -29,8 +29,11 @@ from databricks.sql.thrift_api.TCLIService.ttypes import (
     TSparkRowSetType,
 )
 from databricks.sql.types import SSLOptions
-from databricks.sql.backend.types import CommandId
-
+from databricks.sql.backend.sea.models.base import (
+    ResultData,
+    ExternalLink,
+    ResultManifest,
+)
 from databricks.sql.parameters.native import ParameterStructure, TDbsqlParameter
 
 import logging
@@ -54,16 +57,16 @@ class ResultSetQueue(ABC):
 class ThriftResultSetQueueFactory(ABC):
     @staticmethod
     def build_queue(
-        row_set_type: TSparkRowSetType,
-        t_row_set: TRowSet,
-        arrow_schema_bytes: bytes,
-        max_download_threads: int,
-        ssl_options: SSLOptions,
+        row_set_type: Optional[TSparkRowSetType] = None,
+        t_row_set: Optional[TRowSet] = None,
+        arrow_schema_bytes: Optional[bytes] = None,
+        max_download_threads: Optional[int] = None,
+        ssl_options: Optional[SSLOptions] = None,
         lz4_compressed: bool = True,
-        description: Optional[List[Tuple]] = None,
+        description: Optional[List[Tuple[Any, ...]]] = None,
     ) -> ResultSetQueue:
         """
-        Factory method to build a result set queue.
+        Factory method to build a result set queue for Thrift backend.
 
         Args:
             row_set_type (enum): Row set type (Arrow, Column, or URL).
@@ -78,7 +81,11 @@ class ThriftResultSetQueueFactory(ABC):
             ResultSetQueue
         """
 
-        if row_set_type == TSparkRowSetType.ARROW_BASED_SET:
+        if (
+            row_set_type == TSparkRowSetType.ARROW_BASED_SET
+            and t_row_set is not None
+            and arrow_schema_bytes is not None
+        ):
             arrow_table, n_valid_rows = convert_arrow_based_set_to_arrow_table(
                 t_row_set.arrowBatches, lz4_compressed, arrow_schema_bytes
             )
@@ -86,7 +93,9 @@ class ThriftResultSetQueueFactory(ABC):
                 arrow_table, description
             )
             return ArrowQueue(converted_arrow_table, n_valid_rows)
-        elif row_set_type == TSparkRowSetType.COLUMN_BASED_SET:
+        elif (
+            row_set_type == TSparkRowSetType.COLUMN_BASED_SET and t_row_set is not None
+        ):
             column_table, column_names = convert_column_based_set_to_column_table(
                 t_row_set.columns, description
             )
@@ -96,8 +105,14 @@ class ThriftResultSetQueueFactory(ABC):
             )
 
             return ColumnQueue(ColumnTable(converted_column_table, column_names))
-        elif row_set_type == TSparkRowSetType.URL_BASED_SET:
-            return CloudFetchQueue(
+        elif (
+            row_set_type == TSparkRowSetType.URL_BASED_SET
+            and t_row_set is not None
+            and arrow_schema_bytes is not None
+            and max_download_threads is not None
+            and ssl_options is not None
+        ):
+            return ThriftCloudFetchQueue(
                 schema_bytes=arrow_schema_bytes,
                 start_row_offset=t_row_set.startRowOffset,
                 result_links=t_row_set.resultLinks,
@@ -140,14 +155,40 @@ class SeaResultSetQueueFactory(ABC):
         Returns:
             ResultSetQueue: The appropriate queue for the result data
         """
-
         if sea_result_data.data is not None:
             # INLINE disposition with JSON_ARRAY format
             return JsonQueue(sea_result_data.data)
         elif sea_result_data.external_links is not None:
             # EXTERNAL_LINKS disposition
-            raise NotImplementedError(
-                "EXTERNAL_LINKS disposition is not implemented for SEA backend"
+            if not schema_bytes:
+                raise ValueError(
+                    "Schema bytes are required for EXTERNAL_LINKS disposition"
+                )
+            if not max_download_threads:
+                raise ValueError(
+                    "Max download threads is required for EXTERNAL_LINKS disposition"
+                )
+            if not ssl_options:
+                raise ValueError(
+                    "SSL options are required for EXTERNAL_LINKS disposition"
+                )
+            if not sea_client:
+                raise ValueError(
+                    "SEA client is required for EXTERNAL_LINKS disposition"
+                )
+            if not manifest:
+                raise ValueError("Manifest is required for EXTERNAL_LINKS disposition")
+
+            return SeaCloudFetchQueue(
+                initial_links=sea_result_data.external_links,
+                schema_bytes=schema_bytes,
+                max_download_threads=max_download_threads,
+                ssl_options=ssl_options,
+                sea_client=sea_client,
+                statement_id=statement_id,
+                total_chunk_count=manifest.total_chunk_count,
+                lz4_compressed=lz4_compressed,
+                description=description,
             )
         else:
             # Empty result set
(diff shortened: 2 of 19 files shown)


src/databricks/sql/backend/databricks_client.py
```diff
@@ -22,6 +24,13 @@ class DatabricksClient(ABC):
     - Fetching metadata about catalogs, schemas, tables, and columns
     """
 
+    def __init__(self, ssl_options: SSLOptions, **kwargs):
```
jayantsing-db: these ssl options originate from Session?

src/databricks/sql/backend/databricks_client.py
```diff
@@ -22,6 +24,13 @@ class DatabricksClient(ABC):
...
+        self._use_arrow_native_complex_types = kwargs.get(
+            "_use_arrow_native_complex_types", True
+        )
+        self._max_download_threads = kwargs.get("max_download_threads", 10)
+        self._ssl_options = ssl_options
```
jayantsing-db: can we document the use of these and their presence in ABC? are all implementations supposed to respect these?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -342,7 +344,7 @@ def _results_message_to_execute_response(
 
         # Check for compression
         lz4_compressed = (
-            response.manifest.result_compression == ResultCompression.LZ4_FRAME
+            response.manifest.result_compression == ResultCompression.LZ4_FRAME.value
```
jayantsing-db, line 351: why .value? result_compression is a string val?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -631,6 +633,35 @@ def get_execution_result(
...
+        """
+        Get links for chunks starting from the specified index.
+        Args:
+            statement_id: The statement ID
+            chunk_index: The starting chunk index
+        Returns:
+            ExternalLink: External link for the chunk
+        """
+
+        response_data = self.http_client._make_request(
```
jayantsing-db: http_client should be a private member?

src/databricks/sql/backend/sea/models/responses.py
```diff
@@ -154,3 +154,38 @@ class CreateSessionResponse:
...
+@dataclass
+class GetChunksResponse:
+    """Response from getting chunks for a statement."""
+
+    statement_id: str
+    external_links: List[ExternalLink]
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "GetChunksResponse":
+        """Create a GetChunksResponse from a dictionary."""
```
jayantsing-db, line 179: can we link any public doc if available for this?

src/databricks/sql/backend/sea/queue.py
```diff
@@ -39,11 +59,18 @@ def build_queue(
...
         if manifest.format == ResultFormat.JSON_ARRAY.value:
             # INLINE disposition with JSON_ARRAY format
-            return JsonQueue(sea_result_data.data)
+            return JsonQueue(result_data.data)
         elif manifest.format == ResultFormat.ARROW_STREAM.value:
             # EXTERNAL_LINKS disposition
-            raise NotImplementedError(
-                "EXTERNAL_LINKS disposition is not implemented for SEA backend"
+            return SeaCloudFetchQueue(
+                initial_links=result_data.external_links or [],
```
jayantsing-db: is it possible to remove the notion of initial links? Maybe pass the result data object

src/databricks/sql/backend/sea/queue.py
```diff
@@ -69,3 +96,136 @@ def remaining_rows(self) -> List[List[str]]:
...
+                statement_id, total_chunk_count
+            )
+        )
+
+        initial_link = next((l for l in initial_links if l.chunk_index == 0), None)
+        if not initial_link:
+            return
+
+        self.download_manager = ResultFileDownloadManager(
+            links=[],
```
jayantsing-db: when no initial link, you seem to return at line 149 but here in the presence of links, why is link array empty?

src/databricks/sql/backend/sea/queue.py
```diff
@@ -69,3 +96,136 @@ def remaining_rows(self) -> List[List[str]]:
...
+        self._sea_client = sea_client
+        self._statement_id = statement_id
+
+        logger.debug(
+            "SeaCloudFetchQueue: Initialize CloudFetch loader for statement {}, total chunks: {}".format(
+                statement_id, total_chunk_count
+            )
+        )
+
+        initial_link = next((l for l in initial_links if l.chunk_index == 0), None)
```
jayantsing-db: can we make it simpler and remove the notion of initial link?

src/databricks/sql/backend/sea/queue.py
```diff
@@ -69,3 +96,136 @@ def remaining_rows(self) -> List[List[str]]:
...
+    def _progress_chunk_link(self):
+        """Progress to the next chunk link."""
+        if not self._current_chunk_link:
+            return None
+
+        next_chunk_index = self._current_chunk_link.next_chunk_index
+
+        if next_chunk_index is None:
+            self._current_chunk_link = None
+            return None
```
jayantsing-db: in case of partial/corrupt data, where are we throwing user-facing errors?

src/databricks/sql/backend/sea/queue.py
```diff
@@ -69,3 +96,136 @@ def remaining_rows(self) -> List[List[str]]:
...
+
+        logger.debug(
+            "SeaCloudFetchQueue: Initialize CloudFetch loader for statement {}, total chunks: {}".format(
+                statement_id, total_chunk_count
+            )
+        )
+
+        initial_link = next((l for l in initial_links if l.chunk_index == 0), None)
+        if not initial_link:
+            return
```
jayantsing-db: could you please elaborate on this scenario and the user-facing behaviour?

src/databricks/sql/backend/sea/queue.py
```diff
@@ -69,3 +96,136 @@ def remaining_rows(self) -> List[List[str]]:
...
+
+        self.download_manager = ResultFileDownloadManager(
```
jayantsing-db: why this when the parent class already has the same object?

src/databricks/sql/utils.py
```diff
@@ -201,70 +197,50 @@ def remaining_rows(self) -> "pyarrow.Table":
...
-            schema_bytes (bytes): Table schema in bytes.
-            max_download_threads (int): Maximum number of downloader thread pool threads.
-            start_row_offset (int): The offset of the first row of the cloud fetch links.
-            result_links (List[TSparkArrowResultLink]): Links containing the downloadable URL and metadata.
-            lz4_compressed (bool): Whether the files are lz4 compressed.
-            description (List[List[Any]]): Hive table schema description.
+        Args:
+            max_download_threads: Maximum number of download threads
+            ssl_options: SSL options for downloads
+            schema_bytes: Arrow schema bytes
```
jayantsing-db, line 227: could you please remind me the purpose of schema bytes vs description?

src/databricks/sql/utils.py
```diff
@@ -201,70 +197,50 @@ def remaining_rows(self) -> "pyarrow.Table":
         return slice
 
 
-class CloudFetchQueue(ResultSetQueue):
+class CloudFetchQueue(ResultSetQueue, ABC):
```
jayantsing-db, line 210: could you please explain the refactor to this reply?

varun-edachali-dbx: Old: `CloudFetchQueue` - Queue for Thrift Cloudfetch responses 
New:
- `CloudFetchQueue` - abstract class for general cloudfetch handling. Key responsibilities include: 
    - maintenance of the arrow table
    - general implementations for `next_n_rows` and `remaining_rows`
    - the creation of the arrow table at a given offset by using the common download manager. 
    - Each child class is now only responsible for defining a function to create the "next" table, i.e., creating and returning a table from the next link and updating the state as required. 
    
- `ThriftCloudFetchQueue(CloudFetchQueue)` - Queue for Thrift Cloudfetch responses. For the creation of the next table, there is not much complexity since we're already furnished with the next link from the initial execute response. We can invoke the `_create_table_at_offset` helper. 
- `SeaCloudFetchQueue(CloudFetchQueue)` - Queue for SEA Cloudfetch (i.e. External Links) responses. For the creation of the next table, we have to fetch the link corresponding to the next chunk index before invoking the `_create_table_at_offset` helper. 

Essentially, I broke up the old `CloudFetchQueue` into an abstract class and a concrete child `ThriftCloudFetchQueue`, so I could reduce repetition in the new `SeaCloudFetchQueue`.

I'll add this to the PR description as well.

src/databricks/sql/utils.py
```diff
@@ -309,21 +285,17 @@ def remaining_rows(self) -> "pyarrow.Table":
...
-        )
+    def _create_table_at_offset(self, offset: int) -> Union["pyarrow.Table", None]:
+        """Create next table at the given row offset"""
         # Create next table by retrieving the logical next downloaded file, or return None to signal end of queue
-        downloaded_file = self.download_manager.get_next_downloaded_file(
-            self.start_row_index
-        )
+        if not self.download_manager:
+            logger.debug("CloudFetchQueue: No download manager available")
+            return None
```
jayantsing-db: should this throw error?

src/databricks/sql/utils.py
```diff
@@ -338,19 +310,96 @@ def _create_next_table(self) -> Union["pyarrow.Table", None]:
...
+        )
+        arrow_table = self._create_table_at_offset(self.start_row_index)
+        if arrow_table:
+            self.start_row_index += arrow_table.num_rows
+            logger.debug(
+                "ThriftCloudFetchQueue: Found downloaded file, row count: {}, new start offset: {}".format(
+                    arrow_table.num_rows, self.start_row_index
+                )
+            )
+        return arrow_table
```
jayantsing-db, line 410: what is the current behaviour when arrow_table is null? does that throw error?

src/databricks/sql/session.py
```diff
@@ -64,7 +64,7 @@ def __init__(
         base_headers = [("User-Agent", self.useragent_header)]
         all_headers = (http_headers or []) + base_headers
 
-        self._ssl_options = SSLOptions(
+        self.ssl_options = SSLOptions(
```
jayantsing-db, line 67: why made public?

src/databricks/sql/backend/sea/queue.py
```diff
@@ -69,3 +96,136 @@ def remaining_rows(self) -> List[List[str]]:
...
+            return None
+
+        next_chunk_index = self._current_chunk_link.next_chunk_index
+
+        if next_chunk_index is None:
+            self._current_chunk_link = None
+            return None
+
+        try:
+            self._current_chunk_link = self._sea_client.get_chunk_link(
```
jayantsing-db: this isn't handling expiry? I think current thrift code doesn't handle expiry as well. What happens when the link is expired? What is the user-facing behaviour?

src/databricks/sql/backend/sea/queue.py
jayantsing-db: should this throw error as the only way to progress link is to first consume the previous link?

tests/e2e/common/large_queries_mixin.py
```diff
@@ -42,7 +44,14 @@ def fetch_rows(self, cursor, row_count, fetchmany_size):
             + "assuming 10K fetch size."
         )
 
-    def test_query_with_large_wide_result_set(self):
+    @pytest.mark.parametrize(
+        "extra_params",
+        [
+            {},
+            {"use_sea": True},
+        ],
```
jayantsing-db, line 52: what is this change? specifically the empty dict.

jayantsing-db (changes requested):
Few comments inline.

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: Yes, I've removed the parent constructor and pushed the instantiation of the attributes down to the child classes as discussed during our meeting.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Yes, result_compression is a string val

src/databricks/sql/backend/sea/queue.py
varun-edachali-dbx: Ideally, this should never happen. This code essentially just extracts the link corresponding to chunk index 0, which should always be provided to us. The check for it not existing is just for type safety. In case it does end up happening though, we prematurely return which should break subsequent operations.

src/databricks/sql/utils.py
varun-edachali-dbx: `None` is used to signal that no more arrow tables can be created. We return an empty pyarrow table from the `next_n_rows` and related functions as well in this case.

src/databricks/sql/session.py
varun-edachali-dbx: It is now accessed outside the `Session` class as well, in `SeaResultSetQueueFactory`'s `build_queue` method, so I thought I would make it explicitly public.

src/databricks/sql/backend/sea/queue.py
varun-edachali-dbx: Currently, if the link is expired, a simple `Error` is raised in `_validate_link` in `ResultSetDownloadHandler`. The user will see this error when `task.result()` is invoked on the `Future` when we try to access the downloaded file. 

Re-fetching of links on expiry for the SEA backend will be introduced in #635.

src/databricks/sql/backend/sea/queue.py
varun-edachali-dbx: I agree, I will change this to throw an error

src/databricks/sql/utils.py
varun-edachali-dbx: Both of them represent column metadata:
- `description` is [the 7-tuple](https://peps.python.org/pep-0249/#description) in human-readable form (this is also exposed to the user). Here, it is used to convert the decimals in the arrow table based on the scale and precision. 
- `schema_bytes` is the serialised arrow schema. It is returned by Thrift by the metadata query. It is used to augment the empty table returned by the `CloudFetchQueue` (for Thrift only).

src/databricks/sql/backend/sea/queue.py
varun-edachali-dbx: On second thought, this should not throw an error: there are some queries that do not return any data. The fact that the first link does not exist likely points to an empty response. In this case, we should simply return without worrying about any link fetching. I have reverted from throwing a `ServerOperationError` back to returning `None`. 

This was caught by the integration test `tests/e2e/test_driver.py::TestPySQLCoreSuite::test_create_table_will_return_empty_result_set[extra_params2]`

jayantsing-db: got it.

src/databricks/sql/backend/sea/queue.py
```diff
@@ -72,3 +100,112 @@ def remaining_rows(self) -> List[List[str]]:
...
+
+        return arrow_table
+
+    def _create_next_table(self) -> Union["pyarrow.Table", None]:
+        """Create next table by retrieving the logical next downloaded file."""
+        self._current_chunk_index += 1
+        next_chunk_link = self._get_chunk_link(self._current_chunk_index)
+        if not next_chunk_link:
+            return None
+        return self._create_table_from_link(next_chunk_link)
```
jayantsing-db, line 211: Is this multi-threaded? _create_table_from_link adds link one-by-one. _create_next_table calls _create_table_from_link. This seems to be serial execution

jayantsing-db: first you will consume arrow table from one link, then call create_next_table to get table from next link and so on.  link gets added to queue once the previous link is consumed. isn't this wasteful?

src/databricks/sql/utils.py
```diff
@@ -348,24 +322,94 @@ def _create_next_table(self) -> Union["pyarrow.Table", None]:
...
+            )
+        )
+        if self.result_links:
+            for result_link in self.result_links:
+                logger.debug(
+                    "- start row offset: {}, row count: {}".format(
+                        result_link.startRowOffset, result_link.rowCount
+                    )
+                )
+                self.download_manager.add_link(result_link)
```
jayantsing-db, line 391: this looks multi-threaded. We are adding all links of a fetch request to download manager.

src/databricks/sql/backend/sea/queue.py
jayantsing-db: will this be solved in future PRs where you introduce a separate link downloader?

varun-edachali-dbx: Yes, this will be solved in the future PR where I introduce a separate link downloader. There, I use the multithreading capabilities of the `ResultFileDownloadManager`. 

I intentionally kept this one quite simple and essentially single threaded so link expiry is not a concern, we download a link and create the arrow table the moment we fetch it. That way, this PR in isolation is correct.

src/databricks/sql/utils.py
varun-edachali-dbx: Yes, it is

src/databricks/sql/utils.py
```diff
@@ -348,24 +322,94 @@ def _create_next_table(self) -> Union["pyarrow.Table", None]:
...
+    @abstractmethod
+    def _create_next_table(self) -> Union["pyarrow.Table", None]:
+        """Create next table by retrieving the logical next downloaded file."""
+        pass
+
     def _create_empty_table(self) -> "pyarrow.Table":
-        # Create a 0-row table with just the schema bytes
+        """Create a 0-row table with just the schema bytes."""
+        if not self.schema_bytes:
+            return pyarrow.Table.from_pydict({})
```
jayantsing-db, line 336: why was this added to existing code? Is this just precautionary?

src/databricks/sql/utils.py
```diff
@@ -348,24 +322,94 @@ def _create_next_table(self) -> Union["pyarrow.Table", None]:
...
+        self.table = self._create_next_table()
+
+    def _create_next_table(self) -> Union["pyarrow.Table", None]:
+        logger.debug(
+            "ThriftCloudFetchQueue: Trying to get downloaded file for row {}".format(
+                self.start_row_index
+            )
+        )
+        arrow_table = self._create_table_at_offset(self.start_row_index)
+        if arrow_table:
```
jayantsing-db, line 403: why was this added to the existing code?

src/databricks/sql/utils.py
varun-edachali-dbx: Thrift provides us with arrow schema bytes, SEA does not. For the Thrift (empty) table, we can augment the column schema information into the pyarrow table, but for SEA, we return a completely empty table without this information. 

This code essentially augments the schema information if it is available (which is the case for Thrift, but not SEA).

src/databricks/sql/utils.py
varun-edachali-dbx: This was present in the previous code as well: the function returns `None`  to denote that no more arrow tables can be built (we have consumed the present data). In the earlier code, we returned at this point, and if the arrow table is non null, we fall through and end up incrementing the `start_row_index`. 

In the new flow the table creation is decoupled from the rest of the code (including the updation of the `start_row_index` state), so we perform an explicit check and only increment if `arrow_table` is not `None`.

src/databricks/sql/utils.py
jayantsing-db: Got it. For SEA we don't have easy way to put schema information from ResultManifest into an empty table?

src/databricks/sql/utils.py
jayantsing-db: I don't think we have this in existing code: https://github.com/databricks/databricks-sql-python/blob/ba1eab37909034563ea15bd9e28bbd6df29b1d11/src/databricks/sql/utils.py#L346

varun-edachali-dbx: A [little above the line you linked](https://github.com/databricks/databricks-sql-python/blob/ba1eab37909034563ea15bd9e28bbd6df29b1d11/src/databricks/sql/utils.py#L334) we notice that `if not downloaded_file: return None;`. Thus, if the table is "empty" (`None`) which denotes that there are no more arrow tables to create, we exit this flow early - this prevents us from reaching the call that you've linked above. 

In the new flow, the entire table creation is decoupled from this state updation, so we manually account for this flow by ensuring that `arrow_table` is not `None` before making the update.

src/databricks/sql/utils.py
varun-edachali-dbx: It is doable, but introduces some complexity - we would have to map each type given to us in the `description` (the 7 tuple containing name, type, etc.) to a corresponding pyarrow type (`pa.int64()`, `pa.string()`, etc.) and use that to define our pyarrow table schema, and then return an empty table. I do not see a way to do this without creating a manual mapping ("STRING" -> `pa.string()` etc.) so it could introduce some complexity and take some time. 

If this is worth doing, I'll create a ticket and begin working on it soon.

src/databricks/sql/utils.py
jayantsing-db: got it, thanks

PR #601, opened 2025-06-17
Add test to check thrift field IDs

vikrantpuppala:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Other

## Description
Add test to check thrift field IDs
## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

Checked that the test fails without the known exceptions added

### DIFF
--- a/tests/unit/test_thrift_field_ids.py
+++ b/tests/unit/test_thrift_field_ids.py
@@ -7,13 +7,17 @@ and ensure compatibility with various Thrift implementations and protocols.
 
 import inspect
 import pytest
-import unittest
 
 from databricks.sql.thrift_api.TCLIService import ttypes
 
 
-class TestThriftFieldIds(unittest.TestCase):
-    """Test suite for validating Thrift field ID constraints."""
+class TestThriftFieldIds:
+    """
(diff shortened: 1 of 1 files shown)


tests/unit/test_thrift_field_ids.py
```diff
@@ -0,0 +1,100 @@
+"""
+Unit test to validate that all Thrift-generated field IDs comply with the maximum limit.
```
jprakash-db: Can you move this to the test class docstring, in future we may add more tests for thrifts and having this as a file level docstring is not good

tests/unit/test_thrift_field_ids.py
```diff
@@ -0,0 +1,100 @@
...
+"""
+
+import inspect
+import pytest
+import unittest
+
+from databricks.sql.thrift_api.TCLIService import ttypes
+
+
+class TestThriftFieldIds(unittest.TestCase):
```
jprakash-db: The goal is to move away from `unittest`, plz refractor the test to use `pytest`

PR #604, opened 2025-06-18
Generalise Backend Layer 

varun-edachali-dbx:
## What type of PR is this?
- [x] Refactor

## Description
This PR refactors the codebase to aid the eventual introduction of a new backend client. It is composed of the following three PRs: 
- Separate the `Session` functionality from the `Connection` class to further abstract backend implementation details from the connection class: #571 
- Introduce a general backend interface (`DatabricksClient`): #573 
    - Introduce the `DatabricksClient` interface and make the existing `ThriftBackend` implement it as `ThriftDatabricksClient`. 
    - Introduce `SessionId` and `CommandId` as consistent adapters for backend layers to represent sessions and commands instead of relying on Thrift specific types. 
- Introduce a general `ResultSet` interface: #574 
    - Create a `ResultSet` interface and make the existing ResultSet interface implement it as `ThriftResultSet`. 
    - Make the concrete Thrift backend return a `ThriftResultSet` instead of `ExecuteResponse` from execution relevant commands. 
    - Generalise the return types of methods in the backend interface to prevent alignment with Thrift (introduce `CommandState` to prevent using Thrift's status types).   

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [ ] Manually
- [ ] N/A

[PECOBLR-569](https://databricks.atlassian.net/browse/PECOBLR-569?atlOrigin=eyJpIjoiZGJlZWVmMGIwZmJiNDM0ZDgwYTE0YTZiZGQ0MmVkM2EiLCJwIjoiaiJ9)

[PECOBLR-569]: https://databricks.atlassian.net/browse/PECOBLR-569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
+"""
+Abstract client interface for interacting with Databricks SQL services.
```
jprakash-db: Can you move the docstring within the class

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -857,51 +899,62 @@ def _wait_until_command_done(self, op_handle, initial_operation_status_resp):
...
+        lz4_compression: bool,
+        cursor: "Cursor",
         use_cloud_fetch=True,
         parameters=[],
         async_op=False,
         enforce_embedded_schema_correctness=False,
-    ):
-        assert session_handle is not None
+    ) -> Union["ResultSet", None]:
+        from databricks.sql.result_set import ThriftResultSet
```
jprakash-db: Have all the imports at the top of the file

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -938,58 +991,106 @@ def execute_command(
 
         if async_op:
             self._handle_execute_response_async(resp, cursor)
+            return None
```
jprakash-db, line 1032: Can we have a EmptyResultSet instead of none cc @jayantsing-db

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -938,58 +991,106 @@ def execute_command(
...
-    def get_catalogs(self, session_handle, max_rows, max_bytes, cursor):
-        assert session_handle is not None
+    def get_catalogs(
+        self,
+        session_id: SessionId,
+        max_rows: int,
+        max_bytes: int,
+        cursor: "Cursor",
+    ) -> "ResultSet":
+        from databricks.sql.result_set import ThriftResultSet
```
jprakash-db: Plz have imports at the top of the file

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -999,23 +1100,37 @@ def get_tables(
...
     def get_columns(
         self,
-        session_handle,
-        max_rows,
-        max_bytes,
-        cursor,
+        session_id: SessionId,
+        max_rows: int,
+        max_bytes: int,
+        cursor: "Cursor",
```
jprakash-db: Why is Cursor in quotes ?

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -1089,46 +1224,20 @@ def fetch_results(
...
+            raise ValueError("Not a valid Thrift command ID")
 
-    def cancel_command(self, active_op_handle):
-        logger.debug(
-            "Cancelling command {}".format(
-                self.guid_to_hex_id(active_op_handle.operationId.guid)
-            )
-        )
-        req = ttypes.TCancelOperationReq(active_op_handle)
+        logger.debug("Cancelling command {}".format(command_id.to_hex_guid()))
```
jprakash-db: Use lazy formatting while writing logs

src/databricks/sql/backend/types.py
```diff
@@ -0,0 +1,396 @@
...
+        Returns:
+            The session ID string or None if this is not a SEA session ID
+        """
+
+        if self.backend_type != BackendType.SEA:
+            return None
+
+        return self.guid
+
+    def get_guid(self) -> Any:
```
jprakash-db: Python does not need getters and setters. You can directly access variables or use @property if you need some pre processing

src/databricks/sql/backend/utils/guid_utils.py
```diff
@@ -0,0 +1,23 @@
...
+        OUT  '01ee1d29-a419-1db6-a9c0-8df1feba42dd'
+
+    If conversion to hexadecimal fails, a string representation of the original
+    bytes is returned
+    """
+
+    try:
+        this_uuid = uuid.UUID(bytes=guid)
+    except Exception as e:
+        logger.debug(f"Unable to convert bytes to UUID: {guid!r} -- {str(e)}")
```
jprakash-db: Use lazy logging

src/databricks/sql/backend/utils/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .guid_utils import guid_to_hex_id
```
jprakash-db: I think you can just have a empty __init__ file. Without this also you can access the functions in the folder

src/databricks/sql/client.py
```diff
@@ -342,34 +303,40 @@ def __del__(self):
...
+            if openSessionResp.serverProtocolVersion
+            else {}
+        )
+        session_id = SessionId.from_thrift_handle(
+            openSessionResp.sessionHandle, properties
+        )
+        return Session.get_protocol_version(session_id)
+
+    @property
+    def open(self) -> bool:
```
jprakash-db, line 374: is_open is probably a better name

src/databricks/sql/session.py
```diff
@@ -0,0 +1,150 @@
...
+    def server_parameterized_queries_enabled(protocolVersion):
+        if (
+            protocolVersion
+            and protocolVersion >= ttypes.TProtocolVersion.SPARK_CLI_SERVICE_PROTOCOL_V8
+        ):
+            return True
+        else:
+            return False
+
+    def get_session_id(self) -> SessionId:
```
jprakash-db: You don't need getters and setters in python. Either use @property or similiy define the variable name withouth _ indicating that it is public

tests/unit/test_session.py
```diff
@@ -0,0 +1,196 @@
...
+    TOpenSessionResp,
+    TSessionHandle,
+    THandleIdentifier,
+)
+from databricks.sql.backend.types import SessionId, BackendType
+
+import databricks.sql
+
+
+class SessionTestSuite(unittest.TestCase):
```
jprakash-db: Use pytest as we are moving away from unittest

tests/unit/test_session.py
```diff
@@ -0,0 +1,196 @@
...
+        databricks.sql.connect(
+            _retry_stop_after_attempts_count=54, **self.DUMMY_CONNECTION_ARGS
+        )
+
+        self.assertEqual(
+            mock_client_class.call_args[1]["_retry_stop_after_attempts_count"], 54
+        )
+
+    @patch("%s.session.ThriftDatabricksClient" % PACKAGE_NAME)
+    def test_socket_timeout_passthrough(self, mock_client_class):
```
jprakash-db, line 149: Is this just checking if the values passed is being stored correctly ?

tests/unit/test_session.py
```diff
@@ -0,0 +1,196 @@
...
+        connection = databricks.sql.connect(**self.DUMMY_CONNECTION_ARGS)
+        connection.close()
+
+        # Check that close_session was called with the correct SessionId
+        close_session_call_args = instance.close_session.call_args[0][0]
+        self.assertEqual(close_session_call_args.guid, b"\x22")
+        self.assertEqual(close_session_call_args.secret, b"\x33")
+
+    @patch("%s.session.ThriftDatabricksClient" % PACKAGE_NAME)
+    def test_auth_args(self, mock_client_class):
```
jprakash-db, line 44: I think we can remove these tests, that just add the value and check if they are present. These type of tests don't meaningfully check anything

src/databricks/sql/utils.py
```diff
@@ -26,6 +26,7 @@
     TSparkRowSetType,
```
jprakash-db, line 26: Can you move the utils.py file into the utils folder and probably give a new name to it

src/databricks/sql/session.py
```diff
@@ -0,0 +1,150 @@
...
+            if isinstance(e.args[1], SessionAlreadyClosedError):
+                logger.info("Session was closed by a prior request")
+        except DatabaseError as e:
+            if "Invalid SessionHandle" in str(e):
+                logger.warning(
+                    f"Attempted to close session that was already closed: {e}"
+                )
+            else:
+                logger.warning(
+                    f"Attempt to close session raised an exception at the server: {e}"
```
jprakash-db: Pointing it again , but change all the logging to use lazy logging

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,415 @@
...
+    def fetchmany_arrow(self, size: int) -> "pyarrow.Table":
+        """Fetch the next set of rows as an Arrow table."""
+        pass
+
+    @abstractmethod
+    def fetchall_arrow(self) -> "pyarrow.Table":
+        """Fetch all remaining rows as an Arrow table."""
+        pass
+
+    def close(self) -> None:
```
jprakash-db, line 114: If you are creating an ABC leave all of them as ABC. Don't implement anything

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: In the existing implementation (i.e. the `main` branch), nothing (`None`) is returned for async queries. Returning an `EmptyResultSet` would cause a change to user facing functionality, right? 

Edit: Apologies for the misunderstanding, this is not user facing as I stated before. An `EmptyResultSet` with some defaults would be doable. However, I feel it makes more sense to return `None` because we don't actually have any results yet.

src/databricks/sql/client.py
varun-edachali-dbx: In the existing implementation (i.e. on the `main` branch) the property denoting Connection state is `open`, so changing to `is_open` would be a change to user facing functionality.

tests/unit/test_session.py
varun-edachali-dbx: Yes, that seems to be what it is doing. 

Note: I haven't introduced any new unit tests in this PR, or removed old ones. `test_session.py` is the result of moving the session related tests away from `test_client.py`.

jprakash-db: Ok, then don't remove it if it was already existing

src/databricks/sql/result_set.py
varun-edachali-dbx: Is there any reason for that? 

Abstract classes are allowed to have concrete methods, and in this case the `close` implementation only depends on state that is defined in `ResultSet`, so I think it belongs here. This can help reduce repetition. 

cc @jayantsing-db

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: Quotes basically tell Python that it is a forward reference to be resolved during type checking. It prevents circular imports when I only want to use it for type checking. 

However, we decided to use `from __future__ import annotations` to clean this up. I have removed the string literals around the Cursor references by adding this now. I had done this in some other files, but seems like I missed it here. Thanks for the catch.

src/databricks/sql/utils.py
varun-edachali-dbx: I agree that it should be renamed, does `queue.py` sound fine? 

I think it should remain in the existing directory though. The `Connection`, `Cursor` and `ResultSet` definitions are in the existing directory, so I think the `Queue` definitions should remain here too. 
Also, if you were referring to `backend/utils` by `utils` folder above, then I don't think this should move there because this is not backend specific. 

cc @jayantsing-db

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
+from __future__ import annotations
```
jayantsing-db, line 1: Please highlight the purpose of refactor in PR description.

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+
+from abc import ABC, abstractmethod
+from typing import Dict, List, Optional, Any, Union, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from databricks.sql.client import Cursor
+    from databricks.sql.result_set import ResultSet
```
jayantsing-db, line 8: could you remind me the importance of this?

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+    # == Query Execution, Command Management ==
+    @abstractmethod
+    def execute_command(
+        self,
+        operation: str,
+        session_id: SessionId,
+        max_rows: int,
+        max_bytes: int,
+        lz4_compression: bool,
+        cursor: Cursor,
```
jayantsing-db, line 80: do we need cursor in the function signature? cursor requirement looks like an implementation detail. The concrete implementation of databricks_client should be composed with a cursor object and use that to implement this method if need be.

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+            ServerOperationError: If the command is in an error state
+            DatabaseError: If the command has been closed unexpectedly
+        """
+        pass
+
+    @abstractmethod
+    def get_execution_result(
+        self,
+        command_id: CommandId,
+        cursor: Cursor,
```
jayantsing-db, line 175: do we need cursor in signature? the implementing class can be composed with a cursor and use this in the method. Can be a separate PR to limit the scope.

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+        pass
+
+    # == Metadata Operations ==
+    @abstractmethod
+    def get_catalogs(
+        self,
+        session_id: SessionId,
+        max_rows: int,
+        max_bytes: int,
+        cursor: Cursor,
```
jayantsing-db, line 203: Same as above.

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def get_schemas(
+        self,
+        session_id: SessionId,
+        max_rows: int,
+        max_bytes: int,
+        cursor: Cursor,
```
jayantsing-db, line 232: same as above

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def get_tables(
+        self,
+        session_id: SessionId,
+        max_rows: int,
+        max_bytes: int,
+        cursor: Cursor,
```
jayantsing-db, line 265: same as above.

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+
+        This method fetches metadata about tables available in the specified catalog
+        and schema, or all catalogs and schemas if not specified.
+
+        Args:
+            session_id: The session identifier
+            max_rows: Maximum number of rows to fetch in a single batch
+            max_bytes: Maximum number of bytes to fetch in a single batch
+            cursor: The cursor object that will handle the results
+            catalog_name: Optional catalog name pattern to filter by
```
jayantsing-db, line 282: maybe mention here that if catalog is null, it fetches across all catalogs.

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def get_columns(
+        self,
+        session_id: SessionId,
+        max_rows: int,
+        max_bytes: int,
+        cursor: Cursor,
```
jayantsing-db, line 304: Do we need cursor in the signature? Seems out-of-place when compared with other params.

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+        catalog_name: Optional[str] = None,
+        schema_name: Optional[str] = None,
+        table_name: Optional[str] = None,
+        column_name: Optional[str] = None,
+    ) -> ResultSet:
+        """
+        Retrieves a list of columns, optionally filtered by catalog, schema, table, and column name patterns.
+
+        This method fetches metadata about columns available in the specified table,
+        or all tables if not specified.
```
jayantsing-db, line 314: does it make sense to fetch columns across all tables? Is it added because of backward compatibility?

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Args:
+            session_id: The session identifier
+            max_rows: Maximum number of rows to fetch in a single batch
+            max_bytes: Maximum number of bytes to fetch in a single batch
+            cursor: The cursor object that will handle the results
+            catalog_name: Optional catalog name pattern to filter by
+            schema_name: Optional schema name pattern to filter by
+            table_name: Optional table name pattern to filter by
```
jayantsing-db, line 323: Maybe mention here the behaviour when table_name is null.

src/databricks/sql/backend/databricks_client.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Raises:
+            ValueError: If the session ID is invalid
+            OperationalError: If there's an error retrieving the columns
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def max_download_threads(self) -> int:
```
jayantsing-db, line 338: i think we plan to move this out sometime in the future?

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -91,7 +103,6 @@ def __init__(
         http_headers,
         auth_provider: AuthProvider,
         ssl_options: SSLOptions,
-        staging_allowed_local_path: Union[None, str, List[str]] = None,
```
jayantsing-db, line 94: could you remind me, where is this moved?

src/databricks/sql/backend/thrift_backend.py
jayantsing-db: I think returning empty result set makes sense. But maybe in a separate PR in future to limit the scope of this.

src/databricks/sql/backend/thrift_backend.py
jayantsing-db: what was the purpose of this lazy import? is it supposed to improve module loading?

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -975,58 +1029,106 @@ def execute_command(
...
+        session_id: SessionId,
+        max_rows: int,
+        max_bytes: int,
+        cursor: Cursor,
         catalog_name=None,
         schema_name=None,
-    ):
-        assert session_handle is not None
+    ) -> "ResultSet":
+        from databricks.sql.result_set import ThriftResultSet
```
jayantsing-db: what was the purpose of this lazy import? I see few other methods returning ThriftResultSet don't have this.

src/databricks/sql/backend/types.py
```diff
@@ -0,0 +1,391 @@
...
+        Raises:
+            ValueError: If the provided state is not a recognized TOperationState
+
+        State Mappings:
+            - INITIALIZED_STATE, PENDING_STATE -> PENDING
+            - RUNNING_STATE -> RUNNING
+            - FINISHED_STATE -> SUCCEEDED
+            - ERROR_STATE, TIMEDOUT_STATE, UKNOWN_STATE -> FAILED
+            - CLOSED_STATE -> CLOSED
+            - CANCELED_STATE -> CANCELLED
```
jayantsing-db, line 58: could any of these mappings lead to a user-facing behavioural change?

src/databricks/sql/client.py
```diff
@@ -236,82 +240,39 @@ def read(self) -> Optional[OAuthToken]:
...
-            tls_verify=not kwargs.get(
-                "_tls_no_verify", False
-            ),  # by default - verify cert and host
-            tls_verify_hostname=kwargs.get("_tls_verify_hostname", True),
-            tls_trusted_ca_file=kwargs.get("_tls_trusted_ca_file"),
-            tls_client_cert_file=kwargs.get("_tls_client_cert_file"),
-            tls_client_cert_key_file=kwargs.get("_tls_client_cert_key_file"),
-            tls_client_cert_key_password=kwargs.get("_tls_client_cert_key_password"),
-        )
-
```
jayantsing-db, line 284: all of this moved to Session?

src/databricks/sql/client.py
```diff
@@ -236,82 +240,39 @@ def read(self) -> Optional[OAuthToken]:
...
         )
+        self.staging_allowed_local_path = kwargs.get("staging_allowed_local_path", None)
 
         TelemetryClientFactory.initialize_telemetry_client(
             telemetry_enabled=self.telemetry_enabled,
             session_id_hex=self.get_session_id_hex(),
-            auth_provider=auth_provider,
-            host_url=self.host,
+            auth_provider=self.session.auth_provider,
+            host_url=self.session.host,
```
jayantsing-db, line 275: On these changes conflicting with telemetry, can we get the telemetry folks stamp?

src/databricks/sql/client.py
jayantsing-db: agreed on maintaining backward compatibility in this PR

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,415 @@
...
+    """
+    Abstract base class for result sets returned by different backend implementations.
+
+    This class defines the interface that all concrete result set implementations must follow.
+    """
+
+    def __init__(
+        self,
+        connection: Connection,
+        backend: DatabricksClient,
```
jayantsing-db, line 38: can this be removed/normalised in future? is backend simply, connection.session.backend? ideally, we should be consistent in accessing the internal backend throughout the codebase

src/databricks/sql/result_set.py
```diff
@@ -0,0 +1,415 @@
...
+        )
+
+        res = df.to_numpy(na_value=None, dtype="object")
+        return [ResultRow(*v) for v in res]
+
+    def merge_columnar(self, result1, result2) -> "ColumnTable":
+        """
+        Function to merge / combining the columnar results into a single result
+        :param result1:
+        :param result2:
```
jayantsing-db, line 258: all of this is ported code?

src/databricks/sql/utils.py
jayantsing-db: i think we can consider renaming in future.

jayantsing-db (approved):
Great work, i think the implementation of sql connector makes more sense now and much more open to extension. Some comments inline.

Thanks very much @jprakash-db for the review.

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: I disagree that the concrete implementation of `databricks_client` should be composed with a `Cursor` object: the `databricks_client` is instantiated at the `Connection` level and a `Connection` can have multiple `Cursor`s. If we were to define a new client for each `Cursor`, then the above proposition would be doable. 

That being said, the `Cursor` only seems to be used in the `execute_command` code path in order to set the active command id. It would likely be cleaner to stop passing the `Cursor` and set the `CommandId` using the response from `execute_command`, since we get it there anyway and the `Cursor` is what invokes the `execute_command` method. But, in order to do this we would have to change the async execution flow to return the `CommandId` / some empty `ResultSet` instead of `None` as proposed by @jprakash-db earlier. 

Should I do this?

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: addressed [here](https://github.com/databricks/databricks-sql-python/pull/604/files?w=1#r2203687316)

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: addressed [here](https://github.com/databricks/databricks-sql-python/pull/604/files?w=1#r2203687316)

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: addressed [here]

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: addressed [here](https://github.com/databricks/databricks-sql-python/pull/604/files?w=1#r2203687316)

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: addressed [here](https://github.com/databricks/databricks-sql-python/pull/604/files?w=1#r2203687316)

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: Yes. The existing definition of `get_columns` has the `table_name` as `Optional` so I maintained this behaviour.

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: I think we moved out `ssl_options` etc. but we decided to maintain `max_download_threads` because the creator of the `ResultSet` should have `max_download_threads`. We could change it to a `Connection` level attribute if it seems out of place here. Should I do that?

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: [PECOBLR-649](https://databricks.atlassian.net/browse/PECOBLR-649?atlOrigin=eyJpIjoiZjczMTRhYTgwZDZhNDEyMjk2NGU2NGM1MmMwMDQyZTAiLCJwIjoiaiJ9)

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: You're right - it's not required. I probably added it initially to resolve some circular dependencies, but it is not required in the current state of the codebase. Thanks for the catch!

src/databricks/sql/backend/types.py
varun-edachali-dbx: I think not, because the user should not directly interact with the state at any point. In case of async execution, we provide helper methods through the `Cursor` like `is_query_pending` and `get_async_execution_result` that abstract away the exact state at any point. 

The only other risk lies in `thrift_backend` because some errors are based on state that do not have a unique mapping to another state in `CommandState` (eg: some exceptions are raised particularly based on `ERROR_STATE`, but we if we convert it to `FAILED` then the same error will also be raised for `TIMEDOUT_STATE` and `UNKNOWN_STATE`), but we maintained existing behaviour here by using `TOperationState` as far as possible for internal operations until we finally pass a `CommandState` when we exit the client layer.

src/databricks/sql/client.py
varun-edachali-dbx: cc @vikrantpuppala

src/databricks/sql/client.py
varun-edachali-dbx: Yes

src/databricks/sql/result_set.py
varun-edachali-dbx: Yes, this will be removed soon as a part of #626.

src/databricks/sql/result_set.py
varun-edachali-dbx: Yes

src/databricks/sql/backend/databricks_client.py
varun-edachali-dbx: The `TYPE_CHECKING` block allows us to make circular _references_, we're basically stating that the imports within that block are only required for type checking, and not during run-time. Usually, this is coupled with placing this reference in quotes, but we can avoid doing this by using `from __future__ import annotations`. 

A reference: https://vickiboykis.com/2023/12/11/why-if-type_checking/

I copied the above from [here](https://github.com/databricks/databricks-sql-python/pull/594/files#r2179443651).

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: This was made a `Connection` level property

vikrantpuppala (approved):
telemetry changes lgtm

src/databricks/sql/backend/databricks_client.py
jayantsing-db: please document the corresponding cursor is passed to update its state like command id. that should be sufficient imo.

PR #607, opened 2025-06-19
Introduce `row_limit` param 

varun-edachali-dbx:
## What type of PR is this?
- [x] Feature

## Description
Introduces a `row_limit` param in the `Cursor` that constrains the number of rows in the query result, through `resultRowLimit` in Thrift and through `max_rows` in SEA. 

Despite `row_limit` not existing [in the spec](https://peps.python.org/pep-0249/), we feel it is reasonable to introduce it as a param in the `Cursor`, because: 
- the `Connection` class has a number of non-spec params 
- a risk of non-spec methods is the feature not being discovered (i.e., users not realising this param is available to them), but this is not a major issue, since the feature can effectively be replicated by including `LIMIT` in their SQL query. 

## How is this tested?

- [ ] Unit tests
- [x] E2E Tests
- [ ] Manually
- [ ] N/A

N/A


src/databricks/sql/backend/databricks_client.py
```diff
@@ -103,6 +104,7 @@ def execute_command(
             parameters: List of parameters to bind to the query
             async_op: Whether to execute the command asynchronously
             enforce_embedded_schema_correctness: Whether to enforce schema correctness
+            row_limit: Maximum number of rows to fetch overall. Only supported for SEA protocol.
```
shivam2680: tangential, but you can make this change for thrift as well. 
Add resultRowLimit = maxRows while constructing `TExecuteStatementReq` in `execute_statement` in thfit_backend.

jayantsing-db: thanks @shivam2680

varun-edachali-dbx: Thanks @shivam2680, I added `resultRowLimit` to Thrift and an associated e2e test.

varun-edachali-dbx:
> is this a breaking change in thrift? the thrift users would have to change to row_limit?

There is currently no way for the user to specify a row limit with the Thrift backend. The newly introduced `row_limit` will set `resultRowLimit` in `TExecuteStatementReq` and allow the user to do so. It is not a breaking change because this defaults to `None`, which would maintain current behaviour. 

In the current implementation, `arraysize` is set as the `max_rows` for direct results in case of Thrift. It does not constrain the total rows in the result. `row_limit` and `arraysize` will serve distinct purposes.

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -969,6 +970,7 @@ def execute_command(
             useArrowNativeTypes=spark_arrow_types,
             parameters=parameters,
             enforceEmbeddedSchemaCorrectness=enforce_embedded_schema_correctness,
+            resultRowLimit=row_limit,
```
jayantsing-db, line 973: ok to set null in resultRowLimit?

src/databricks/sql/client.py
```diff
@@ -335,6 +335,7 @@ def cursor(
         self,
         arraysize: int = DEFAULT_ARRAY_SIZE,
         buffer_size_bytes: int = DEFAULT_RESULT_BUFFER_SIZE_BYTES,
+        row_limit: Optional[int] = None,
```
jayantsing-db, line 338: please document in docstring

jayantsing-db (approved):

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: Yes, `None` is valid input. Most of the integration tests run with it set as `None`.

PR #608, opened 2025-06-19
Add functionality for export of latency logs via telemetry

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Add functionality for export of latency logs via telemetry

Example latency log:
    {
      "frontend_log_event_id": "30565427-1299-4fbc-bec0-25378cfe912b",
      "context": {
        "client_context": {
          "timestamp_millis": 1750592858108,
          "user_agent": "PyDatabricksSqlConnector/4.0.4"
        }
      },
      "entry": {
        "sql_driver_log": {
          "session_id": "01f04f5e-b1a8-1dd1-962a-e40efd092bdd",
          "system_configuration": {
            "driver_version": "4.0.4",
            "os_name": "Darwin",
            "os_version": "24.5.0",
            "os_arch": "arm64",
            "runtime_version": "3.13.3",
            "runtime_vendor": "CPython",
            "char_set_encoding": "utf-8",
            "locale_name": "en_US"
          },
          "driver_connection_params": {
            "http_path": "/sql/1.0/warehouses/864004c1b3961382",
            "mode": "THRIFT",
            "host_info": {
              "host_url": "e2-dogfood.staging.cloud.databricks.com",
              "port": 443
            },
            "auth_mech": "PAT"
          },
          "sql_statement_id": "01f04f5e-b1d6-1a1e-afe4-e99e1ccb8805",
          "sql_operation": {
            "statement_type": "sql",
            "is_compressed": true,
            "execution_result": "inline_arrow",
            "retry_count": 0
          },
          "operation_latency_ms": 518
        }
      }
    }
## How is this tested?

- [x] Unit tests 
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

PECOBLR-554


src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,131 @@
...
+from typing import Optional
+from uuid import UUID
+from databricks.sql.telemetry.telemetry_client import TelemetryClientFactory
+from databricks.sql.telemetry.models.event import (
+    SqlExecutionEvent,
+)
+from databricks.sql.telemetry.models.enums import ExecutionResultFormat, StatementType
+from databricks.sql.utils import ColumnQueue
+
+# Helper to get statement_id/query_id from instance if available
```
vikrantpuppala: don't need this since we have docstring

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,131 @@
...
+    if hasattr(instance, "query_id"):
+        return instance.query_id
+
+    if hasattr(instance, "command_id") and instance.command_id:
+        return str(UUID(bytes=instance.command_id.operationId.guid))
+
+    return None
+
+
+def _get_session_id_hex(instance) -> Optional[str]:
```
vikrantpuppala: can you add docstring which explains which if conditions is for which class, can we also check class instead of hasattr? that will make it clear which class we refer to in each case

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,131 @@
...
+    1. Direct lz4_compression attribute (Connection)
+    2. Through connection attribute (Cursor/ResultSet)
+    3. Through thrift_backend attribute (Cursor)
+    """
+    if hasattr(instance, "lz4_compression"):
+        return instance.lz4_compression
+    if hasattr(instance, "connection") and instance.connection:
+        return instance.connection.lz4_compression
+    if hasattr(instance, "thrift_backend") and instance.thrift_backend:
+        return instance.thrift_backend.lz4_compressed
```
vikrantpuppala: similar to previous comment i'd prefer something like below which is self explainatory to read:
```
    if isinstance(instance, Connection):
        return instance.lz4_compression
    if isinstance(instance, Cursor/ResultSet) and instance.connection:
        return instance.connection.lz4_compression
    if isinstance(instance, Cursor) and instance.thrift_backend:
        return instance.thrift_backend.lz4_compressed
```

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,131 @@
...
+    if hasattr(instance, "active_result_set") and instance.active_result_set:
+        if isinstance(instance.active_result_set.results, ColumnQueue):
+            return ExecutionResultFormat.COLUMNAR_INLINE
+        return ExecutionResultFormat.INLINE_ARROW
+
+    if hasattr(instance, "thrift_backend") and instance.thrift_backend:
+        if hasattr(instance.thrift_backend, "_use_arrow_native_complex_types"):
+            return ExecutionResultFormat.INLINE_ARROW
+
+    return ExecutionResultFormat.FORMAT_UNSPECIFIED
```
vikrantpuppala: all of these functions are ifs for Cursor/Connection/ResultSet, can we do something here like have a base class and have implementations for each class?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,131 @@
...
+            try:
+                result = func(self, *args, **kwargs)
+                return result
+            finally:
+                end_time = time.perf_counter()
+                duration_ms = int((end_time - start_time) * 1000)
+
+                session_id_hex = _get_session_id_hex(self)
+
+                if session_id_hex:
```
vikrantpuppala: we're inherently ignoring calls to this decorator when we cannot figure out the session_id_hex, when can this happen?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    # TODO: implement this
+
+    return 0
+
+
+def log_latency():
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(self, *args, **kwargs):
```
vikrantpuppala, line 191: what happens if something fails in this decorator? we should be handling that gracefully

src/databricks/sql/telemetry/latency_logger.py
vikrantpuppala: which would anyway also happen since TelemetryClientFactory.get_telemetry_client would return no-op telemetry client?

src/databricks/sql/thrift_backend.py
```diff
@@ -583,6 +584,9 @@ def open_session(self, session_configuration, catalog, schema):
             self._transport.close()
             raise
 
+    def get_session_id_hex(self) -> Optional[str]:
+        return self._session_id_hex
```
vikrantpuppala: do we need to define this function because _session_id_hex is intenal?

saishreeeee: yes

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: yes, get_telemetry_client can handle None values. I will remove the check. Also, session_id_hex is set during connection initialization when it calls open session of thrift backend. So it is highly unlikely session_id_hex would be None.

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: both the functions: get_telemetry_client and export_latency_log have try/catch blocks in them.

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,152 @@
...
+from databricks.sql.telemetry.telemetry_client import TelemetryClientFactory
+from databricks.sql.telemetry.models.event import (
+    SqlExecutionEvent,
+)
+from databricks.sql.telemetry.models.enums import ExecutionResultFormat, StatementType
+from databricks.sql.utils import ColumnQueue, CloudFetchQueue, ArrowQueue
+from uuid import UUID
+
+
+class TelemetryExtractor:
```
vikrantpuppala, line 13: please add docstrings

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,152 @@
...
+    def get_session_id_hex(self) -> Optional[str]:
+        return self.connection.get_session_id_hex()
+
+    def get_is_compressed(self) -> bool:
+        return self.connection.lz4_compression
+
+    def get_execution_result(self) -> ExecutionResultFormat:
+        if self.active_result_set is None:
+            return ExecutionResultFormat.FORMAT_UNSPECIFIED
+
```
vikrantpuppala, line 78: let's remove this if condition?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,152 @@
...
+    def get_retry_count(self) -> int:
+        if (
+            hasattr(self.thrift_backend, "retry_policy")
+            and self.thrift_backend.retry_policy
+        ):
+            return len(self.thrift_backend.retry_policy.history)
+        return 0
+
+    def get_statement_type(self) -> StatementType:
+        # TODO: Implement this
```
vikrantpuppala: why does this need to be a TODO?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,152 @@
...
+        return self.lz4_compressed
+
+    def get_execution_result(self) -> ExecutionResultFormat:
+        if isinstance(self.results, ColumnQueue):
+            return ExecutionResultFormat.COLUMNAR_INLINE
+        elif isinstance(self.results, CloudFetchQueue):
+            return ExecutionResultFormat.EXTERNAL_LINKS
+        elif isinstance(self.results, ArrowQueue):
+            return ExecutionResultFormat.INLINE_ARROW
+        return ExecutionResultFormat.FORMAT_UNSPECIFIED
```
vikrantpuppala, line 122: since this code is common can we abstract this out to separate function?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def get_statement_type(self) -> StatementType:
+        # TODO: Implement this
```
vikrantpuppala: lets see if we can solve these todo's right now

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,152 @@
...
+        return 0
+
+
+def get_extractor(obj):
+    if obj.__class__.__name__ == "Cursor":
+        return CursorExtractor(obj)
+    elif obj.__class__.__name__ == "ResultSet":
+        return ResultSetExtractor(obj)
+    else:
+        return TelemetryExtractor(obj)
```
vikrantpuppala: can we throw exception instead

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -299,6 +306,32 @@ def export_failure_log(self, error_name, error_message):
...
+                    sql_driver_log=TelemetryEvent(
+                        session_id=self._session_id_hex,
+                        system_configuration=TelemetryHelper.get_driver_system_configuration(),
+                        driver_connection_params=self._driver_connection_params,
+                        sql_statement_id=sql_statement_id,
+                        sql_operation=sql_execution_event,
+                        operation_latency_ms=latency_ms,
+                    )
+                ),
+            )
```
vikrantpuppala: i think a lot of this code is common between the export methods, can we abstract this out?

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: If `self.active_result_set` is `None` and we try to access `self.active_result_set.results` in the subsequent `isinstance` checks, Python would raise an AttributeError because you can't access attributes on a `None` object.

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: will derive it from function names

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,223 @@
...
+
+    Args:
+        obj: The object to create an extractor for. Can be a Cursor, ResultSet,
+             or any other object.
+
+    Returns:
+        TelemetryExtractor: A specialized extractor instance:
+            - CursorExtractor for Cursor objects
+            - ResultSetExtractor for ResultSet objects
+            - TelemetryExtractor (base) for all other objects
```
vikrantpuppala: update comment since we actually throw error otherwise

tests/unit/test_fetches.py
```diff
@@ -79,7 +87,7 @@ def fetch_results(
 
             return results, batch_index < len(batch_list)
 
-        mock_thrift_backend = Mock()
+        mock_thrift_backend = FetchTests.mock_thrift_backend_with_retry_policy()
```
vikrantpuppala: do we need to create a method if we're using it in a single place?

saishreeeee: yeah I will change that

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,223 @@
...
+        obj: The object to create an extractor for. Can be a Cursor, ResultSet,
+             or any other object.
+
+    Returns:
+        TelemetryExtractor: A specialized extractor instance:
+            - CursorExtractor for Cursor objects
+            - ResultSetExtractor for ResultSet objects
+            - Throws an NotImplementedError for all other objects
+    """
+    if obj.__class__.__name__ == "Cursor":
```
jprakash-db, line 150: use instance of , instead of string matching name

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,223 @@
...
+            finally:
+                end_time = time.perf_counter()
+                duration_ms = int((end_time - start_time) * 1000)
+
+                extractor = get_extractor(self)
+                session_id_hex = extractor.get_session_id_hex()
+                statement_id = extractor.get_statement_id()
+
+                sql_exec_event = SqlExecutionEvent(
+                    statement_type=statement_type,
```
jprakash-db, line 214: Can we export the name of the function? Like Cursor.execute is the one whom this latency refers to cc @vikrantpuppala

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,223 @@
...
+            result = None
+            try:
+                result = func(self, *args, **kwargs)
+                return result
+            finally:
+                end_time = time.perf_counter()
+                duration_ms = int((end_time - start_time) * 1000)
+
+                extractor = get_extractor(self)
+                session_id_hex = extractor.get_session_id_hex()
```
jprakash-db: Add error handling when certain attributes can't be accessed. Currently it will fail everything, even if one attribute is not present/ accessible . Within the extractor return default values for fields that can't be accessed

tests/unit/test_fetches.py
```diff
@@ -80,6 +80,9 @@ def fetch_results(
             return results, batch_index < len(batch_list)
 
         mock_thrift_backend = Mock()
+        mock_retry_policy = Mock()
```
jprakash-db: What is the point of these changes? They don't seem to be used anywhere

tests/unit/test_telemetry.py
```diff
@@ -96,6 +99,12 @@ def test_export_failure_log(self, noop_telemetry_client):
             error_name="TestError", error_message="Test error message"
         )
 
+    def test_export_latency_log(self, noop_telemetry_client):
```
jprakash-db: What is the point of these tests? `test_export_latency_log`, `test_export_failure_log` and `test_close` on a `NoOpTelemetryClient`, you aren't testing anything

tests/unit/test_telemetry.py
```diff
@@ -181,6 +190,40 @@ def test_export_failure_log(
         
         client._export_event.assert_called_once_with(mock_frontend_log.return_value)
 
+    @patch("databricks.sql.telemetry.telemetry_client.TelemetryFrontendLog")
```
jprakash-db: I am unable to understand the point of many of the tests in this file? Such as  `test_export_latency_log`, `test_flush`, test_telemetry_request_callback_success` `test_close`. All these tests are calling a function once and then asserting it is being called once, what is the point? Like there is no obvious flow we are testing here cc @vikrantpuppala

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: I had used isinstance check initially but that leads to a circular import, hence went with the string matching name.

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: There have been updates in the fields ([doc](https://docs.google.com/document/d/12pT11BSbqDoRjL0kIo6-xkkx0m3CFmgSRD2lA40jRVE/edit?tab=t.0)), operation_name will fix that issue. (will be added in a follow-up PR)

tests/unit/test_fetches.py
saishreeeee: The tests here check the functionality of functions which have the `log_latency()` decorator applied to them. The decorator tries to access `len(thrift_backend.retry_policy.history)` but `thrift_backend` is mocked here, so this raises an error `TypeError: object of type Mock has no len()`.

tests/unit/test_telemetry.py
saishreeeee: This is to check that calling these functions using `NoOpTelemetryClient` does not raise exceptions. When telemetry is disabled, the no-op implementation is used and that shouldn't cause issues.

tests/unit/test_telemetry.py
saishreeeee: In case of `test_export_latency_log`, the test ensures that `TelemetryFrontendLog` is called once to create the log structure and that `_export_event` gets called with that frontend log as its argument. The same goes for other tests, they verify that the functions they test invoke the expected functions.

vikrantpuppala (approved):

jprakash-db (approved):
LGTM, Thanks for the making the changes

PR #616, opened 2025-06-25
Testing for telemetry

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Other

## Description
Added multithread tests for race conditions

## How is this tested?

- [ ] Unit tests
- [x] E2E Tests
- [ ] Manually
- [ ] N/A

PECOBLR-584

### DIFF
--- a/tests/e2e/test_concurrent_telemetry.py
+++ b/tests/e2e/test_concurrent_telemetry.py
@@ -2,6 +2,7 @@ import threading
 from unittest.mock import patch
 import pytest
 
+from databricks.sql.telemetry.models.enums import StatementType
 from databricks.sql.telemetry.telemetry_client import TelemetryClient, TelemetryClientFactory
 from tests.e2e.test_driver import PySQLPytestTestCase
 
@@ -49,6 +50,9 @@ class TestE2ETelemetry(PySQLPytestTestCase):
         num_threads = 5
         captured_telemetry = []
         captured_telemetry_lock = threading.Lock()
+        captured_session_ids = []
+        captured_statement_ids = []
+        capture_info_lock = threading.Lock()
 
         original_send_telemetry = TelemetryClient._send_telemetry
 
@@ -62,8 +66,17 @@ class TestE2ETelemetry(PySQLPytestTestCase):
             def execute_query_worker(thread_id):
                 """Each thread creates a connection and executes a query."""
                 with self.connection(extra_params={"enable_telemetry": True}) as conn:
+                    # Capture the session ID from the connection before executing the query
+                    session_id_hex = conn.get_session_id_hex()
+                    with capture_info_lock:
+                        captured_session_ids.append(session_id_hex)
+                    
                     with conn.cursor() as cursor:
                         cursor.execute(f"SELECT {thread_id}")
+                        # Capture the statement ID after executing the query
+                        statement_id = cursor.query_id
+                        with capture_info_lock:
+                            captured_statement_ids.append(statement_id)
                         cursor.fetchall()
 
             # Run the workers concurrently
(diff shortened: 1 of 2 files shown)


tests/unit/test_telemetry.py
```diff
@@ -498,4 +499,397 @@ def test_global_exception_hook(self, mock_handle_exception, telemetry_system_res
...
+        
+        # Should not raise exception (executor shutdown is wrapped in try-catch)
+        TelemetryClientFactory.close(session_id_hex)
+        
+        # Verify executor shutdown was attempted
+        mock_executor.shutdown.assert_called_once_with(wait=True)
+
+
+
+class TestTelemetryRaceConditions:
```
vikrantpuppala: should we add these in a separate file? @jprakash-db what's the sop in python?

tests/unit/test_telemetry.py
```diff
@@ -498,4 +499,397 @@ def test_global_exception_hook(self, mock_handle_exception, telemetry_system_res
...
+        for i in range(15):
+            thread = threading.Thread(target=mixed_operations, args=(i,))
+            threads.append(thread)
+            thread.start()
+        
+        # Wait for all threads to complete
+        for thread in threads:
+            thread.join()
+        
+        assert len(operation_results) == 15
```
vikrantpuppala: i think we should have some tests where there are concurrently threads executing queries and the telemetry is being captured correctly, is that possible to do?

tests/unit/test_telemetry.py
```diff
@@ -498,4 +499,397 @@ def test_global_exception_hook(self, mock_handle_exception, telemetry_system_res
...
+        
+        # Wait for all threads to complete
+        for thread in threads:
+            thread.join() 
+       
+        # Verify client is no longer available after close
+        client = TelemetryClientFactory.get_telemetry_client(session_id_hex)
+        assert isinstance(client, NoopTelemetryClient)
+
+    def test_telemetry_client_factory_mixed_concurrent_operations(self, race_condition_setup):
```
vikrantpuppala: do we need to have separate tests for this i.e. concurrent_init, concurrent_get, concurrent_flush etc? can we not simply have a larger thread pool and just do all of these in a single test?

saishreeeee: Keeping separate tests helps in finding which operation specifically is causing the race condition. The `test_telemetry_client_factory_mixed_concurrent_operations` tests the mixed-operation scenario which is more realistic. So both scenarios (focused on a particular operation and the mixed one) are covered.

tests/e2e/test_concurrent_telemetry.py
```diff
@@ -0,0 +1,82 @@
...
+                TelemetryClientFactory._executor.shutdown(wait=True)
+
+            # --- VERIFICATION ---
+            assert len(captured_telemetry) == num_threads * 2 # 2 events per thread (initial_telemetry_log, latency_log (execute))
+
+            events_with_latency = [
+                e for e in captured_telemetry
+                if e.entry.sql_driver_log.operation_latency_ms is not None and e.entry.sql_driver_log.sql_statement_id is not None
+            ]
+            assert len(events_with_latency) == num_threads # 1 event per thread (execute)
```
vikrantpuppala: can we do assertions on content of these logs as well?

tests/e2e/test_concurrent_telemetry.py
```diff
@@ -0,0 +1,137 @@
...
+                        # Capture the statement ID after executing the query
+                        statement_id = cursor.query_id
+                        with capture_info_lock:
+                            captured_statement_ids.append(statement_id)
+                        cursor.fetchall()
+
+            # Run the workers concurrently
+            run_in_threads(execute_query_worker, num_threads, pass_index=True)
+
+            if TelemetryClientFactory._executor:
```
jprakash-db, line 104: Why is there another teardown here?

tests/e2e/test_concurrent_telemetry.py
```diff
@@ -0,0 +1,137 @@
...
+                e for e in captured_telemetry
+                if e.entry.sql_driver_log.operation_latency_ms is None
+                and e.entry.sql_driver_log.driver_connection_params is not None
+                and e.entry.sql_driver_log.system_configuration is not None
+            ]
+            latency_logs = [
+                e for e in captured_telemetry
+                if e.entry.sql_driver_log.operation_latency_ms is not None 
+                and e.entry.sql_driver_log.sql_statement_id is not None 
+                and e.entry.sql_driver_log.sql_operation.statement_type == StatementType.QUERY
```
jprakash-db, line 133: For latency logs, isn't just the check of` ==StatementType.QUERY` enough ?

tests/e2e/test_concurrent_telemetry.py
```diff
@@ -0,0 +1,137 @@
...
+        captured_session_ids = []
+        captured_statement_ids = []
+        capture_info_lock = threading.Lock()
+
+        original_send_telemetry = TelemetryClient._send_telemetry
+
+        def send_telemetry_wrapper(self_client, events):
+            with captured_telemetry_lock:
+                captured_telemetry.extend(events)
+            original_send_telemetry(self_client, events)
```
jprakash-db, line 59: post capture do we actually need to send the request to the server ? As anyway we are not using this to confirm anything in the test . cc @vikrantpuppala

tests/e2e/test_concurrent_telemetry.py
```diff
@@ -0,0 +1,137 @@
...
+                """Each thread creates a connection and executes a query."""
+                with self.connection(extra_params={"enable_telemetry": True}) as conn:
+                    # Capture the session ID from the connection before executing the query
+                    session_id_hex = conn.get_session_id_hex()
+                    with capture_info_lock:
+                        captured_session_ids.append(session_id_hex)
+                    
+                    with conn.cursor() as cursor:
+                        cursor.execute(f"SELECT {thread_id}")
+                        # Capture the statement ID after executing the query
```
jprakash-db, line 95: add random small wait over here to accurately test context switches

tests/e2e/test_concurrent_telemetry.py
```diff
@@ -0,0 +1,137 @@
...
+            TelemetryClientFactory._executor.shutdown(wait=True)
+            TelemetryClientFactory._executor = None
+        TelemetryClientFactory._initialized = False
+
+    def test_concurrent_queries_sends_telemetry(self):
+        """
+        An E2E test where concurrent threads execute real queries against
+        the staging endpoint, while we capture and verify the generated telemetry.
+        """
+        num_threads = 5
```
jprakash-db: 5 threads is to less, atleast use 30-40

tests/e2e/test_concurrent_telemetry.py
```diff
@@ -0,0 +1,137 @@
...
+        # --- SETUP ---
+        if TelemetryClientFactory._executor:
+            TelemetryClientFactory._executor.shutdown(wait=True)
+        TelemetryClientFactory._clients.clear()
+        TelemetryClientFactory._executor = None
+        TelemetryClientFactory._initialized = False
+
+        yield  # This is where the test runs
+
+        # --- TEARDOWN ---
```
jprakash-db: Teardown needs to be inside finally block

tests/e2e/test_concurrent_telemetry.py
saishreeeee: Wanted to make sure all the assertions were made once all the telemetry was sent, and since it is async had to wait for executor to shutdown.

tests/e2e/test_concurrent_telemetry.py
saishreeeee: Latency log, could have `StatementType` query, metadata or sql. Checking whether `entry.sql_driver_log.operation_latency_ms` is not None confirms that it is a latency log. Added the other checks to make it more precise.

tests/e2e/test_concurrent_telemetry.py
saishreeeee: Isn't the use of `yield` in a pytest fixture the same as a `try...finally` block? It is also used in the pytest fixture in `test_complex_types.py`.

tests/e2e/test_concurrent_telemetry.py
saishreeeee: Have added checks on the response received from the server. Wanted it to be an e2e test, so want to have the network calls in it.

tests/e2e/test_concurrent_telemetry.py
jprakash-db: No, just yielding will not cause the cleanup code , after the yield block if there is an error. We need to improve the test_complex_types also then. 
nit: Also isn't just having cleanup sufficent as it is running after every test. Do you need to cleanup before and after

PR #617, opened 2025-06-25
Add retry mechanism to telemetry requests

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Retry mechanism for telemetry requests

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [ ] Manually
- [ ] N/A

PECOBLR-586

### DIFF
--- a/tests/unit/test_telemetry_retry.py
+++ b/tests/unit/test_telemetry_retry.py
@@ -1,0 +1,107 @@
+import pytest
+from unittest.mock import patch, MagicMock
+import io
+import time
+
+from databricks.sql.telemetry.telemetry_client import TelemetryClientFactory
+from databricks.sql.auth.retry import DatabricksRetryPolicy
+
+PATCH_TARGET = 'urllib3.connectionpool.HTTPSConnectionPool._get_conn'
+
+def create_mock_conn(responses):
+    """Creates a mock connection object whose getresponse() method yields a series of responses."""
+    mock_conn = MagicMock()
+    mock_http_responses = []
+    for resp in responses:
+        mock_http_response = MagicMock()
+        mock_http_response.status = resp.get("status")
+        mock_http_response.headers = resp.get("headers", {})
+        body = resp.get("body", b'{}')
+        mock_http_response.fp = io.BytesIO(body)
+        def release():
+            mock_http_response.fp.close()
+        mock_http_response.release_conn = release
+        mock_http_responses.append(mock_http_response)
+    mock_conn.getresponse.side_effect = mock_http_responses
+    return mock_conn
+
+class TestTelemetryClientRetries:
+    @pytest.fixture(autouse=True)
+    def setup_and_teardown(self):
+        TelemetryClientFactory._initialized = False
+        TelemetryClientFactory._clients = {}
+        TelemetryClientFactory._executor = None
+        yield
+        if TelemetryClientFactory._executor:
+            TelemetryClientFactory._executor.shutdown(wait=True)
+        TelemetryClientFactory._initialized = False
+        TelemetryClientFactory._clients = {}
+        TelemetryClientFactory._executor = None
+
+    def get_client(self, session_id, num_retries=3):
+        """
+        Configures a client with a specific number of retries.
+        """
+        TelemetryClientFactory.initialize_telemetry_client(
+            telemetry_enabled=True,
+            session_id_hex=session_id,
+            auth_provider=None,
+            host_url="test.databricks.com",
+        )
+        client = TelemetryClientFactory.get_telemetry_client(session_id)
+        
+        retry_policy = DatabricksRetryPolicy(
+            delay_min=0.01,
+            delay_max=0.02,
+            stop_after_attempts_duration=2.0,
+            stop_after_attempts_count=num_retries, 
+            delay_default=0.1,
+            force_dangerous_codes=[],
+            urllib3_kwargs={'total': num_retries}
+        )
+        adapter = client._http_client.session.adapters.get("https://")
+        adapter.max_retries = retry_policy
+        return client
+
+    @pytest.mark.parametrize(
+    "status_code, description",
+    [
+        (401, "Unauthorized"),
+        (403, "Forbidden"),
+        (501, "Not Implemented"),
+        (200, "Success"),
+    ],
+    )
+    def test_non_retryable_status_codes_are_not_retried(self, status_code, description):
+        """
+        Verifies that terminal error codes (401, 403, 501) and success codes (200) are not retried.
+        """
+        # Use the status code in the session ID for easier debugging if it fails
+        client = self.get_client(f"session-{status_code}")
+        mock_responses = [{"status": status_code}]
+
+        with patch(PATCH_TARGET, return_value=create_mock_conn(mock_responses)) as mock_get_conn:
+            client.export_failure_log("TestError", "Test message")
+            TelemetryClientFactory.close(client._session_id_hex)
+
+            mock_get_conn.return_value.getresponse.assert_called_once()
+
+    def test_exceeds_retry_count_limit(self):
+        """
+        Verifies that the client retries up to the specified number of times before giving up.
+        Verifies that the client respects the Retry-After header and retries on 429, 502, 503.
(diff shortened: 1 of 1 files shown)


src/databricks/sql/exc.py
```diff
@@ -21,11 +19,11 @@ def __init__(
...
         error_name = self.__class__.__name__
-        if session_id_hex:
-            telemetry_client = TelemetryClientFactory.get_telemetry_client(
-                session_id_hex
-            )
-            telemetry_client.export_failure_log(error_name, self.message)
+
+        from databricks.sql.telemetry.telemetry_client import TelemetryClientFactory
+
+        telemetry_client = TelemetryClientFactory.get_telemetry_client(session_id_hex)
```
vikrantpuppala: why this change?

tests/e2e/test_telemetry_retry.py
```diff
@@ -0,0 +1,213 @@
...
+                create_urllib3_response(200)
+            ]
+            
+            client.export_failure_log("TestError", "Test message")
+            self.wait_for_async_request()
+            TelemetryClientFactory.close(client._session_id_hex)
+            
+            # Check that the request was retried (should be 2 calls: initial + 1 retry)
+            assert mock_get_conn.return_value.getresponse.call_count == 2
+            assert "Retrying after" in caplog.text
```
vikrantpuppala: can we collapse these tests a bit i.e. test 503, 429, retry-after header, retry count etc in a single test (first response returns 503, second does 429 etc), appreciate the detailed testing but i would like it to be more readable and maintainable in the long term

tests/e2e/test_telemetry_retry.py
```diff
@@ -0,0 +1,213 @@
...
+        """Test that 400 responses are retried (this is the current behavior for telemetry)"""
+        client, _ = self.get_client("session-400")
+        with patch(PATCH_TARGET) as mock_get_conn:
+            mock_get_conn.return_value.getresponse.return_value = create_urllib3_response(400)
+            
+            client.export_failure_log("TestError", "Test message")
+            self.wait_for_async_request()
+            TelemetryClientFactory.close(client._session_id_hex)
+            
+            # Based on the logs, 400 IS being retried (this is the actual behavior for CommandType.OTHER)
```
vikrantpuppala: 400 is being retried? are you sure?

tests/e2e/test_telemetry_retry.py
```diff
@@ -0,0 +1,213 @@
...
+        with patch(PATCH_TARGET) as mock_get_conn:
+            mock_get_conn.return_value.getresponse.return_value = create_urllib3_response(403)
+            
+            client.export_failure_log("TestError", "Test message")
+            self.wait_for_async_request()
+            TelemetryClientFactory.close(client._session_id_hex)
+            
+            # 403 should not be retried based on the retry policy
+            mock_get_conn.return_value.getresponse.assert_called_once()
+            assert "Telemetry request failed with status code: 403" in caplog.text
```
vikrantpuppala: would like to collapse 4xx error codes into a single test with something like parametrized tests

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -31,6 +33,24 @@
 logger = logging.getLogger(__name__)
 
 
+class TelemetryHTTPAdapter(HTTPAdapter):
```
vikrantpuppala: let's see if we can skip this adapter

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -146,6 +166,11 @@ class TelemetryClient(BaseTelemetryClient):
     It uses a thread pool to handle asynchronous operations, that it gets from the TelemetryClientFactory.
     """
 
+    TELEMETRY_RETRY_STOP_AFTER_ATTEMPTS_COUNT = 3
+    TELEMETRY_RETRY_DELAY_MIN = 0.5  # seconds
+    TELEMETRY_RETRY_DELAY_MAX = 5.0  # seconds
+    TELEMETRY_RETRY_STOP_AFTER_ATTEMPTS_DURATION = 30.0
```
vikrantpuppala: let's align these values across drivers or have these consistent across the python driver

tests/e2e/test_telemetry_retry.py
```diff
@@ -0,0 +1,213 @@
...
+    mock_response.headers = headers or {}
+    mock_response.msg = headers or {}  # For urllib3~=1.0 compatibility
+    mock_response.data = body
+    mock_response.read.return_value = body
+    mock_response.get_redirect_location.return_value = False
+    mock_response.closed = False
+    mock_response.isclosed.return_value = False
+    return mock_response
+
+@pytest.mark.usefixtures("caplog")
```
vikrantpuppala: do we really need caplog?

src/databricks/sql/exc.py
saishreeeee: circular import: telemetry_client.py now imports from retry.py and retry.py imports errors from exc.py

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Problems with shifting adapter logic to `send_telemetry`:
If the retry timer is started in `send_telemetry`, it will start in the main thread but the telemetry request is submitted to the background thread pool (async) and might start later, making the timer inaccurate.
The `request.Session` object does not have a simple `retry_policy` argument to apply the custom retry policy for a single call. 
As `HTTPAdapter` class is part of the `requests` library, there is no need for explicit teardown needed for the timer if the adapter approach is used. The adapter's `send()` acts as a perfect wrapper/ interceptor for the entire request/retry lifecycle.

tests/e2e/test_telemetry_retry.py
saishreeeee: Used it for debugging previously, changed it now.

jprakash-db:
Can you try to incorporate the common http client - https://github.com/databricks/databricks-sql-python/blob/576eafc449e24ebc097a3ecb5754fe505d2af7ab/src/databricks/sql/common/http.py#L42

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -427,6 +429,7 @@ def close(session_id_hex):
                 )
                 try:
                     TelemetryClientFactory._executor.shutdown(wait=True)
+                    TelemetryHttpClient.get_instance().close()
```
jprakash-db: This should be just `TelemtryHttpClient.close`

tests/e2e/test_telemetry_retry.py
```diff
@@ -0,0 +1,107 @@
...
+            delay_max=0.02,
+            stop_after_attempts_duration=2.0,
+            stop_after_attempts_count=num_retries, 
+            delay_default=0.1,
+            force_dangerous_codes=[],
+            urllib3_kwargs={'total': num_retries}
+        )
+        adapter = client._http_client.session.adapters.get("https://")
+        adapter.max_retries = retry_policy
+        return client, adapter
```
jprakash-db: What is the point of returning adapter?

saishreeeee: removed it

tests/e2e/test_telemetry_retry.py
vikrantpuppala, line 1: is this e2e test?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -201,7 +203,7 @@ def _send_telemetry(self, events):
         try:
             logger.debug("Submitting telemetry request to thread pool")
             future = self._executor.submit(
-                requests.post,
+                self._http_client.post,
```
vikrantpuppala, line 211: can we remove requests import

tests/e2e/test_telemetry_retry.py
saishreeeee: It is not, shifted to unit tests.

PR #618, opened 2025-06-27
Enhance SEA HTTP Client 

varun-edachali-dbx:
## What type of PR is this?
- [x] Feature

## Description

Enhance the current, simplified SEA Http Client by introducing retries. Most of the auth related functionality is pulled from the Thrift backend. All of the retry-related tests are expected to pass since they represent user-facing functionality (in terms of exceptions to be thrown in particular cases). 

We continue using a low-level connection pool instead of using the higher level `requests` module because the latter does not have first-class support for passing a password to a client certificate, which we want to support. There are workarounds [involving subclassing the `HttpAdapter` and invoking `init_poolmanager`](https://github.com/psf/requests/issues/2519#issuecomment-493762086), but the docstring of the method states that it should not be invoked from client code, so we do not consider it a long term solution. 

The exceptions thrown by the SEA and Thrift backends are not completely the same (yet), because this PR passes `None` to the context and session id of the `RequestError` object. This misparity will be logged and addressed later. Here, the context is expected to have the following keys as per the existing class definition: 
```
class RequestError(OperationalError):
    """Thrown if there was a error during request to the server.
    Its context will have the following keys:
    "method": The RPC method name that failed
    "session-id": The Thrift session guid
    "query-id": The Thrift query guid (if available)
    "http-code": HTTP response code to RPC request (if available)
    "error-message": Error message from the HTTP headers (if available)
    "original-exception": The Python level original exception
    "no-retry-reason": Why the request wasn't retried (if available)
    "bounded-retry-delay": The maximum amount of time an error will be retried before giving up
    "attempt": current retry number / maximum number of retries
    "elapsed-seconds": time that has elapsed since first attempting the RPC request
    """

    pass
```

### How?

In `_make_request()`, the client sets up the retry context before making the HTTP request:
- Determines the operation type (EXECUTE_STATEMENT, GET_OPERATION_STATUS, etc.) based on the API path and HTTP method
- Sets the command type on the DatabricksRetryPolicy - this influences retry decisions (e.g., EXECUTE_STATEMENT is only retried for 429/503 status codes)
- Starts the retry timer to track total request duration across all retry attempts
- Makes the HTTP request through urllib3 with the retry policy attached

The retry policy then handles all retry logic automatically - determining when to retry based on status codes and command types, calculating backoff delays, and enforcing attempt/duration limits. If retries are exhausted, urllib3 raises a MaxRetryError which gets caught and wrapped in a RequestError.

## How is this tested?

- [ ] Unit tests
- [x] E2E Tests
- [ ] Manually
- [ ] N/A


src/databricks/sql/auth/retry.py
```diff
@@ -123,15 +123,15 @@ def __init__(
...
             total=_attempts_remaining,
             respect_retry_after_header=True,
             backoff_factor=self.delay_min,
             allowed_methods=["POST"],
             status_forcelist=[429, 503, *self.force_dangerous_codes],
         )
-
-        urllib3_kwargs.update(**_urllib_kwargs_we_care_about)
+        _urllib_kwargs_we_care_about.update(**urllib3_kwargs)
+        urllib3_kwargs = _urllib_kwargs_we_care_about
```
jayantsing-db: this introduces a change in existing behaviour. This seems to be overriding `_urllib_kwargs_we_care_about` which was not the case earlier

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -38,72 +54,172 @@ def __init__(
...
+            "_retry_stop_after_attempts_duration", 900.0
+        )
+        self._retry_delay_default = kwargs.get("_retry_delay_default", 5.0)
+        self.force_dangerous_codes = kwargs.get("_retry_dangerous_codes", [])
+
+        # Connection pooling settings
+        self.max_connections = kwargs.get("max_connections", 10)
+
+        # Setup retry policy
+        self.enable_v3_retries = kwargs.get("_enable_v3_retries", True)
```
jayantsing-db, line 95: what is the purpose of this here?

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -38,72 +54,172 @@ def __init__(
...
+                delay_min=self._retry_delay_min,
+                delay_max=self._retry_delay_max,
+                stop_after_attempts_count=self._retry_stop_after_attempts_count,
+                stop_after_attempts_duration=self._retry_stop_after_attempts_duration,
+                delay_default=self._retry_delay_default,
+                force_dangerous_codes=self.force_dangerous_codes,
+                urllib3_kwargs=urllib3_kwargs,
+            )
+        else:
+            # Legacy behavior - no automatic retries
```
jayantsing-db, line 117: this means the backend has to take care of retrying? could you add reference to thrift code where there is a similar case?

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -38,72 +54,172 @@ def __init__(
...
+                "Legacy retry behavior is enabled for this connection."
+                " This behaviour is not supported for the SEA backend."
+            )
+            self.retry_policy = 0
 
-        # Create a session for connection pooling
-        self.session = requests.Session()
+        # Handle proxy settings
+        try:
+            proxy = urllib.request.getproxies().get(self.scheme)
```
jayantsing-db, line 128: add comment on how this picks up proxy config from env?

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -38,72 +54,172 @@ def __init__(
...
+        except (KeyError, AttributeError):
+            proxy = None
+        else:
+            if self.host and urllib.request.proxy_bypass(self.host):
```
jayantsing-db, line 134: add comment on the scenario here

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -38,72 +54,172 @@ def __init__(
...
+        # Initialize connection pool
+        self._pool = None
+        self._open()
+
+    def _basic_proxy_auth_headers(self, proxy_parsed) -> Optional[Dict[str, str]]:
+        """Create basic auth headers for proxy if credentials are provided."""
+        if proxy_parsed is None or not proxy_parsed.username:
+            return None
+        ap = f"{urllib.parse.unquote(proxy_parsed.username)}:{urllib.parse.unquote(proxy_parsed.password)}"
+        return make_headers(proxy_basic_auth=ap)
```
jayantsing-db, line 160: are there existing tests for proxy?

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -38,72 +54,172 @@ def __init__(
...
+                {
+                    "cert_reqs": ssl.CERT_REQUIRED
+                    if self.ssl_options.tls_verify
+                    else ssl.CERT_NONE,
+                    "ca_certs": self.ssl_options.tls_trusted_ca_file,
+                    "cert_file": self.ssl_options.tls_client_cert_file,
+                    "key_file": self.ssl_options.tls_client_cert_key_file,
+                    "key_password": self.ssl_options.tls_client_cert_key_password,
+                }
+            )
```
jayantsing-db, line 180: is this code motivated by the existing thrift client code?

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -112,75 +228,77 @@ def _make_request(
...
-            response = call(
-                url=url,
-                headers=headers,
-                json=data,
-                params=params,
-            )
+        # Set command type for retry policy
+        command_type = self._get_command_type_from_path(path, method)
+        self.set_retry_command_type(command_type)
+        self.start_retry_timer()
```
jayantsing-db, line 254: could you document the retry flow and working for this client in PR desc?

jayantsing-db (commented):
Few comments

jayantsing-db:
>The exceptions thrown by the SEA and Thrift backends are not completely the same (yet), because this PR passes None to the context and session id of the RequestError object. This misparity will be logged and addressed later.

Could you add details on this? Is there a follow-up PR? what is the context and session id?

src/databricks/sql/backend/sea/utils/http_client.py
varun-edachali-dbx: There is [a test](https://github.com/databricks/databricks-sql-python/blob/59d28b0d28a12776e87b0c8b7ce3e7fed280db95/tests/e2e/common/retry_test_mixins.py#L433) that ensures we log a certain warning in case we're using legacy retries (i.e. `enable_v3_retries` is `False`). This test helps catch that.

src/databricks/sql/backend/sea/utils/http_client.py
varun-edachali-dbx: Yes, exactly. It has effectively been picked up from the existing Thrift code.

src/databricks/sql/auth/retry.py
varun-edachali-dbx: The primary goal is to force `urllib` to respect `GET` and `DELETE` requests (it is currently only configured for `POST`). I've replaced this aggressive over-riding with some code that simply adds the above two methods to the `allowed_methods`.

src/databricks/sql/backend/sea/utils/http_client.py
varun-edachali-dbx: There do not seem to be any e2e tests in the repo that test proxy use.

src/databricks/sql/backend/sea/utils/http_client.py
varun-edachali-dbx: Yes, exactly - if we don't enable v3 retries then the backend has to take care of retrying. The `DatabricksRetryPolicy` is only instantiated in the Thrift backend if v3 retries are enabled, and the [retries are only triggered](https://github.com/databricks/databricks-sql-python/blob/fe8cd576f74698c64c77d4de08696b7b8081ba0b/src/databricks/sql/backend/thrift_backend.py#L406) in this case as well. Otherwise, the backend [has to "manually" take care ](https://github.com/databricks/databricks-sql-python/blob/fe8cd576f74698c64c77d4de08696b7b8081ba0b/src/databricks/sql/backend/thrift_backend.py#L498)of retrying.

src/databricks/sql/auth/retry.py
```diff
@@ -127,7 +127,7 @@ def __init__(
             total=_attempts_remaining,
             respect_retry_after_header=True,
             backoff_factor=self.delay_min,
-            allowed_methods=["POST"],
+            allowed_methods=["POST", "GET", "DELETE"],
```
jayantsing-db, line 130: thrift RPCs are all post?

jayantsing-db (approved):
LGTM

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -14,10 +23,17 @@ class SeaHttpClient:
...
-    This client handles the HTTP communication with the SEA endpoints,
-    including authentication, request formatting, and response parsing.
+    This client uses urllib3 for robust HTTP communication with retry policies
+    and connection pooling.
     """
 
+    retry_policy: Union[DatabricksRetryPolicy, int]
+    _pool: Optional[Union[HTTPConnectionPool, HTTPSConnectionPool]]
+    proxy_uri: Optional[str]
+    realhost: Optional[str]
```
jayantsing-db: this is specific to proxy case. should this be renamed to proxy_host?

src/databricks/sql/auth/retry.py
varun-edachali-dbx: Yes, this can be seen in the Thrift HTTP Client where [POST is hardcoded](https://github.com/databricks/databricks-sql-python/blob/aee6863199414281589be9815f84151de82b7347/src/databricks/sql/auth/thrift_http_client.py#L186)

PR #621, opened 2025-06-30
[PECOBLR-587] Azure Service Principal Credential Provider

jprakash-db:
## Description
This pull request introduces support for Azure Service Principal M2M (SP) authentication to the PySQL Connector

## Key Changes
- AzureServicePrincipalCredentialProvider, this is the provider that will deal with getting the creds using the azure credentials
- ClientCredentialsTokenSource, is responsible for managing the token lifecycle for the flow where credentials are obtained using the `grant_type: client_credentials`
- Introduced a common HTTP client called DatabricksHttpClient that can be used to unify the Http logic across the connector, to ensure standard client level behaviour

## New dependencies
- Introduced library [PyJWT](https://pypi.org/project/PyJWT/) which is required for handling JWT parsing functionalities

## Tests
Expanded unit tests to cover:
- AzureServicePrincipalCredentialProvider.
- ClientCredentialsTokenSource 

Manual Testing
- Tested the Azure M2M SP flow by creating a SP using Azure Entra ID and then mapping it to a workspace SP.  Further queries were made to the workspace using the Azure Entra ID credentials to test it on Databricks Workspace

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -816,8 +816,6 @@ class Cursor:
                 transformed_operation, normalized_parameters, param_structure
             )
 
-        print(f"Prepared operation: {prepared_operation}")
-        print(f"Prepared parameters: {prepared_params}")
         self._check_not_closed()
         self._close_and_clear_active_result_set()
         execute_response = self.thrift_backend.execute_command(


pyproject.toml
```diff
@@ -20,16 +20,18 @@ requests = "^2.18.1"
...
 ]
-python-dateutil = "^2.8.0"
+pyjwt = "^2.0.0"
+
 
 [tool.poetry.extras]
 pyarrow = ["pyarrow"]
 
-[tool.poetry.dev-dependencies]
+[tool.poetry.group.dev.dependencies]
```
vikrantpuppala, line 34: does the poetry update have to go in the same PR?

src/databricks/sql/auth/auth.py
```diff
@@ -103,9 +113,15 @@ def get_client_id_and_redirect_port(use_azure_auth: bool):
...
     auth_type = kwargs.get("auth_type")
-    (client_id, redirect_port_range) = get_client_id_and_redirect_port(
-        auth_type == AuthType.AZURE_OAUTH.value
-    )
+    client_id = kwargs.get("oauth_client_id")
+    redirect_port_range = kwargs.get("oauth_redirect_port_range")
+
+    if auth_type == AuthType.AZURE_SP_M2M.value:
+        pass
+    else:
```
vikrantpuppala: why not just do if `if auth_type != AuthType.AZURE_SP_M2M.value:`

src/databricks/sql/auth/auth.py
```diff
@@ -119,9 +135,12 @@ def get_python_sql_connector_auth_provider(hostname: str, **kwargs):
...
         tls_client_cert_file=kwargs.get("_tls_client_cert_file"),
         oauth_scopes=PYSQL_OAUTH_SCOPES,
-        oauth_client_id=kwargs.get("oauth_client_id") or client_id,
+        oauth_client_id=client_id,
+        oauth_client_secret=kwargs.get("oauth_client_secret"),
+        azure_tenant_id=kwargs.get("azure_tenant_id"),
+        azure_workspace_resource_id=kwargs.get("azure_workspace_resource_id"),
         oauth_redirect_port_range=[kwargs["oauth_redirect_port"]]
-        if kwargs.get("oauth_client_id") and kwargs.get("oauth_redirect_port")
+        if client_id and kwargs.get("oauth_redirect_port")
```
vikrantpuppala: this is behaviour change from before i.e. earlier we required client_id to come in from kwargs and now it is ok even if we derive it from get_client_id_and_redirect_port, is this intended?

src/databricks/sql/auth/auth.py
```diff
@@ -24,6 +18,9 @@ def __init__(
         auth_type: Optional[str] = None,
         oauth_scopes: Optional[List[str]] = None,
         oauth_client_id: Optional[str] = None,
+        oauth_client_secret: Optional[str] = None,
+        azure_tenant_id: Optional[str] = None,
```
samikshya-db: Can we add a todo to remove the need of passing tenant id ? https://github.com/databricks/databricks-sdk-py/pull/638

Also see : https://databricks.atlassian.net/browse/PECOBLR-212

src/databricks/sql/auth/authenticators.py
```diff
@@ -1,15 +1,21 @@
...
+    ClientCredentialsTokenSource,
+)
+from databricks.sql.auth.endpoint import get_oauth_endpoints
+from databricks.sql.auth.common import AuthType, get_effective_azure_login_app_id
 
 # Private API: this is an evolving interface and it will change in the future.
 # Please must not depend on it in your applications.
 from databricks.sql.experimental.oauth_persistence import OAuthToken, OAuthPersistence
 
+logger = logging.getLogger(__name__)
```
vikrantpuppala: don't think this is being used, can we add logging?

src/databricks/sql/auth/common.py
```diff
@@ -0,0 +1,28 @@
...
+        ".dev.azuredatabricks.net": "62a912ac-b58e-4c1d-89ea-b2dbfc7358fc",
+        ".staging.azuredatabricks.net": "4a67d088-db5c-48f1-9ff2-0aace800ae68",
+    }
+
+    for domain, app_id in azure_app_ids.items():
+        if domain in hostname:
+            return app_id
+
+    # default databricks resource id
+    return "2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"
```
vikrantpuppala: can we add these IDs as constants at the top so they're in one place

src/databricks/sql/auth/oauth.py
```diff
@@ -6,19 +6,63 @@
...
+
+
+class RefreshableTokenSource(ABC):
+    @abstractmethod
+    def get_token(self) -> Token:
+        pass
+
+    @abstractmethod
+    def refresh(self) -> Token:
+        pass
```
vikrantpuppala, line 63: can we add a comment here that we have duplicate code here with the sdk (with the code pointer) and in the long term we should try to unify?

src/databricks/sql/auth/oauth.py
```diff
@@ -258,3 +302,63 @@ def get_tokens(self, hostname: str, scope=None):
...
+        self.extra_params = extra_params
+        self.token: Optional[Token] = None
+        self._http_client = DatabricksHttpClient.get_instance()
+
+    def get_token(self) -> Token:
+        if self.token is None or self.token.is_expired():
+            self.token = self.refresh()
+        return self.token
+
+    def refresh(self) -> Token:
```
vikrantpuppala, line 337: how is the refresh mechanism being handled for the existing credential providers? is there opportunity to dedup/reuse?

src/databricks/sql/common/http.py
```diff
@@ -0,0 +1,65 @@
...
+    expires_on: int = 0
+    not_before: int = 0
+    resource: str = ""
+    access_token: str = ""
+    refresh_token: str = ""
+
+
+# Singleton class for common Http Client
+class DatabricksHttpClient:
+    ## TODO: Unify all the http clients in the PySQL Connector
```
vikrantpuppala, line 43: why can't this be done right now? can we not use the existing http client?

src/databricks/sql/auth/auth.py
```diff
@@ -1,19 +1,13 @@
-from enum import Enum
 from typing import Optional, List
 
 from databricks.sql.auth.authenticators import (
     AuthProvider,
     AccessTokenAuthProvider,
     ExternalAuthProvider,
     DatabricksOAuthProvider,
+    AzureServicePrincipalCredentialProvider,
```
gopalldb, line 8: is this coming from sdk?

src/databricks/sql/auth/auth.py
```diff
@@ -45,7 +45,17 @@ def __init__(
...
     if cfg.credentials_provider:
         return ExternalAuthProvider(cfg.credentials_provider)
-    if cfg.auth_type in [AuthType.DATABRICKS_OAUTH.value, AuthType.AZURE_OAUTH.value]:
+    elif cfg.auth_type == AuthType.AZURE_SP_M2M.value:
+        return ExternalAuthProvider(
+            AzureServicePrincipalCredentialProvider(
+                cfg.hostname,
+                cfg.oauth_client_id,
+                cfg.oauth_client_secret,
+                cfg.azure_tenant_id,
```
gopalldb, line 22: We were talking about this in JDBC if tenant ID can be derived dynamically and not be a required config

src/databricks/sql/auth/oauth.py
```diff
@@ -6,19 +6,63 @@
...
+            decoded_token = jwt.decode(
+                self.access_token, options={"verify_signature": False}
+            )
+            exp_time = decoded_token.get("exp")
+            current_time = time.time()
+            buffer_time = 30  # 30 seconds buffer
+            return exp_time and (exp_time - buffer_time) <= current_time
+        except Exception as e:
+            logger.error("Failed to decode token: %s", e)
+            return e
```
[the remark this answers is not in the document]
gopalldb: Fix this

src/databricks/sql/common/http.py
```diff
@@ -0,0 +1,65 @@
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+from enum import Enum
+import threading
+from dataclasses import dataclass
+
+# Enums for HTTP Methods
+class HttpMethod(str, Enum):
+    GET = "GET"
```
gopalldb, line 16: there are no existing enums for these?

src/databricks/sql/common/http.py
```diff
@@ -0,0 +1,65 @@
+import requests
```
samikshya-db, line 1: @varun-edachali-dbx  has added a http client already, I suggest you two collaborate to push the common http to main first https://github.com/databricks/databricks-sql-python/blob/sea-migration/src/databricks/sql/backend/sea/utils/http_client.py

samikshya-db (commented):
I don't have I have much objection with this implementation. But, what would it take for us to include dependency on SDK as it has the support for this already?

pyproject.toml
jprakash-db: It was just a minor update

src/databricks/sql/auth/auth.py
jprakash-db: No, the changes are inspired from SDK

src/databricks/sql/auth/auth.py
jprakash-db: Fixed this

src/databricks/sql/auth/oauth.py
jprakash-db: added it

src/databricks/sql/auth/oauth.py
jprakash-db: In the existing setup only U2M browser Auth is supported and all functions are hard coded for its needs. The long term strategy is to unify this using SDK

src/databricks/sql/common/http.py
jprakash-db: No

src/databricks/sql/common/http.py
jprakash-db: Currently the http client is a thrift http client. Unifying both of them can be achieved but will require a separate PR

src/databricks/sql/auth/auth.py
jprakash-db: fixed

src/databricks/sql/common/http.py
jprakash-db: Will undertake the final merge when SEA is ready to be merged

src/databricks/sql/auth/common.py
```diff
@@ -0,0 +1,96 @@
...
+        if domain in hostname:
+            return app_id
+
+    # default databricks resource id
+    return AzureAppId.PROD.value[1]
+
+
+def get_azure_tenant_id_from_host(host: str, http_client=None) -> str:
+    """
+    Load the Azure tenant ID from the Azure Databricks login page.
```
vikrantpuppala, line 74: can you make this more detailed on how you're getting the tenant id from the host?

samikshya-db (approved):

PR #622, opened 2025-07-01
Make telemetry batch size configurable and add time-based flush

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
The flush timer is centralized in `TelemetryClientFactory`, single background thread to manage all connections. Keeping it in `TelemetryClient` would mean creating a timer thread per connection.

Used `threading.Thread` with `threading.Event`. The `threading.Event` acts as a thread-safe shutdown signal, and its `wait(timeout)` method allows the thread to wait for the next flush interval while remaining immediately responsive to a shutdown command.
While `threading.Timer` could be used, it would create a new thread every flush interval as we need to create a timer after each execution.

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

[design doc](https://docs.google.com/document/d/1LBhEd6sJL66hf-4L0HThtvp1pb_iCS8Wcp3Yf5zR4h8/edit?tab=t.0#heading=h.dcuodgmlkfek)
PECOBLR-654

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -254,6 +254,9 @@ class Connection:
         self.telemetry_enabled = (
             self.client_telemetry_enabled and self.server_telemetry_enabled
         )
+        self.telemetry_batch_size = kwargs.get(
+            "telemetry_batch_size", TelemetryClientFactory.DEFAULT_BATCH_SIZE
+        )
 
         try:
             self.session = Session(
@@ -290,6 +293,7 @@ class Connection:
             session_id_hex=self.get_session_id_hex(),
             auth_provider=self.session.auth_provider,
             host_url=self.session.host,
+            batch_size=self.telemetry_batch_size,
         )
 
         self._telemetry_client = TelemetryClientFactory.get_telemetry_client(
--- a/src/databricks/sql/telemetry/telemetry_client.py
+++ b/src/databricks/sql/telemetry/telemetry_client.py
@@ -138,8 +138,6 @@ class TelemetryClient(BaseTelemetryClient):
     TELEMETRY_AUTHENTICATED_PATH = "/telemetry-ext"
     TELEMETRY_UNAUTHENTICATED_PATH = "/telemetry-unauth"
 
-    DEFAULT_BATCH_SIZE = 100
-
     def __init__(
         self,
         telemetry_enabled,
@@ -147,10 +145,11 @@ class TelemetryClient(BaseTelemetryClient):
         auth_provider,
         host_url,
         executor,
+        batch_size,
     ):
         logger.debug("Initializing TelemetryClient for connection: %s", session_id_hex)
         self._telemetry_enabled = telemetry_enabled
-        self._batch_size = self.DEFAULT_BATCH_SIZE
+        self._batch_size = batch_size
         self._session_id_hex = session_id_hex
         self._auth_provider = auth_provider
         self._user_agent = None
@@ -318,7 +317,7 @@ class TelemetryClient(BaseTelemetryClient):
 class TelemetryClientFactory:
     """
     Static factory class for creating and managing telemetry clients.
-    It uses a thread pool to handle asynchronous operations.
+    It uses a thread pool to handle asynchronous operations and a single flush thread for all clients.
     """
 
     _clients: Dict[
@@ -331,6 +330,13 @@ class TelemetryClientFactory:
     _original_excepthook = None
     _excepthook_installed = False
 
+    # Shared flush thread for all clients
+    _flush_thread = None
+    _flush_event = threading.Event()
+    _flush_interval_seconds = 90
+
+    DEFAULT_BATCH_SIZE = 100
+
     @classmethod
     def _initialize(cls):
         """Initialize the factory if not already initialized"""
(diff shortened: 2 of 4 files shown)


src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -141,10 +141,13 @@ def __init__(
         auth_provider,
         host_url,
         executor,
+        batch_size=None,
```
vikrantpuppala: can we put the default here instead

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -367,6 +403,7 @@ def initialize_telemetry_client(
         session_id_hex,
         auth_provider,
         host_url,
+        batch_size=None,
```
vikrantpuppala: do we need default arg?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Setting the default in `TelemetryClient.__init__` failed because if a user doesn't provide the `telemetry_batch_size` argument to `Connection`, `kwargs.get()` passes `None` down the call chain, overriding any default in the `TelemetryClient` constructor. Have changed it to set default in `Connection.__init__`, ensuring `TelemetryClientFactory` and `TelemetryClient` always receive an explicit integer value.

src/databricks/sql/client.py
```diff
@@ -273,6 +276,7 @@ def read(self) -> Optional[OAuthToken]:
             session_id_hex=self.get_session_id_hex(),
             auth_provider=self.session.auth_provider,
             host_url=self.session.host,
+            batch_size=TelemetryClientFactory.DEFAULT_BATCH_SIZE,
```
vikrantpuppala: shouldn't this be self.telemetry_batch_size?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -334,11 +340,39 @@ def _initialize(cls):
...
             cls._initialized = True
             logger.debug(
                 "TelemetryClientFactory initialized with thread pool (max_workers=10)"
             )
 
+    @classmethod
+    def _start_flush_thread(cls):
+        """Start the shared background thread for periodic flushing of all clients"""
+        cls._flush_event.clear()
+        cls._flush_thread = threading.Thread(target=cls._flush_worker, daemon=True)
```
vikrantpuppala, line 360: why a daemon thred?

tests/unit/test_telemetry.py
```diff
@@ -276,6 +280,7 @@ def test_factory_shutdown_flow(self):
                 session_id_hex=session,
                 auth_provider=AccessTokenAuthProvider("token"),
                 host_url="test-host.com",
+                batch_size=TelemetryClientFactory.DEFAULT_BATCH_SIZE
```
vikrantpuppala, line 282: do we have a test to check for non-default batch size being propogated correctly?

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: A daemon thread is used for the background telemetry flusher so it doesn't prevent the main application from exiting. Without it, Python would wait for the thread's long sleep interval to finish before the program can shut down, making the user's script appear to hang. This ensures the main program exits immediately as expected.

src/databricks/sql/client.py
saishreeeee: yes my bad I will change that

tests/unit/test_telemetry.py
saishreeeee: I have checked it manually, will add a test too.

src/databricks/sql/telemetry/telemetry_client.py
vikrantpuppala: My concern here:
```
Daemon threads can be killed mid-operation when the main program exits, which might lead to data loss or corruption.
```
which means that this might not perform the flush if the main program is killed?

saishreeeee: That is already handled by the `exception_hook`. The function `_handle_unhandled_exception` flushes all the Telemetry clients.

vikrantpuppala (approved):

tests/unit/test_telemetry.py
```diff
@@ -258,6 +261,7 @@ def test_factory_error_handling(self):
                 session_id_hex=session_id,
                 auth_provider=AccessTokenAuthProvider("token"),
                 host_url="test-host.com",
+                batch_size=TelemetryClientFactory.DEFAULT_BATCH_SIZE
```
jprakash-db, line 263: why do you need to pass the default value ? Inside the function add the default value, pass it only when users sends a new value

tests/unit/test_telemetry.py
```diff
@@ -290,3 +295,37 @@ def test_factory_shutdown_flow(self):
         TelemetryClientFactory.close(session2)
         assert TelemetryClientFactory._initialized is False
         assert TelemetryClientFactory._executor is None
+
+
+@patch("databricks.sql.client.Session")
+def test_connection_with_custom_batch_size(MockSession):
+    """
```
jprakash-db: Don't think this test is needed. You are just setting a value and checking if it is set

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -334,11 +340,39 @@ def _initialize(cls):
...
             cls._install_exception_hook()
+            cls._start_flush_thread()
             cls._initialized = True
             logger.debug(
                 "TelemetryClientFactory initialized with thread pool (max_workers=10)"
             )
 
+    @classmethod
+    def _start_flush_thread(cls):
+        """Start the shared background thread for periodic flushing of all clients"""
```
jprakash-db, line 358: Just a thought is this background thread and its overhead required and is the timely refresh feature needed ? Like anyway we send the final telemtry logs while shutdown, and  we can reduce the default batch size to have a near similar effect without the tbackground thread overload cc @vikrantpuppala

tests/unit/test_telemetry.py
saishreeeee: We set the default value of `telemetry_batch_size` if required during extraction from kwargs in connection initialization, so I did not set a default value for `batch_size` in the `initialize_telemetry_client` function in the `TelemetryClientFactory` class.

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: This is a pattern followed in telemetry in the JDBC driver.

tests/unit/test_telemetry.py
saishreeeee: removed it

jprakash-db (approved):

PR #624, opened 2025-07-04
Fix potential resource leak in `CloudFetchQueue`

varun-edachali-dbx:
## What type of PR is this?
- [x] Bug Fix

## Description

The `ResultFileDownloadManager` in the `CloudFetchQueue` is not shut down alongside the `ResultSet`.

The `ResultFileDownloadManager` on being requested for a file ensures that the "download queue is always full", i.e. it adds upto `max_download_threads` Futures into `_download_tasks`. In case `ResultSet.close()` is invoked after this, `_shutdown_manager` (defined for `ResultFileDownloadManager`) is NOT called, so these Futures keep executing in the background and the files keep being downloaded. 

We define a `close` for the `CloudFetchQueue` that invokes `_shutdown_manager`.

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

N/A


src/databricks/sql/client.py
```diff
@@ -1509,6 +1509,7 @@ def close(self) -> None:
             if isinstance(e.args[1], CursorAlreadyClosedError):
                 logger.info("Operation was canceled by a prior request")
         finally:
+            self.results.close()
```
jprakash-db: nit: Can you move the `self.results.close()`, within the try block. - Logical step will be first close the resultSet and then close the command on the server. Secondly don't want any error in the finally part, this will lead to op_state variables in unset state

jprakash-db (approved):
LGTM, Thanks for making the changes

src/databricks/sql/client.py
```diff
@@ -1596,6 +1596,7 @@ def close(self) -> None:
         been closed on the server for some other reason, issue a request to the server to close it.
         """
         try:
+            self.results.close()
```
jayantsing-db, line 1599: this might be not possible in this PR. I think ideally cursor should just close contained result sets. result set close should then close the queues and commands. Maybe in SEA flow, we should keep this in mind.

jayantsing-db (approved):

PR #628, opened 2025-07-07
Add optional telemetry support to the python connector

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Initial telemetry logs, failure logs, latency logs via telemetry

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

[Design doc](https://docs.google.com/document/d/1LBhEd6sJL66hf-4L0HThtvp1pb_iCS8Wcp3Yf5zR4h8/edit?tab=t.0)
PECOBLR-585

### DIFF
--- a/src/databricks/sql/telemetry/telemetry_client.py
+++ b/src/databricks/sql/telemetry/telemetry_client.py
@@ -401,11 +401,9 @@ class TelemetryClientFactory:
     @staticmethod
     def get_telemetry_client(session_id_hex):
         """Get the telemetry client for a specific connection"""
-        try:
-            return TelemetryClientFactory._clients[session_id_hex]
-        except Exception as e:
-            logger.debug("Failed to get telemetry client: %s", e)
-            return NoopTelemetryClient()
+        return TelemetryClientFactory._clients.get(
+            session_id_hex, NoopTelemetryClient()
+        )
 
     @staticmethod
     def close(session_id_hex):


src/databricks/sql/exc.py
```diff
@@ -1,21 +1,32 @@
 import json
 import logging
 
-logger = logging.getLogger(__name__)
+from databricks.sql.telemetry.telemetry_client import TelemetryClientFactory
 
+logger = logging.getLogger(__name__)
```
vikrantpuppala: don't think we're using this?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,231 @@
...
+            - CursorExtractor for Cursor objects
+            - ResultSetExtractor for ResultSet objects
+            - Throws an NotImplementedError for all other objects
+    """
+    if obj.__class__.__name__ == "Cursor":
+        return CursorExtractor(obj)
+    elif obj.__class__.__name__ == "ResultSet":
+        return ResultSetExtractor(obj)
+    else:
+        raise NotImplementedError(f"No extractor found for {obj.__class__.__name__}")
```
vikrantpuppala: will this error break anything or just be printed in debug mode?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(self, *args, **kwargs):
+            start_time = time.perf_counter()
+            result = None
+            try:
+                result = func(self, *args, **kwargs)
+                return result
+            finally:
```
vikrantpuppala, line 186: what if anything in this finally block fails?

src/databricks/sql/telemetry/models/enums.py
```diff
@@ -0,0 +1,42 @@
...
+    FORMAT_UNSPECIFIED = "format_unspecified"
+    INLINE_ARROW = "inline_arrow"
+    EXTERNAL_LINKS = "external_links"
+    COLUMNAR_INLINE = "columnar_inline"
+
+
+class StatementType(Enum):
+    NONE = "none"
+    QUERY = "query"
+    SQL = "sql"
```
vikrantpuppala: query type is used when sql command is expected to return results, in jdbc it makes sense due to the nature of interface, but do we need both query and sql in python connector?

src/databricks/sql/telemetry/models/enums.py
```diff
@@ -0,0 +1,42 @@
...
+class AuthFlow(Enum):
+    TOKEN_PASSTHROUGH = "token_passthrough"
+    BROWSER_BASED_AUTHENTICATION = "browser_based_authentication"
+
+
+class AuthMech(Enum):
+    CLIENT_CERT = "CLIENT_CERT"  # ssl certificate authentication
+    PAT = "PAT"  # Personal Access Token authentication
+    DATABRICKS_OAUTH = "DATABRICKS_OAUTH"  # Databricks-managed OAuth flow
+    EXTERNAL_AUTH = "EXTERNAL_AUTH"  # External identity provider (AWS, Azure, etc.)
```
vikrantpuppala: I think @jprakash-db is adding some enums in his latest PR #621, is there scope to re-use?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,439 @@
...
+        executor,
+    ):
+        logger.debug("Initializing TelemetryClient for connection: %s", session_id_hex)
+        self._telemetry_enabled = telemetry_enabled
+        self._batch_size = 10  # TODO: Decide on batch size
+        self._session_id_hex = session_id_hex
+        self._auth_provider = auth_provider
+        self._user_agent = None
+        self._events_batch = []
+        self._lock = threading.Lock()
```
vikrantpuppala: do we not need RLock here?

src/databricks/sql/exc.py
saishreeeee: yes will remove it

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: changing it to log in debug mode, and added an `if extractor not None` check before exporting the latency log

src/databricks/sql/telemetry/models/enums.py
saishreeeee: the `execute` and `execute_async` functions have `StatementType` set as query and functions like `_handle_staging_get`, `_handle_staging_put` and `_handle_staging_remove` have it set as sql.

src/databricks/sql/client.py
```diff
@@ -234,6 +246,12 @@ def read(self) -> Optional[OAuthToken]:
             server_hostname, **kwargs
         )
 
+        self.server_telemetry_enabled = True
+        self.client_telemetry_enabled = kwargs.get("enable_telemetry", False)
```
jprakash-db, line 250: Isn't telemetry enabled by default and the user needs to manually opt out right ? cc @vikrantpuppala

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,237 @@
...
+class TelemetryExtractor:
+    """
+    Base class for extracting telemetry information from various object types.
+
+    This class serves as a proxy that delegates attribute access to the wrapped object
+    while providing a common interface for extracting telemetry-related data.
+    """
+
+    def __init__(self, obj):
+        """
```
jprakash-db: __init__ and __get_attr__ don't need python doc, they are standard functions

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -0,0 +1,237 @@
...
+            - CursorExtractor for Cursor objects
+            - ResultSetExtractor for ResultSet objects
+            - None for all other objects
+    """
+    if obj.__class__.__name__ == "Cursor":
+        return CursorExtractor(obj)
+    elif obj.__class__.__name__ == "ResultSet":
+        return ResultSetExtractor(obj)
+    else:
+        logger.error(f"No extractor found for {obj.__class__.__name__}")
```
jprakash-db: nit: follow lazy logging

src/databricks/sql/telemetry/models/endpoint_models.py
```diff
@@ -0,0 +1,43 @@
+import json
+from dataclasses import dataclass, asdict
+from typing import List, Optional
+
+
+@dataclass
+class TelemetryRequest:
```
jprakash-db: Why is JsonSerializableMixin not used here ?

src/databricks/sql/telemetry/models/endpoint_models.py
```diff
@@ -0,0 +1,43 @@
...
+    uploadTime: int
+    items: List[str]
+    protoLogs: Optional[List[str]]
+
+    def to_json(self):
+        return json.dumps(asdict(self))
+
+
+@dataclass
+class TelemetryResponse:
```
jprakash-db: Same comment: 
> Why is JsonSerializableMixin not used here ?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,437 @@
...
+class NoopTelemetryClient(BaseTelemetryClient):
+    """
+    NoopTelemetryClient is a telemetry client that does not send any events to the server.
+    It is used when telemetry is disabled.
+    """
+
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
```
jprakash-db, line 106: This is not thread safe, use a lock when dealing with a single instance creation. Use double-checked locking

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,437 @@
...
+            self._auth_provider.add_headers(headers)
+
+        try:
+            logger.debug("Submitting telemetry request to thread pool")
+            future = self._executor.submit(
+                requests.post,
+                url,
+                data=json.dumps(request),
+                headers=headers,
+                timeout=10,
```
jprakash-db: Any basis for this 10 seconds ?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def _telemetry_request_callback(self, future):
+        """Callback function to handle telemetry request completion"""
+        try:
+            response = future.result()
+
+            if response.status_code == 200:
+                logger.debug("Telemetry request completed successfully")
+            else:
+                logger.debug(
```
jprakash-db, line 222: log the error message also when non 200 status code

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,437 @@
...
+            logger.debug("Failed to initialize telemetry client: %s", e)
+            # Fallback to NoopTelemetryClient to ensure connection doesn't fail
+            TelemetryClientFactory._clients[session_id_hex] = NoopTelemetryClient()
+
+    @staticmethod
+    def get_telemetry_client(session_id_hex):
+        """Get the telemetry client for a specific connection"""
+        try:
+            if session_id_hex in TelemetryClientFactory._clients:
+                return TelemetryClientFactory._clients[session_id_hex]
```
jprakash-db: too verbose, just check if id present otherwise within exception return NoopClient

tests/unit/test_telemetry.py
```diff
@@ -0,0 +1,286 @@
...
+from databricks.sql.telemetry.models.enums import AuthMech, AuthFlow
+from databricks.sql.auth.authenticators import (
+    AccessTokenAuthProvider,
+    DatabricksOAuthProvider,
+    ExternalAuthProvider,
+)
+
+
+@pytest.fixture
+def telemetry_system_reset():
```
jprakash-db: This fixture is only being used in the `TestTelemetryFactory`, move it within that class. Also you can scope a fixture to run before every test, you don't need to pass it again and again in every function

tests/unit/test_telemetry.py
```diff
@@ -0,0 +1,286 @@
...
+        assert TelemetryHelper.get_auth_flow(pat_auth) is None
+        
+        # Test None auth provider
+        assert TelemetryHelper.get_auth_flow(None) is None
+
+
+class TestTelemetryFactory:
+    """Tests for TelemetryClientFactory lifecycle and management."""
+
+    def test_client_lifecycle_flow(self, telemetry_system_reset):
```
jprakash-db: telemetry_system_reset, fixture can be scoped to function, no need to pass it again and again

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -0,0 +1,437 @@
...
+                return AuthFlow.TOKEN_PASSTHROUGH
+            else:
+                return AuthFlow.BROWSER_BASED_AUTHENTICATION
+        elif isinstance(auth_provider, ExternalAuthProvider):
+            return AuthFlow.CLIENT_CREDENTIALS
+        else:
+            return None
+
+
+class BaseTelemetryClient(ABC):
```
jprakash-db: Can you move this ABC to the telemetry utils, because common interfaces etc should always be kept separate. This will prevent any circular import problems further down

src/databricks/sql/telemetry/models/enums.py
jprakash-db: yes, we can have a common AuthType variable

src/databricks/sql/client.py
vikrantpuppala: To start with, we are going to be opt in, this will be changed later

saishreeeee: It is disabled by default as of now.

src/databricks/sql/telemetry/models/endpoint_models.py
saishreeeee: I will add it.

src/databricks/sql/telemetry/models/endpoint_models.py
saishreeeee: Yes I will add it.

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Have changed it to 900. This the value used in thrift backend and in the JDBC driver.

jprakash-db (approved):

src/databricks/sql/telemetry/telemetry_client.py
saishreeeee: Have changed it return `TelemetryClientFactory._clients.get(session_id_hex, NoopTelemetryClient())

PR #631, opened 2025-07-08
SEA: add support for `Hybrid` disposition 

varun-edachali-dbx:
## What type of PR is this?
- [x] Feature

## Description
Following discussions surrounding the inefficiency of the existing `EXTERNAL_LINKS` disposition due to the repeated calls made to the control plane, we introduce support for the `Hybrid` disposition, which provides batches of links and prevents the need to query for every link. 

In case of small results, we get the response in the `attachment` key in the arrow format, which further boosts the efficiency. 

## How is this tested?
- [ ] Unit tests
- [x] E2E Tests
- [ ] Manually
- [ ] N/A

[Benchmarking of EXTERNAL_LINKS](https://docs.google.com/document/d/1MBeefNNR_854L94Mf5YAa8bjv1tz5behz2eFz_eES8A/edit?usp=sharing)
[PECOBLR-639](https://databricks.atlassian.net/browse/PECOBLR-639?atlOrigin=eyJpIjoiYzA1NTk0Yzc3NjNlNDcyMWJjMTU0YzE4NGE4ODg3NmUiLCJwIjoiaiJ9)

[PECOBLR-639]: https://databricks.atlassian.net/browse/PECOBLR-639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/databricks/sql/backend/sea/queue.py
+++ b/src/databricks/sql/backend/sea/queue.py
@@ -161,7 +161,6 @@ class SeaCloudFetchQueue(CloudFetchQueue):
         self._sea_client = sea_client
         self._statement_id = statement_id
         self._total_chunk_count = total_chunk_count
-        self._total_chunk_count = total_chunk_count
 
         logger.debug(
             "SeaCloudFetchQueue: Initialize CloudFetch loader for statement {}, total chunks: {}".format(


src/databricks/sql/backend/sea/models/responses.py
```diff
@@ -91,6 +92,11 @@ def _parse_result(data: Dict[str, Any]) -> ResultData:
                 )
             )
 
+    # Handle attachment field - decode from base64 if present
+    attachment = result_data.get("attachment")
+    if attachment is not None:
+        attachment = base64.b64decode(attachment)
+
```
jayantsing-db, line 99: nice! how did you deduce this and is this holistic?

src/databricks/sql/backend/sea/queue.py
```diff
@@ -62,6 +69,18 @@ def build_queue(
...
+                arrow_file = (
+                    ResultSetDownloadHandler._decompress_data(result_data.attachment)
+                    if lz4_compressed
+                    else result_data.attachment
+                )
+                arrow_table = create_arrow_table_from_arrow_file(
+                    arrow_file, description
+                )
+                logger.debug(f"Created arrow table with {arrow_table.num_rows} rows")
+                return ArrowQueue(arrow_table, manifest.total_row_count)
```
jayantsing-db, line 82: nice work reusing ArrowQueue. curious, how do we integrate metadata when we are using attachment? Although i think this is exactly the same behaviour as existing thrift client?

src/databricks/sql/backend/sea/queue.py
```diff
@@ -142,6 +161,7 @@ def __init__(
         self._sea_client = sea_client
         self._statement_id = statement_id
         self._total_chunk_count = total_chunk_count
+        self._total_chunk_count = total_chunk_count
```
jayantsing-db: repetitive?

jayantsing-db (approved):
Thanks. Some minor questions, rest LGTM

src/databricks/sql/backend/sea/models/responses.py
varun-edachali-dbx: This came as a side effect of trying to re-use the `ArrowQueue`. I realised that attachment was supposed to be raw bytes but it comes as a string in the JSON payload, so I explicitly serialised it to bytes. This code may not be necessary if we introduce an object mapper.

src/databricks/sql/backend/sea/queue.py
varun-edachali-dbx: I believe that the column metadata is integrated into the attachment. Simply de-serialising the data seems to create a typed arrow table. You're right that this is exactly the same behaviour as the existing Thrift client.

PR #633, opened 2025-07-10
SEA: Reduce network calls for synchronous commands

varun-edachali-dbx:
## What type of PR is this?
- [x] Refactor

## Description
In `execute_command` we first send the execute request to the server, following which, if the request is synchronous, we poll for the request state until it is no longer in the pending state. Following this, we make an additional `GET` request to the server to get the final request information. There are two areas of improvement: 
- if the request is small and completes execution (i.e., reaches the SUCCEEDED state) within the `wait_timeout`, then we need not poll for completion or make another `GET` request to the server. We can immediately construct our `ResultSet` with the provided response. 
- While we poll for the request state, if the state reaches SUCCEEDED then the response is accompanied by the response data that we need to construct the `ResultSet`. We need not make another `GET` request to the server after we are done polling and can instead utilise the last response provided. 

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

N/A

### DIFF
--- a/src/databricks/sql/backend/sea/backend.py
+++ b/src/databricks/sql/backend/sea/backend.py
@@ -6,7 +6,6 @@ import re
 from typing import Any, Dict, Tuple, List, Optional, Union, TYPE_CHECKING, Set
 
 from databricks.sql.backend.sea.models.base import ExternalLink, ResultManifest
-from databricks.sql.backend.sea.models.responses import GetStatementResponse
 from databricks.sql.backend.sea.utils.constants import (
     ALLOWED_SESSION_CONF_TO_DEFAULT_VALUES_MAP,
     ResultFormat,
@@ -43,6 +42,7 @@ from databricks.sql.backend.sea.models import (
     DeleteSessionRequest,
     StatementParameter,
     ExecuteStatementResponse,
+    GetStatementResponse,
     CreateSessionResponse,
 )
 from databricks.sql.backend.sea.models.responses import GetChunksResponse


src/databricks/sql/backend/sea/models/responses.py
```diff
@@ -124,26 +124,6 @@ def from_dict(cls, data: Dict[str, Any]) -> "ExecuteStatementResponse":
...
-    def from_dict(cls, data: Dict[str, Any]) -> "GetStatementResponse":
-        """Create a GetStatementResponse from a dictionary."""
-        return cls(
-            statement_id=data.get("statement_id", ""),
-            status=_parse_status(data),
-            manifest=_parse_manifest(data),
-            result=_parse_result(data),
-        )
-
-
```
jayantsing-db: I think you would still need this? Probably part of different PRs

src/databricks/sql/backend/sea/backend.py
```diff
@@ -358,6 +357,28 @@ def _results_message_to_execute_response(
...
 
+    def _response_to_result_set(
+        self, response: ExecuteStatementResponse, cursor: Cursor
+    ) -> SeaResultSet:
+        """
+        Convert a SEA response to a SeaResultSet.
+        """
+
+        # Create and return a SeaResultSet
+        from databricks.sql.backend.sea.result_set import SeaResultSet
```
jayantsing-db: does this lazy import has perf gain (lazily module loading when needed)?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -324,7 +323,7 @@ def _extract_description_from_manifest(
         return columns
 
     def _results_message_to_execute_response(
-        self, response: GetStatementResponse
+        self, response: ExecuteStatementResponse
```
jayantsing-db: Didn't know that GetStatementResponse and ExecuteStatementResponse have the same fields wrt results (interchangeable).

src/databricks/sql/backend/sea/backend.py
```diff
@@ -378,7 +399,7 @@ def _check_command_not_in_failed_or_closed_state(
 
     def _wait_until_command_done(
         self, response: ExecuteStatementResponse
-    ) -> CommandState:
+    ) -> ExecuteStatementResponse:
```
jayantsing-db: Seems odd to me that this method does polling and still ends up return ExecuteStatementResponse. Semantically, this is a response to ExecuteRequest

src/databricks/sql/backend/sea/backend.py
```diff
@@ -574,9 +591,25 @@ def get_query_state(self, command_id: CommandId) -> CommandState:
             path=self.STATEMENT_PATH_WITH_ID.format(sea_statement_id),
             data=request.to_dict(),
         )
+        response = ExecuteStatementResponse.from_dict(response_data)
```
jayantsing-db: Is it okay to return ExecuteResponse as a result of Polling?

jayantsing-db (commented):
Some questions inline.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Yes, I did not realise this at first either, but this can be confirmed by comparing the response in the REST reference as well: 
- `ExecuteStatementResponse`: https://docs.databricks.com/api/workspace/statementexecution/executestatement
- `GetStatementResponse`: https://docs.databricks.com/api/workspace/statementexecution/getstatement

This does make sense logically to me as well, the purpose of the GET is to get the info related to an execution statement.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Changed to `GetStatementResponse`

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: It could have a slight performance impact for the start-up time because we're deferring the import until we need it, but I do not think it is anything significant. 

In this case, I did it in order to resolve a circular dependency by deferring the import until run-time. However, I just realised that with some other refactoring I did this is no longer a concern. I have removed this lazy import.

PR #634, opened 2025-07-13
Chunk download latency

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Record chunk download latency
Added `chunk_id` to `SqlExecutionEvent` and `ResultDownloadHandler`

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

Ran the query: `SELECT * FROM RANGE(20000000)`
Latency was recorded for 4 chunks.
The latency log for the 1st chunk:
```
{
  "frontend_log_event_id": "e684ff49-9026-49af-abb6-687a6a16e565",
  "context": {
    "client_context": {
      "timestamp_millis": 1752467452769,
      "user_agent": "PyDatabricksSqlConnector/4.0.5"
    }
  },
  "entry": {
    "sql_driver_log": {
      "session_id": "01f0606b-4825-118d-abd2-87f5748c7dd3",
      "system_configuration": {
        "driver_version": "4.0.5",
        "os_name": "Darwin",
        "os_version": "24.5.0",
        "os_arch": "arm64",
        "runtime_version": "3.13.3",
        "runtime_vendor": "CPython",
        "char_set_encoding": "utf-8",
        "locale_name": "en_US"
      },
      "driver_connection_params": {
        "http_path": <HTTP_PATH>,
        "mode": "THRIFT",
        "host_info": {
          "host_url": <SERVER_HOSTNAME>,
          "port": 443
        },
        "auth_mech": "PAT"
      },
      "sql_statement_id": "01f0606b-4854-1f89-a0cb-d5860332ef81",
      "sql_operation": {
        "statement_type": "QUERY",
        "is_compressed": true,
        "execution_result": "EXTERNAL_LINKS",
        "chunk_id": 0
      },
      "operation_latency_ms": 15882
    }
  }
}
```

PECOBLR-653

### DIFF
--- a/src/databricks/sql/backend/thrift_backend.py
+++ b/src/databricks/sql/backend/thrift_backend.py
@@ -166,6 +166,7 @@ class ThriftDatabricksClient(DatabricksClient):
         self._use_arrow_native_complex_types = kwargs.get(
             "_use_arrow_native_complex_types", True
         )
+
         self._use_arrow_native_decimals = kwargs.get("_use_arrow_native_decimals", True)
         self._use_arrow_native_timestamps = kwargs.get(
             "_use_arrow_native_timestamps", True
--- a/src/databricks/sql/backend/thrift_backend.py
+++ b/src/databricks/sql/backend/thrift_backend.py
@@ -240,10 +240,6 @@ class ThriftDatabricksClient(DatabricksClient):
     def max_download_threads(self) -> int:
         return self._max_download_threads
 
-    @property
-    def max_download_threads(self) -> int:
-        return self._max_download_threads
-
     # TODO: Move this bounding logic into DatabricksRetryPolicy for v3 (PECO-918)
     def _initialize_retry_args(self, kwargs):
         # Configure retries & timing: use user-settings or defaults, and bound
--- a/src/databricks/sql/backend/thrift_backend.py
+++ b/src/databricks/sql/backend/thrift_backend.py
@@ -9,12 +9,11 @@ from typing import List, Optional, Union, Any, TYPE_CHECKING
 from uuid import UUID
 
 from databricks.sql.result_set import ThriftResultSet
-
+from databricks.sql.telemetry.models.event import StatementType
 
 if TYPE_CHECKING:
     from databricks.sql.client import Cursor
     from databricks.sql.result_set import ResultSet
-    from databricks.sql.telemetry.models.event import StatementType
 
 from databricks.sql.backend.types import (
     CommandState,
@@ -833,7 +832,7 @@ class ThriftDatabricksClient(DatabricksClient):
         return execute_response, is_direct_results
 
     def get_execution_result(
-        self, command_id: CommandId, cursor: "Cursor", statement_type: StatementType
+        self, command_id: CommandId, cursor: "Cursor"
     ) -> "ResultSet":
         thrift_handle = command_id.to_thrift_handle()
         if not thrift_handle:
@@ -889,6 +888,7 @@ class ThriftDatabricksClient(DatabricksClient):
             arrow_schema_bytes=schema_bytes,
             result_format=t_result_set_metadata_resp.resultFormat,
         )
+        execute_response.command_id.set_statement_type(StatementType.QUERY)
 
         return ThriftResultSet(
             connection=cursor.connection,
@@ -902,7 +902,6 @@ class ThriftDatabricksClient(DatabricksClient):
             ssl_options=self._ssl_options,
             is_direct_results=is_direct_results,
             session_id_hex=self._session_id_hex,
-            statement_type=statement_type,
         )
 
     def _wait_until_command_done(self, op_handle, initial_operation_status_resp):
@@ -968,7 +967,6 @@ class ThriftDatabricksClient(DatabricksClient):
         max_bytes: int,
         lz4_compression: bool,
         cursor: Cursor,
-        statement_type: StatementType,
         use_cloud_fetch=True,
         parameters=[],
         async_op=False,
@@ -1030,6 +1028,8 @@ class ThriftDatabricksClient(DatabricksClient):
             if resp.directResults and resp.directResults.resultSet:
                 t_row_set = resp.directResults.resultSet.results
 
+            execute_response.command_id.set_statement_type(StatementType.QUERY)
+
             return ThriftResultSet(
                 connection=cursor.connection,
                 execute_response=execute_response,
@@ -1042,7 +1042,6 @@ class ThriftDatabricksClient(DatabricksClient):
                 ssl_options=self._ssl_options,
                 is_direct_results=is_direct_results,
                 session_id_hex=self._session_id_hex,
-                statement_type=statement_type,
             )
 
     def get_catalogs(
@@ -1051,7 +1050,6 @@ class ThriftDatabricksClient(DatabricksClient):
         max_rows: int,
         max_bytes: int,
         cursor: "Cursor",
-        statement_type: StatementType,
     ) -> "ResultSet":
         thrift_handle = session_id.to_thrift_handle()
         if not thrift_handle:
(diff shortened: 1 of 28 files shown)


src/databricks/sql/backend/thrift_backend.py
```diff
@@ -1257,6 +1286,7 @@ def fetch_results(
         arrow_schema_bytes,
         description,
         use_cloud_fetch=True,
+        statement_id=None,
```
vikrantpuppala: why do we need to have a default arg?

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -66,11 +68,18 @@ def __init__(
         settings: DownloadableResultSettings,
         link: TSparkArrowResultLink,
         ssl_options: SSLOptions,
+        chunk_id: int,
+        session_id_hex: Optional[str] = None,
+        statement_id: Optional[str] = None,
```
vikrantpuppala: why default args?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -64,6 +66,8 @@ def get_execution_result(self) -> ExecutionResultFormat:
         if self.active_result_set is None:
             return ExecutionResultFormat.FORMAT_UNSPECIFIED
 
+        from databricks.sql.utils import ColumnQueue, CloudFetchQueue, ArrowQueue
```
vikrantpuppala, line 68: is this because of some circular dependency?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -73,13 +77,13 @@ def get_execution_result(self) -> ExecutionResultFormat:
...
-            return len(self.thrift_backend.retry_policy.history)
+        if hasattr(self.backend, "retry_policy") and self.backend.retry_policy:
+            return len(self.backend.retry_policy.history)
         return 0
 
+    def get_chunk_id(self):
+        return None
+
 
 class ResultSetExtractor(TelemetryExtractor):
```
vikrantpuppala: do we still need this extractor if we're not keeping latencies for fetch?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -196,6 +233,17 @@ def _safe_call(func_to_call):
                 duration_ms = int((end_time - start_time) * 1000)
 
                 extractor = get_extractor(self)
+                print(
```
vikrantpuppala: remove

src/databricks/sql/backend/thrift_backend.py
saishreeeee: I have added these args due to telemetry, don't want it to cause errors in case statement_id is not available

src/databricks/sql/cloudfetch/downloader.py
saishreeeee: don't want it to cause issues in case these args are not available as they have been added for telemetry

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: yes

src/databricks/sql/telemetry/latency_logger.py
saishreeeee: we don't, I will remove it.

src/databricks/sql/result_set.py
```diff
@@ -234,7 +238,13 @@ def __init__(
                 lz4_compressed=execute_response.lz4_compressed,
                 description=execute_response.description,
                 ssl_options=ssl_options,
+                session_id_hex=session_id_hex,
+                statement_id=execute_response.command_id.to_hex_guid(),
+                statement_type=self.statement_type,
+                chunk_id=self.chunk_id,
             )
+            if t_row_set and t_row_set.resultLinks:
+                self.chunk_id = len(t_row_set.resultLinks)
```
vikrantpuppala: what's this for?

saishreeeee: `execute_command` called: returns a `ThriftResultSet` which contains n chunks. So next chunk_id would be n.
`fetchall` called: returns the remaining m chunks. Next chunk_id += m, which happens when `fetch_results` is called.

src/databricks/sql/client.py
```diff
@@ -1060,7 +1070,7 @@ def tables(
         )
         return self
 
-    @log_latency(StatementType.METADATA)
+    @log_latency()
```
jprakash-db: I feel why is the code modified to fit telemetry? like the statement type within log latency decorator made sense. Why is it now a part of the statement type in the code ?

tests/unit/test_cloud_fetch_queue.py
```diff
@@ -63,6 +63,10 @@ def test_initializer_adds_links(self, mock_create_next_table):
             result_links=result_links,
             max_download_threads=10,
             ssl_options=SSLOptions(),
+            session_id_hex=Mock(),
```
jprakash-db, line 66: @vikrantpuppala  I think we should introduce some variables like the connection context in JDBC . 
- Passing these variables in so many functions is just bloating the code.  A future TODO 
- This will also simply the TelemetryExtractor:

src/databricks/sql/client.py
saishreeeee: The chunk downloader function doesn't know the `StatementType`, hence had to pass it down as arg.
Now have changed it to always set `statement_type` to `StatementType.QUERY` as metadata does not use cloud fetch.

jprakash-db (approved):

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -1294,9 +1302,16 @@ def fetch_results(
...
+            session_id_hex=self._session_id_hex,
+            statement_id=command_id.to_hex_guid(),
+            chunk_id=chunk_id,
         )
 
-        return queue, resp.hasMoreRows
+        return (
+            queue,
+            resp.hasMoreRows,
+            len(resp.results.resultLinks) if resp.results.resultLinks else 0,
```
varun-edachali-dbx, line 1313: Is `resp.results` guaranteed to not be `None`?

src/databricks/sql/result_set.py
```diff
@@ -234,7 +237,12 @@ def __init__(
                 lz4_compressed=execute_response.lz4_compressed,
                 description=execute_response.description,
                 ssl_options=ssl_options,
+                session_id_hex=session_id_hex,
+                statement_id=execute_response.command_id.to_hex_guid(),
+                chunk_id=self.num_downloaded_chunks,
             )
+            if t_row_set and t_row_set.resultLinks:
```
varun-edachali-dbx: `if t_row_set` is already checked in the enclosing conditional, it need not be checked again, right?

src/databricks/sql/result_set.py
```diff
@@ -215,6 +217,7 @@ def __init__(
             :param ssl_options: SSL options for cloud fetch
             :param is_direct_results: Whether there are more rows to fetch
         """
+        self.num_downloaded_chunks = 0
```
varun-edachali-dbx, line 220: Should this be instantiated in the parent `ResultSet` (so it can be re-used for `SeaResultSet`)?

src/databricks/sql/backend/thrift_backend.py
saishreeeee: Yes, `resp.results.satrtRowOffset` is also accessed above.

src/databricks/sql/result_set.py
saishreeeee: removed it

src/databricks/sql/result_set.py
saishreeeee: This is required for `ThriftResultSet` only because, `build_queue` is called during  initialization and `fetch_results`, so we need keep track of the `chunk_id`. While in `SeaResultSet`, `build_queue` is only called once, and the remaining links use `add_link`.

varun-edachali-dbx (approved):
LGTM

Issue #435, opened 2024-09-02
Allow ingesting in-memory file-like objects


dhirschfeld:
Writing large amounts of data to disk, only for `databricks-sql-connector` to then read it back in from disk, is incredibly inefficient.

It would be much more efficient to be able to provide a file-like object to use instead of a filepath. In that way a user could write the data to an in-memory `io.BytesIO` object instead of writing the data to disk.

dhirschfeld:
i.e. allow passing through `fh` rather than creating it internally by opening a file from the filesystem:
https://github.com/databricks/databricks-sql-python/blob/d31063ca918167412153a368c13a99055bf89c02/src/databricks/sql/client.py#L656-L668

kravets-levko:
Hi @dhirschfeld! This indeed sounds like an intersting feature, thank you for sharing it! I have to talk with the rest of team first. Databricks SQL `GET` and `PUT` commands should have local file path specified, but I don't know if we ever considered using streams instead of real files. If we agree that there are no risks with this approach - we would have to implement it across all drivers eventually

susodapop:
Some added context, @dhirschfeld's idea is exactly how the e2e tests for this feature behave (since we ran them in github actions where we don't have a real file system to write to). Should be a straightforward modification.

gopalldb:
This is an interesting ask. We have implemented very similar solution in other SQL driver, should be able to port that solution here.

jeffreyh1003:
+1 to OP's ask. This would be a huge win.

PR #643, opened 2025-07-21
streaming ingestion support for PUT operation

sreekanth-db:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Currently if users want to ingest an in-memory object or input stream  (which they might have generated by some processing) using the PUT command, they first need to save it into a file on the disk and need to pass the the file path in the PUT command. This involves unnecessary disk I/O for writing and reading back from the disk. In this PR we are providing an interface to directly pass the in-memory object to the command to remove disk I/O

Related issue: https://github.com/databricks/databricks-sql-python/issues/435

Example usage: https://github.com/sreekanth-db/databricks-sql-python/blob/4cd2cee3d793760ebab964191136bbd19b33d399/examples/streaming_put.py (used for manual testing as well)

Unit tests and e2e tests are also included in the changed files

Integration tests successful run: https://github.com/databricks/databricks-sql-python/actions/runs/16801948579/job/47585191807

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [x] Manually
- [ ] N/A

### DIFF
--- a/tests/unit/test_streaming_put.py
+++ b/tests/unit/test_streaming_put.py
@@ -1,9 +1,7 @@
-
 import io
 from unittest.mock import patch, Mock, MagicMock
 
 import pytest
-import requests
 
 import databricks.sql.client as client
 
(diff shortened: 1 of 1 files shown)


examples/streaming_put.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+Simple example of streaming PUT operations.
+
+This demonstrates the basic usage of streaming PUT with the __input_stream__ token.
+"""
+
+import io
+import os
+from databricks import sql
+
```
jprakash-db, line 11: nit: We don't define an executable file in example, ref the other examples just a connect and execution part is enough. The main, etc are not needed

src/databricks/sql/client.py
```diff
@@ -696,6 +715,60 @@ def _handle_staging_operation(
...
+            presigned_url: The presigned URL for upload
+            stream: Binary stream to upload
+            headers: Optional HTTP headers
+
+        Raises:
+            OperationalError: If the upload fails
+        """
+
+        # Prepare headers
+        http_headers = dict(headers) if headers else {}
```
jprakash-db: This check is redundant, we already are checking and passing a dict, otherwise you can define in the func argument as `headers: dict = {}` because expectation is to either not pass header or give a dict, never None

src/databricks/sql/client.py
```diff
@@ -696,6 +715,60 @@ def _handle_staging_operation(
...
+
+        try:
+            # Stream directly to presigned URL
+            response = requests.put(
+                url=presigned_url,
```
jprakash-db: Plz try to integrated the pre existing http client https://github.com/databricks/databricks-sql-python/blob/701f7f6d9f7fc7f8f067530db820ac4937f5ba4d/src/databricks/sql/common/http.py#L44

src/databricks/sql/client.py
```diff
@@ -696,6 +715,60 @@ def _handle_staging_operation(
...
+            CREATED = requests.codes.created  # 201
+            ACCEPTED = requests.codes.accepted  # 202
+            NO_CONTENT = requests.codes.no_content  # 204
+
+            if response.status_code not in [OK, CREATED, NO_CONTENT, ACCEPTED]:
+                raise OperationalError(
+                    f"Staging operation over HTTP was unsuccessful: {response.status_code}-{response.text}",
+                    session_id_hex=self.connection.get_session_id_hex(),
+                )
+
```
jprakash-db: This part is being repeated in both streaming and the existing flow, can we separated out this in a volume error handing util

src/databricks/sql/client.py
```diff
@@ -820,47 +894,62 @@ def execute(
...
         )
+        try:
+            # Store stream data if provided
+            self._input_stream_data = None
+            if input_stream is not None:
+                # Validate stream has required methods
+                if not hasattr(input_stream, "read"):
+                    raise TypeError(
+                        "input_stream must be a binary stream with read() method"
+                    )
```
jprakash-db: I don't feel we need to be try catching this, any Runtime error such as user providing incorrect data should fail naturally

src/databricks/sql/client.py
```diff
@@ -820,47 +894,62 @@ def execute(
...
+                self._input_stream_data = input_stream
```
jprakash-db: Do we need store this as a class variable ? Because this input_stream is just required by this function. We can just pass it down.

src/databricks/sql/client.py
```diff
@@ -820,47 +894,62 @@ def execute(
...
 
-        return self
+            if self.active_result_set and self.active_result_set.is_staging_operation:
+                self._handle_staging_operation(
+                    staging_allowed_local_path=self.connection.staging_allowed_local_path
+                )
+
+            return self
+        finally:
+            # Clean up stream data
```
jprakash-db: Don't think we need to handle the stream cleanup. Anyway stream is stored on their disk, so not cleaning it won't help with memory spikes

tests/e2e/common/streaming_put_tests.py
```diff
@@ -0,0 +1,51 @@
...
+        # Create test data
+        test_data = b"Hello, streaming world! This is test data."
+        timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+        filename = f"stream_test_{timestamp}.txt"
+        
+        with self.connection() as conn:
+            with conn.cursor() as cursor:
+                with io.BytesIO(test_data) as stream:
+                    cursor.execute(
+                        f"PUT '__input_stream__' INTO '/Volumes/{catalog}/{schema}/e2etests/{filename}'",
```
jprakash-db: What happens if a file already exists ? Does it override its contents or appends it ?
Asking because everytime this test runs we are creating new files and not cleaning up after test ends. So we have unnecessary test junk on server

tests/unit/test_streaming_put.py
```diff
@@ -0,0 +1,201 @@
+
+import io
+import unittest
+from unittest.mock import patch, Mock, MagicMock
+import databricks.sql.client as client
+from databricks.sql import ProgrammingError
+import requests
+
```
jprakash-db, line 5: Plz use pytest, we are actively deprecating unittest

tests/unit/test_streaming_put.py
```diff
@@ -0,0 +1,201 @@
...
+
+class TestStreamingPutUnit(unittest.TestCase):
+    """Unit tests for streaming PUT functionality."""
+    
+    def setUp(self):
+        """Set up test fixtures."""
```
jprakash-db: Use pytest fixtures

tests/unit/test_streaming_put.py
```diff
@@ -0,0 +1,201 @@
...
+            self.cursor.execute(
+                "PUT '__input_stream__' INTO '/Volumes/test/cat/schema/vol/file.txt'",
+                input_stream=test_stream
+            )
+            
+            # Verify staging handler was called
+            mock_handler.assert_called_once()
+            
+            # Verify the finally block cleanup
+            self.assertIsNone(self.cursor._input_stream_data)
```
jprakash-db: Same thoughts as the try catch cleanup in execute

src/databricks/sql/client.py
```diff
@@ -625,6 +626,33 @@ def _handle_staging_operation(
...
+        # Handle __input_stream__ token for PUT operations
+        if (
+            row.operation == "PUT"
+            and getattr(row, "localFile", None) == "__input_stream__"
+        ):
+            if not self._input_stream_data:
+                raise ProgrammingError(
+                    "No input stream provided for streaming operation",
+                    session_id_hex=self.connection.get_session_id_hex(),
+                )
```
jprakash-db: If needed this should be within the `handle_staging_put_stream` function as this error handling is the concern of that function

src/databricks/sql/client.py
```diff
@@ -696,6 +715,60 @@ def _handle_staging_operation(
...
+
+            if response.status_code not in [OK, CREATED, NO_CONTENT, ACCEPTED]:
+                raise OperationalError(
+                    f"Staging operation over HTTP was unsuccessful: {response.status_code}-{response.text}",
+                    session_id_hex=self.connection.get_session_id_hex(),
+                )
+
+            if response.status_code == ACCEPTED:
+                logger.debug(
+                    f"Response code {ACCEPTED} from server indicates upload was accepted "
```
jprakash-db: Can you integrate lazy logging in all the logger logs, plz refer other logging examples

src/databricks/sql/client.py
```diff
@@ -783,6 +856,7 @@ def execute(
         self,
         operation: str,
         parameters: Optional[TParameterCollection] = None,
+        input_stream: Optional[BinaryIO] = None,
```
jprakash-db: Introducing a new argument should be at the last (after `enforce_embedded_schema`), otherwise this will break the existing users functions

examples/streaming_put.py
sreekanth-db: updated the code to follow existing conventions

src/databricks/sql/client.py
sreekanth-db: removed the redundant check

src/databricks/sql/client.py
sreekanth-db: updated the code to use existing DatabricksHttpClient

src/databricks/sql/client.py
sreekanth-db: separated out repeated code to a new method

src/databricks/sql/client.py
sreekanth-db: removed the comprehensive checks for incorrect input, letting it fail naturally

src/databricks/sql/client.py
sreekanth-db: removed class variable

src/databricks/sql/client.py
sreekanth-db: removed class variable, so no cleanup required now

tests/e2e/common/streaming_put_tests.py
sreekanth-db: Ideally file with same name should't exist as we are creating it with the timestamp included. Updated the code to do cleanup after testing

tests/unit/test_streaming_put.py
sreekanth-db: updated the code to use pytest

tests/unit/test_streaming_put.py
sreekanth-db: updated the code to use pytest

tests/unit/test_streaming_put.py
sreekanth-db: since we removed class level variable, this test is not required now, removed it

src/databricks/sql/client.py
sreekanth-db: updated the error handling to happen in` handle_staging_put_stream` method

src/databricks/sql/client.py
sreekanth-db: updated all the logging statements to use lazy logging

src/databricks/sql/client.py
sreekanth-db: kept the input_stream at the end of existing arguments

src/databricks/sql/client.py
```diff
@@ -696,6 +738,43 @@ def _handle_staging_operation(
...
+        http_client = DatabricksHttpClient.get_instance()
+
+        # Stream directly to presigned URL
+        with http_client.execute(
+            method=HttpMethod.PUT,
+            url=presigned_url,
+            data=stream,
+            headers=headers,
+            timeout=300,  # 5 minute timeout
+        ) as response:
```
jprakash-db: @vikrantpuppala can you plz discuss , on the unified approach being introduced. Thanks

src/databricks/sql/client.py
```diff
@@ -696,6 +738,43 @@ def _handle_staging_operation(
...
+            self._validate_staging_http_response(response, "stream upload")
+
```
jprakash-db: can stream upload and file upload be some kind of enum ?

tests/e2e/common/streaming_put_tests.py
```diff
@@ -0,0 +1,65 @@
...
+import pytest
+from datetime import datetime
+
+logger = logging.getLogger(__name__)
+
+
+class PySQLStreamingPutTestSuiteMixin:
+    """Test suite for streaming PUT operations."""
+
+    def test_streaming_put_basic(self, catalog, schema):
```
jprakash-db, line 16: A better approach is to 
- First delete the file named suppose file_name1.txt
- Then try put operation on the file_name1.txt 
- Then perform the assertions
In this way suppose if the cleanup fails repeatedly we don't create garbage copies of filename_timestamp. There is just a file_name1.txt that needs to be cleaned in case cleanup commands fail. This current approach will fill the server with garbage files and cleanup will be hard

jprakash-db:
@vikrantpuppala Plz have a look into the http_client usage

src/databricks/sql/client.py
sreekanth-db: updated the code to use enums

tests/e2e/common/streaming_put_tests.py
sreekanth-db: updated the test

tests/unit/test_streaming_put.py
```diff
@@ -0,0 +1,154 @@
...
+            with pytest.raises(client.OperationalError) as excinfo:
+                cursor._handle_staging_put_stream(
+                    presigned_url=presigned_url,
+                    stream=test_stream
+                )
+            
+            # Check for the actual error message format
+            assert "500" in str(excinfo.value)
+
+    def test_handle_staging_put_stream_network_error(self, cursor):
```
jprakash-db: Don't think network test is needed, as its not a meaningful test

tests/unit/test_streaming_put.py
```diff
@@ -0,0 +1,154 @@
...
+    """Unit tests for streaming PUT functionality."""
+
+    @pytest.fixture
+    def mock_connection(self):
+        return Mock()
+
+    @pytest.fixture
+    def mock_backend(self):
+        return Mock()
+
```
jprakash-db: are these fixtures even needed? You can directly set connection=Mock(),

jprakash-db (approved):
Minor Nits. LGTM. Thanks for making the changes

PR #645, opened 2025-07-22
SEA: Execution Phase 

varun-edachali-dbx:
## What type of PR is this?
- [x] Feature

## Description
Introduce a preliminary SEA backend client. This is composed of the following PRs: 
- #582 
- #583 
- #587 
- #588 
- #589 
- #590 
- #593 
- #595 
- #607 
- #610 
- #620 
- #631 
- #633 

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [x] Manually
- [ ] N/A

Manually, using the test scripts in `examples/experimental`. 

[Design Document](https://docs.google.com/document/d/1Y-eXLhNqqhrMVGnOlG8sdFrCxBTN1GdQvuKG4IfHmo0/edit?usp=sharing)


src/databricks/sql/backend/sea/backend.py
```diff
@@ -0,0 +1,796 @@
...
+    GetStatementResponse,
+    CreateSessionResponse,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def _filter_session_configuration(
+    session_configuration: Optional[Dict[str, Any]],
+) -> Dict[str, str]:
```
jayantsing-db, line 54: please add some docstring here.

src/databricks/sql/backend/sea/backend.py
```diff
@@ -0,0 +1,796 @@
...
+        Args:
+            name: The name of the session configuration parameter
+
+        Returns:
+            The default value if the parameter is supported, None otherwise
+        """
+        return ALLOWED_SESSION_CONF_TO_DEFAULT_VALUES_MAP.get(name.upper())
+
+    @staticmethod
+    def get_allowed_session_configurations() -> List[str]:
```
jayantsing-db: where are these used?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -0,0 +1,796 @@
...
+        response_data = self._http_client._make_request(
+            method="POST", path=self.STATEMENT_PATH, data=request.to_dict()
+        )
+        response = ExecuteStatementResponse.from_dict(response_data)
+        statement_id = response.statement_id
+
+        command_id = CommandId.from_sea_statement_id(statement_id)
+
+        # Store the command ID in the cursor
+        cursor.active_command_id = command_id
```
jayantsing-db, line 533: what is the spec for active_command_id? what is the lifecycle? Does it remain set to previous command_id until a new one comes?

src/databricks/sql/backend/sea/utils/filters.py
```diff
@@ -0,0 +1,152 @@
...
+        # Get all remaining rows
+        all_rows = result_set.results.remaining_rows()
+
+        # Filter rows
+        filtered_rows = [row for row in all_rows if filter_func(row)]
+
+        # Reuse the command_id from the original result set
+        command_id = result_set.command_id
+
+        # Create an ExecuteResponse with the filtered data
```
jayantsing-db: nit
```suggestion
        # Create an ExecuteResponse for the filtered data
```

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -0,0 +1,186 @@
...
+        self.session = requests.Session()
+
+        # Configure SSL verification
+        if ssl_options.tls_verify:
+            self.session.verify = ssl_options.tls_trusted_ca_file or True
+        else:
+            self.session.verify = False
+
+        # Configure client certificates if provided
+        if ssl_options.tls_client_cert_file:
```
jayantsing-db, line 67: i am definitely missing something but it seems you are already respecting ssl_options?

src/databricks/sql/backend/sea/utils/http_client.py
```diff
@@ -0,0 +1,186 @@
...
+        method = method.upper()
+        if method == "GET":
+            return self.session.get
+        if method == "POST":
+            return self.session.post
+        if method == "DELETE":
+            return self.session.delete
+        raise ValueError(f"Unsupported HTTP method: {method}")
+
+    def _make_request(
```
jayantsing-db, line 101: this doesn't have retries yet right?

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -886,26 +872,21 @@ def get_execution_result(
...
-            max_download_threads=self.max_download_threads,
-            lz4_compressed=lz4_compressed,
-            description=description,
-            ssl_options=self._ssl_options,
-        )
+        lz4_compressed = t_result_set_metadata_resp.lz4Compressed
+        is_staging_operation = t_result_set_metadata_resp.isStagingOperation
+        is_direct_results = resp.hasMoreRows
+
+        status = self.get_query_state(command_id)
```
jayantsing-db: will this add an extra CP RPC in the existing thrift flow? How was status deduced earlier?

jayantsing-db: Earlier the polling was done in `get_async_execution_result`

jayantsing-db: So i think this is redundant and can be removed. Please note that the polling request has a server-side time sleep of 5 seconds so this is an expensive call (but this is not applicable in this case as command will be finished).

src/databricks/sql/result_set.py
```diff
@@ -145,50 +164,70 @@ def __init__(
...
 
-        # Initialize results queue
-        if execute_response.arrow_queue:
-            # In this case the server has taken the fast path and returned an initial batch of
-            # results
-            self.results = execute_response.arrow_queue
-        else:
-            # In this case, there are results waiting on the server so we fetch now for simplicity
+        # Initialize results queue if not provided
+        if not self.results:
```
jayantsing-db, line 226: self.results is a parent class member which initialises with the non-null queue if passed to parent?

jayantsing-db (approved):
Some minor questions/suggestions. Rest LGTM. thanks!

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Thanks for the catch - I was using them earlier but they're redundant now. Removing them.

src/databricks/sql/backend/sea/backend.py
varun-edachali-dbx: Yes, that's right. It is set to some command id when `execute` is invoked on the `Cursor`. From then on, it is essentially used to track the active operation and is passed into the backend methods for operations that require it, eg: when `cancel()` is invoked on the `Cursor` we pass the command id into the backend's `cancel_command` method. 

It is only re-set when a new one comes in (through another `execute`) or when the `Cursor` is closed - at which point it becomes `None`. 

I think it makes sense to eventually remove this attribute and store it as an attribute of the `ResultSet`, since the `active_result_set` of the `Cursor` essentially has the same life cycle and is also associated with a command.

src/databricks/sql/backend/sea/utils/http_client.py
varun-edachali-dbx: We're *mostly* respecting it, but since we're using the `requests` module we can't support key passwords as documented [on line 77.](https://github.com/databricks/databricks-sql-python/blob/1444a6711a80358d41e892a47b989cf85fc0f0e8/src/databricks/sql/backend/sea/utils/http_client.py#L78) 

This has been detailed more in the description of  #618.

src/databricks/sql/backend/sea/utils/http_client.py
varun-edachali-dbx: Yes, this does not support retries. #618 will add support for retries.

src/databricks/sql/backend/thrift_backend.py
varun-edachali-dbx: Yes, looks like earlier it was extracted from the response of the initial `FetchResults` request. I've made the change.

src/databricks/sql/result_set.py
varun-edachali-dbx: Yes, exactly. This mostly respects the existing method of initialisation as well - we initialise the `queue` with the arrow queue if it is provided in the initial execute response, otherwise we use `fill_results_buffer` to fetch it from the server.

PR #646, opened 2025-07-22
Telemetry server-side flag integration

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -245,12 +245,6 @@ class Connection:
         self.use_cloud_fetch = kwargs.get("use_cloud_fetch", True)
         self._cursors = []  # type: List[Cursor]
 
-        self.server_telemetry_enabled = True
-        self.client_telemetry_enabled = kwargs.get("enable_telemetry", False)
-        self.telemetry_enabled = (
-            self.client_telemetry_enabled and self.server_telemetry_enabled
-        )
-
         self.session = Session(
             server_hostname,
             http_path,
@@ -267,6 +261,17 @@ class Connection:
             kwargs.get("use_inline_params", False)
         )
         self.staging_allowed_local_path = kwargs.get("staging_allowed_local_path", None)
+
+        self.client_telemetry_enabled = kwargs.get("enable_telemetry", False)
+        if self.client_telemetry_enabled:
+            self.server_telemetry_enabled = TelemetryHelper.is_server_telemetry_enabled(
+                self
+            )
+            self.telemetry_enabled = (
+                self.client_telemetry_enabled and self.server_telemetry_enabled
+            )
+        else:
+            self.telemetry_enabled = False
 
         TelemetryClientFactory.initialize_telemetry_client(
             telemetry_enabled=self.telemetry_enabled,
--- a/src/databricks/sql/common/feature_flag.py
+++ b/src/databricks/sql/common/feature_flag.py
@@ -1,0 +1,177 @@
+# databricks/sql/common/feature_flags.py
+
+import threading
+import time
+import requests
+from dataclasses import dataclass, field
+from concurrent.futures import ThreadPoolExecutor
+from typing import Dict, Optional, List, Any, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from databricks.sql.client import Connection
+
+
+@dataclass
+class FeatureFlagEntry:
+    """Represents a single feature flag from the server response."""
+
+    name: str
+    value: str
+
+
+@dataclass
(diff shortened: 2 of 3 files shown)


src/databricks/sql/common/feature_flag.py
```diff
@@ -0,0 +1,181 @@
...
+
+@dataclass
+class FeatureFlagEntry:
+    """Represents a single feature flag from the server response."""
+
+    name: str
+    value: str
+
+
+@dataclass
```
samikshya-db, line 20: Can you add a TODO to test this one server flag is available?

src/databricks/sql/client.py
```diff
@@ -268,6 +262,17 @@ def read(self) -> Optional[OAuthToken]:
         )
         self.staging_allowed_local_path = kwargs.get("staging_allowed_local_path", None)
 
+        self.client_telemetry_enabled = kwargs.get("enable_telemetry", False)
+        if self.client_telemetry_enabled:
+            self.server_telemetry_enabled = TelemetryHelper.is_server_telemetry_enabled(
```
samikshya-db: Can you also add a flag to force enable telemetry? In which case, there will be 2 cases in which telemetry is enabled : 

1. If Force enable flag is set
2. If enable_telemetry is true && server flag is true

src/databricks/sql/common/feature_flag.py
```diff
@@ -0,0 +1,181 @@
...
+
+import threading
+import time
+import requests
+from dataclasses import dataclass, field
+from concurrent.futures import ThreadPoolExecutor
+from typing import Dict, Optional, List, Any, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from databricks.sql.client import Connection
```
vikrantpuppala, line 9: what's this?

src/databricks/sql/common/feature_flag.py
```diff
@@ -0,0 +1,181 @@
...
+            # If cache has never been loaded, perform a synchronous, blocking fetch.
+            if self._flags is None:
+                self._refresh_flags()
+
+            # If a proactive background refresh is needed, start one. This is non-blocking.
+            elif self._is_refresh_needed():
+                # We don't check for an in-flight refresh; the executor queues the task, which is safe.
+                self._executor.submit(self._refresh_flags)
+
+            assert self._flags is not None
```
vikrantpuppala, line 95: does this mean we will just fail if we aren't able to fetch flags? shouldn't we do safe defaults?

src/databricks/sql/common/feature_flag.py
saishreeeee: It is for static type checking and prevents circular imports

src/databricks/sql/common/feature_flag.py
saishreeeee: If there is an exception in `_refresh_flags()`, `_flags` is set to an empty Dict. If the particular flag we are checking is set to `None`, we use the default value passed in the args.

samikshya-db (approved):

src/databricks/sql/common/feature_flag.py
```diff
@@ -0,0 +1,179 @@
...
+
+            # Now, return the value from the populated cache.
+            flag_value = self._flags.get(name)
+            if flag_value is None:
+                return default_value
+            return flag_value.lower() == "true"
```
vikrantpuppala: don't think we should have the assumption that the flag_value would be a boolean. It'd be great if we can make this extensible to other types of safe flags.

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -98,6 +103,19 @@ def get_auth_flow(auth_provider):
...
+    @staticmethod
+    def is_telemetry_enabled(connection: "Connection") -> bool:
+        if connection.force_enable_telemetry:
+            return True
+
+        if connection.enable_telemetry:
+            context = FeatureFlagsContextFactory.get_instance(connection)
+            return context.is_feature_enabled(
+                TelemetryHelper.TELEMETRY_FEATURE_FLAG_NAME, default_value=False
+            )
```
vikrantpuppala, line 115: it'd be great if we make this more extensive by instead defining a function like getFlagValue(flagName, defaultValue) and then do whatever we need to based on the datatype of that flag

PR #653, opened 2025-07-28
SEA: Allow large metadata responses 

varun-edachali-dbx:
## What type of PR is this?

- [x] Feature

## Description

Currently, for metadata queries using the SEA backend, `use_cloud_fetch` is hardcoded to `False`. However, this leads to it being less robust to large metadata queries. Thus, here, we re-use `use_cloud_fetch` from the connection params instead of hardcoding it to `False` to aid robustness. 

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [x] Manually - using large metadata queries in #642 
- [ ] N/A

[Design Document](https://docs.google.com/document/d/1Y-eXLhNqqhrMVGnOlG8sdFrCxBTN1GdQvuKG4IfHmo0/edit?usp=sharing)


jayantsing-db (approved):
do we have integration tests for these?

varun-edachali-dbx:
> do we have integration tests for these?

There are currently no e2e tests running for the metadata queries because there is some work to be done with regards to normalisation. I have added a note to add it into #639.

PR #657, opened 2025-07-30
Added code coverage workflow to test the code coverage from unit and e2e tests

msrathore-db:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Added code coverage workflow to test the code coverage from unit and e2e tests. 
## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

Added some untested functions and unit tests to see if the check is sensitive to new additions.
Used act to locally test the workflow.

[PECOBLR-621](https://databricks.atlassian.net/browse/PECOBLR-621)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,8 @@ mypy = "^1.10.1"
 pylint = ">=2.12.0"
 black = "^22.3.0"
 pytest-dotenv = "^0.5.2"
+pytest-cov = "^4.0.0"
+coverage = "^7.0.0"
 numpy = [
     { version = ">=1.16.6", python = ">=3.8,<3.11" },
     { version = ">=1.23.4", python = ">=3.11" },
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -37,6 +37,8 @@ mypy = "^1.10.1"
 pylint = ">=2.12.0"
 black = "^22.3.0"
 pytest-dotenv = "^0.5.2"
+pytest-cov = "^4.0.0"
+coverage = "^7.0.0"
```
jprakash-db: Why do you need pytest-cov and coverage both ? pytest-cov has coverage already as a dependency right?

msrathore-db: Makes sense
Resolved.

jprakash-db (approved):
Overall LGTM. Since the coverage is above 85%. You can add the tests to fail if coverage drops below 85 %

PR #658, opened 2025-07-30
Telemetry circuit breaker

saishreeeee:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A


jprakash-db:
Just confused on why is there even a need for circuit breaker, isn't just handling a good retry mechanism with proper retry behaviour on `retry-after` time or status code based retry enough?  cc @vikrantpuppala

README.md
```diff
@@ -11,7 +11,7 @@ You are welcome to file an issue here for general use cases. You can also contac
 
 ## Requirements
 
-Python 3.8 or above is required.
+Python 3.9 or above is required.
```
vikrantpuppala, line 14: @jprakash-db are we ok with python version upgrades?

samikshya-db: This should be ok, given DBT tests pass.

src/databricks/sql/common/http.py
```diff
@@ -141,7 +149,14 @@ def post(self, url: str, **kwargs) -> requests.Response:
         This is a blocking call intended to be run in a background thread.
         """
         logger.debug("Executing telemetry POST request to: %s", url)
-        return self.session.post(url, **kwargs)
+        try:
+            return self.circuit_breaker.call(self.session.post, url, **kwargs)
+        except CircuitBreakerError as e:
+            logger.error("Circuit breaker error: %s", e)
```
samikshya-db: let's keep all logs pretty much silent for telemetry. use trace instead.

PR #663, opened 2025-08-05
query tags integration

sreekanth-db:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
This PR adds **Query Tags support** to the Python SQL Connector. Query Tags allow users to attach key-value pairs to SQL executions that appear in the `system.query.history` table.

### Changes Made:

1. **Created Example File** (`examples/query_tags_example.py`):
   - New file demonstrating Query Tags usage with session configuration

2. **Enhanced Integration Test** (`tests/e2e/test_driver.py`):
   - Modified `test_ssp_passthrough()` to include Query Tags in session configuration
   - Added parametrization to test both Thrift and SEA backends

3. **Enhanced Unit Tests** (`tests/unit/test_session.py`):
   - Updated `test_configuration_passthrough()` test to verify query tags are properly passed through to the thrift backend
   - Updated `test_session_management()` to include query tags in session configuration to ensure query tags work correctly in SEA backend

### Implementation Details:

**Added Query Tags Support in SEA flow** (`src/databricks/sql/backend/sea/utils/constants.py`):
   - Added `"QUERY_TAGS": ""` to `ALLOWED_SESSION_CONF_TO_DEFAULT_VALUES_MAP` to enable query tags support in the SEA (Statement Execution API) backend flow

### Usage example:

```python
import databricks.sql

connection = databricks.sql.connect(
    server_hostname='host',
    http_path='/sql/1.0/warehouses/xxx',
    access_token='token',
    session_configuration={
        'QUERY_TAGS': 'team:engineering,test:query-tags',
        'ansi_mode': False
    }
)
```

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [x] Manually
- [ ] N/A

### Testing Details:

**Unit Tests**: 
- `test_configuration_passthrough()` - Verifies Query Tags flow in thrift mode
- `test_session_management()` - Verifies Query Tags flow in SEA mode
- All existing tests still pass

**E2E Tests**: 
- Enhanced `test_ssp_passthrough()` to include Query Tags in both thirft and SEA modes

**Manual Testing**:
- Ran `examples/query_tags_example.py` with real backend connection
- Verified Query Tags are accepted and queries execute successfully

### DIFF
--- a/tests/e2e/test_driver.py
+++ b/tests/e2e/test_driver.py
@@ -848,21 +848,10 @@ class TestPySQLCoreSuite(
                 query = "select * from range(1000000000)"
                 cursor.execute(query)
 
-    @pytest.mark.parametrize(
-        "extra_params",
-        [
-            {
-                "use_sea": False,
-            },
-            {
-                "use_sea": True,
-            },
-        ],
-    )
-    def test_ssp_passthrough(self, extra_params):
+    def test_ssp_passthrough(self):
         for enable_ansi in (True, False):
             with self.cursor(
-                {"session_configuration": {"ansi_mode": enable_ansi, "QUERY_TAGS": "team:marketing,dashboard:abc123,driver:python"}, **extra_params}
+                {"session_configuration": {"ansi_mode": enable_ansi}}
             ) as cursor:
                 cursor.execute("SET ansi_mode")
                 assert list(cursor.fetchone()) == ["ansi_mode", str(enable_ansi)]
@@ -908,7 +897,7 @@ class TestPySQLCoreSuite(
     )
     def test_multi_timestamps_arrow(self, extra_params):
         with self.cursor(
-            {"session_configuration": {"ansi_mode": False}, **extra_params}
+            {"session_configuration": {"ansi_mode": False, "query_tags": "test:multi-timestamps,driver:python"}, **extra_params}
         ) as cursor:
             query, expected = self.multi_query()
             expected = [


tests/e2e/test_driver.py
```diff
@@ -851,7 +851,7 @@ def test_socket_timeout_user_defined(self):
     def test_ssp_passthrough(self):
         for enable_ansi in (True, False):
             with self.cursor(
-                {"session_configuration": {"ansi_mode": enable_ansi}}
+                {"session_configuration": {"ansi_mode": enable_ansi, "QUERY_TAGS": "team:marketing,dashboard:abc123,driver:python"}}
```
shivam2680: I think this can be omitted

sreekanth-db: Testing pass through of query tags config (Server-Side Properties) to backend as discussed offline

tests/e2e/test_driver.py
```diff
@@ -848,10 +848,21 @@ def test_socket_timeout_user_defined(self):
...
+            {
+                "use_sea": True,
+            },
+        ],
+    )
+    def test_ssp_passthrough(self, extra_params):
         for enable_ansi in (True, False):
             with self.cursor(
-                {"session_configuration": {"ansi_mode": enable_ansi}}
+                {"session_configuration": {"ansi_mode": enable_ansi, "QUERY_TAGS": "team:marketing,dashboard:abc123,driver:python"}, **extra_params}
```
nikhilsuri-db: Should we add a test with invalid query param format & see if we are getting the required error?

tests/unit/test_sea_backend.py
```diff
@@ -683,12 +687,14 @@ def test_filter_session_configuration(self):
...
             "STATEMENT_TIMEOUT": 7200,  # uppercase key
             "TiMeZoNe": "America/New_York",  # mixed case key
+            "QueRy_TaGs": "team:marketing,test:case-insensitive",
         }
         result = _filter_session_configuration(case_insensitive_config)
         expected_case_result = {
             "ansi_mode": "false",
             "statement_timeout": "7200",
             "timezone": "America/New_York",
+            "query_tags": "team:marketing,test:case-insensitive",
```
nikhilsuri-db, line 697: Similarly two scenarios we can add here
1. Not passing the query tags config param
2. Passing non parsable  query tags -> **"team:marketing,test"**

tests/e2e/test_driver.py
sreekanth-db: In the backend design docs, it is mentioned that invalid query tags will be ignored without raising error so that actual query execution won't result in failure. Tested manually in local by passing invalid/non-parsable query tags - backend doesn't throw error

tests/unit/test_sea_backend.py
sreekanth-db: 1. `_filter_session_configuration ` method loops through provided session config fields and filters out invalid params, so separate unit test for not passing query tags would be an overkill. For e2e tests, we already have few tests in which we don't pass query tags as part of session config. 

2. We don't do any parsing of query tags in the driver and we send the user provided value as is to the backend. As mentioned in the other comment, backend would ignore the non parsable query tags without raising the error

tests/e2e/test_driver.py
```diff
@@ -848,10 +848,21 @@ def test_socket_timeout_user_defined(self):
                 query = "select * from range(1000000000)"
                 cursor.execute(query)
 
-    def test_ssp_passthrough(self):
+    @pytest.mark.parametrize(
+        "extra_params",
```
samikshya-db: can we club this test into some other e2e test? (as this may not require a separate test tbh)

sreekanth-db: We are actually not creating a new test, just updating the existing one to include query tags in both SEA and Thrift modes. If the existing test itself is redundant, i think we can take that up in a separate refactoring PR

samikshya-db: I was asking as the following were not present 
```
[
            {
                "use_sea": False,
            },
            {
                "use_sea": True,
            },
        ],
    )
```

Let's club where this is already present (as this adds to the time taking to run all tests)

sreekanth-db: got it, updated the tests

PR #668, opened 2025-08-08
Refactor codebase to use a unified http client

vikrantpuppala:
## What type of PR is this?

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Refactor codebase to use a unified http client
## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [x] Manually
- [ ] N/A


src/databricks/sql/auth/oauth.py
```diff
@@ -209,10 +213,13 @@ def __send_token_request(token_request_url, data):
             "Accept": "application/json",
             "Content-Type": "application/x-www-form-urlencoded",
         }
-        response = requests.post(
-            url=token_request_url, data=data, headers=headers, auth=IgnoreNetrcAuth()
+        # Use unified HTTP client
+        from databricks.sql.common.unified_http_client import IgnoreNetrcAuth
+        response = self.http_client.request(
```
jprakash-db, line 200: How is a static method using class http_client ?

src/databricks/sql/client.py
```diff
@@ -292,6 +291,7 @@ def read(self) -> Optional[OAuthToken]:
             auth_provider=self.session.auth_provider,
             host_url=self.session.host,
             batch_size=self.telemetry_batch_size,
+            http_client=self.session.http_client,
```
jprakash-db, line 302: so we are not keeping different client for Telemetry ( I believe there is different retry behaviour for it )

src/databricks/sql/client.py
```diff
@@ -744,16 +744,20 @@ def _handle_staging_put(
             )
 
         with open(local_file, "rb") as fh:
-            r = requests.put(url=presigned_url, data=fh, headers=headers)
+            r = self.connection.session.http_client.request('PUT', presigned_url, body=fh.read(), headers=headers)
```
jprakash-db: @sreekanth-db   Can you discuss with @vikrantpuppala and integrate the new common client approach in volume operations

src/databricks/sql/common/unified_http_client.py
```diff
@@ -0,0 +1,226 @@
...
+            return self.request('PUT', url, body=file_obj.read(), headers=headers)
+
+    def download_file(self, url: str, file_path: str, headers: Optional[Dict[str, str]] = None) -> None:
+        """
+        Download a file using GET method.
+        
+        Args:
+            url: URL to download from
+            file_path: Path where to save the downloaded file
+            headers: Optional headers
```
jprakash-db: don't think they should be part of the Http client

src/databricks/sql/utils.py
```diff
@@ -104,15 +105,16 @@ def build_queue(
...
-                start_row_offset=t_row_set.startRowOffset,
-                result_links=t_row_set.resultLinks,
-                lz4_compressed=lz4_compressed,
-                description=description,
                 max_download_threads=max_download_threads,
                 ssl_options=ssl_options,
                 session_id_hex=session_id_hex,
                 statement_id=statement_id,
                 chunk_id=chunk_id,
+                http_client=http_client,
```
jprakash-db, line 113: why is the order of params changed ?

PR #673, opened 2025-08-08
Refactor codebase to use a unified http client

vikrantpuppala:
## What type of PR is this?

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## Description
This PR refactors the codebase to use a unified HTTP client across all HTTP operations in the Databricks SQL connector. The primary purpose is to consolidate HTTP client usage, eliminating various singleton HTTP clients and direct requests library usage in favor of a single UnifiedHttpClient based on urllib3.

Key changes:

- Introduced UnifiedHttpClient class that provides a centralized HTTP client with retry policies, connection pooling, SSL support, and proxy support
- Removed singleton HTTP clients (DatabricksHttpClient, TelemetryHttpClient) and replaced direct requests usage
- Updated all components (auth, telemetry, staging operations, cloud fetch, feature flags) to use the unified client

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [x] Manually tested auth flows
- [ ] N/A


src/databricks/sql/auth/common.py
```diff
@@ -52,6 +66,24 @@ def __init__(
...
+        # HTTP client configuration
+        self.ssl_options = ssl_options
+        self.socket_timeout = socket_timeout
+        self.retry_stop_after_attempts_count = retry_stop_after_attempts_count or 5
+        self.retry_delay_min = retry_delay_min or 1.0
+        self.retry_delay_max = retry_delay_max or 60.0
+        self.retry_stop_after_attempts_duration = (
+            retry_stop_after_attempts_duration or 900.0
+        )
+        self.retry_delay_default = retry_delay_default or 5.0
```
vikrantpuppala: @jprakash-db please check it these defaults are ok. please keep in mind that these are now being applied to - auth, telemetry, cloud fetch and uc volume ops

src/databricks/sql/auth/oauth.py
```diff
@@ -203,16 +190,17 @@ def __send_auth_code_token_request(
         data = f"{token_request_body}&code_verifier={verifier}"
         return self.__send_token_request(token_request_url, data)
 
-    @staticmethod
-    def __send_token_request(token_request_url, data):
+    def __send_token_request(self, token_request_url, data):
```
vikrantpuppala, line 193: the two methods calling this were not static so changed this as well

src/databricks/sql/auth/retry.py
```diff
@@ -355,8 +355,8 @@ def should_retry(self, method: str, status_code: int) -> Tuple[bool, str]:
         logger.info(f"Received status code {status_code} for {method} request")
 
         # Request succeeded. Don't retry.
-        if status_code == 200:
-            return False, "200 codes are not retried"
+        if status_code // 100 == 2:
+            return False, "2xx codes are not retried"
```
vikrantpuppala: DELETE volume operations return 204 which should be classified as success

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -14,19 +13,6 @@
...
-    total=5,  # max retry attempts
-    backoff_factor=1,  # min delay, 1 second
-    # TODO: `backoff_max` is supported since `urllib3` v2.0.0, but we allow >= 1.26.
-    #       The default value (120 seconds) used since v1.26 looks reasonable enough
-    # backoff_max=60,  # max delay, 60 seconds
-    # retry all status codes below 100, 429 (Too Many Requests), and all codes above 500,
-    # excluding 501 Not implemented
-    status_forcelist=[*range(0, 101), 429, 500, *range(502, 1000)],
-)
-
```
vikrantpuppala, line 29: this was not being used

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -150,6 +153,44 @@ def _flush(self):
         pass
 
 
+class TelemetryHttpClientSingleton:
```
vikrantpuppala: Since telemetry runs in background in a separate thread pool, using the client created in Connection class directly was causing issues as the http client would close with a Connection but telemetry is shared

tests/e2e/common/staging_ingestion_tests.py
```diff
@@ -68,15 +68,20 @@ def test_staging_ingestion_life_cycle(self, ingestion_user):
...
         remove_query = f"REMOVE 'stage://tmp/{ingestion_user}/tmp/11/16/file1.csv'"
-
-        with self.connection(extra_params={"staging_allowed_local_path": "/"}) as conn:
+        # Use minimal retry settings to fail fast for staging operations
+        extra_params = {
+            "staging_allowed_local_path": "/",
+            "_retry_stop_after_attempts_count": 1,
+            "_retry_delay_max": 10,
+        }
+        with self.connection(extra_params=extra_params) as conn:
```
vikrantpuppala, line 77: earlier UC volume operations were not retried at all, now they are.

src/databricks/sql/auth/auth.py
```diff
@@ -53,6 +55,8 @@ def get_auth_provider(cfg: ClientContext):
                 cfg.oauth_redirect_port_range,
                 cfg.oauth_client_id,
                 cfg.oauth_scopes,
+                http_client,
+                cfg.auth_type or "databricks-oauth",
```
jprakash-db: Plz don't use magic strings, use enums

src/databricks/sql/auth/common.py
jprakash-db: Why are you using or and then defining the values, can't you directly define the default value in the initializer itself. `relay_delay_min : float = 1.0`

src/databricks/sql/auth/common.py
```diff
@@ -52,6 +66,24 @@ def __init__(
...
+        self.retry_dangerous_codes = retry_dangerous_codes or []
+        self.http_proxy = http_proxy
+        self.proxy_username = proxy_username
+        self.proxy_password = proxy_password
+        self.pool_connections = pool_connections or 10
```
jprakash-db, line 85: same comment, or values can be set as default during function call

src/databricks/sql/auth/common.py
```diff
@@ -78,23 +110,22 @@ def get_azure_tenant_id_from_host(host: str, http_client=None) -> str:
...
 
-    if http_client is None:
-        http_client = DatabricksHttpClient.get_instance()
-
     login_url = f"{host}/aad/auth"
     logger.debug("Loading tenant ID from %s", login_url)
-    with http_client.execute(HttpMethod.GET, login_url, allow_redirects=False) as resp:
-        if resp.status_code // 100 != 3:
+
+    with http_client.request_context("GET", login_url, allow_redirects=False) as resp:
```
jprakash-db: Why is the `HttpMethod.GET` being replaced by "GET" ?

src/databricks/sql/auth/common.py
```diff
@@ -78,23 +110,22 @@ def get_azure_tenant_id_from_host(host: str, http_client=None) -> str:
...
+        if resp.status // 100 != 3:
             raise ValueError(
-                f"Failed to get tenant ID from {login_url}: expected status code 3xx, got {resp.status_code}"
+                f"Failed to get tenant ID from {login_url}: expected status code 3xx, got {resp.status}"
```
jprakash-db: use lazy logging

src/databricks/sql/auth/common.py
```diff
@@ -78,23 +110,22 @@ def get_azure_tenant_id_from_host(host: str, http_client=None) -> str:
...
-        if len(path_segments) < 2:
-            raise ValueError(f"Invalid path in Location header: {url.path}")
-        return path_segments[1]
+
+    # The Location header has the following form: https://login.microsoftonline.com/<tenant-id>/oauth2/authorize?...
+    # The domain may change depending on the Azure cloud (e.g. login.microsoftonline.us for US Government cloud).
+    url = urlparse(entra_id_endpoint)
+    path_segments = url.path.split("/")
+    if len(path_segments) < 2:
+        raise ValueError(f"Invalid path in Location header: {url.path}")
```
jprakash-db, line 130: lazy logging plz

src/databricks/sql/client.py
```diff
@@ -252,10 +254,14 @@ def read(self) -> Optional[OAuthToken]:
             "telemetry_batch_size", TelemetryClientFactory.DEFAULT_BATCH_SIZE
         )
 
+        client_context = self._build_client_context(server_hostname, **kwargs)
```
jprakash-db: nit: feel build_client_context can be a util function and not a class one that takes any **kwargs and populates the ClientContext object

src/databricks/sql/client.py
```diff
@@ -395,7 +447,7 @@ def get_protocol_version(openSessionResp: TOpenSessionResp):
     @property
     def open(self) -> bool:
         """Return whether the connection is open by checking if the session is open."""
-        return self.session.is_open
+        return hasattr(self, "session") and self.session.is_open
```
jprakash-db: why won't it have the session object ? Like why is this being checked

src/databricks/sql/client.py
```diff
@@ -744,25 +796,27 @@ def _handle_staging_put(
             )
 
         with open(local_file, "rb") as fh:
-            r = requests.put(url=presigned_url, data=fh, headers=headers)
+            r = self.connection.session.http_client.request(
+                "PUT", presigned_url, body=fh.read(), headers=headers
```
jprakash-db: "PUT", etc can be enums

src/databricks/sql/client.py
```diff
@@ -783,30 +837,38 @@ def _handle_staging_get(
                 session_id_hex=self.connection.get_session_id_hex(),
             )
 
-        r = requests.get(url=presigned_url, headers=headers)
+        r = self.connection.session.http_client.request(
```
jprakash-db: What is the logic for having a http_client per `session` and not per `connection` and is there any difference ? Like I feel we could just have it as `self.connection.http_client`

src/databricks/sql/cloudfetch/downloader.py
```diff
@@ -104,50 +91,47 @@ def run(self) -> DownloadedFile:
...
+                "ResultSetDownloadHandler: downloaded file size {} does not match the expected value {}".format(
+                    len(decompressed_data), self.link.bytesNum
                 )
             )
 
-            return DownloadedFile(
-                decompressed_data,
-                self.link.startRowOffset,
-                self.link.rowCount,
+        logger.debug(
```
jprakash-db, line 126: use lazy logging

src/databricks/sql/common/unified_http_client.py
```diff
@@ -0,0 +1,206 @@
...
+            url: URL to request
+            headers: Optional headers dict
+            **kwargs: Additional arguments passed to urllib3 request
+
+        Returns:
+            urllib3.HTTPResponse: The HTTP response object with data pre-loaded
+        """
+        with self.request_context(method, url, headers=headers, **kwargs) as response:
+            # Read the response data to ensure it's available after context exit
+            response._body = response.data
```
jprakash-db, line 205: What about the other parts of response? Like status , etc are they not deleted ?

src/databricks/sql/session.py
```diff
@@ -185,3 +193,7 @@ def close(self) -> None:
             logger.error("Attempt to close session raised a local exception: %s", e)
 
         self.is_open = False
+
+        # Close HTTP client if it exists
+        if hasattr(self, "http_client") and self.http_client:
```
jprakash-db: why is this `hasattr` check even needed ?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -150,6 +153,44 @@ def _flush(self):
...
+                    cls._instance = super().__new__(cls)
+                    cls._instance._http_client = None
+                    cls._instance._initialized = False
+        return cls._instance
+
+    def get_http_client(self, client_context):
+        """Get or create the singleton HTTP client."""
+        if not self._initialized and client_context:
+            with self._lock:
+                if not self._initialized:
```
jprakash-db: Is this needed ? Can we create a uuid within client context and use that as a key and create a UnifiedHttpClient when that particular uuid is not present. I feel creating a singleton instance of the class just to anyway call get_http_client with client context seems uneccesarily verbose

src/databricks/sql/utils.py
```diff
@@ -104,15 +105,16 @@ def build_queue(
...
-                result_links=t_row_set.resultLinks,
-                lz4_compressed=lz4_compressed,
-                description=description,
                 max_download_threads=max_download_threads,
                 ssl_options=ssl_options,
                 session_id_hex=session_id_hex,
                 statement_id=statement_id,
                 chunk_id=chunk_id,
+                http_client=http_client,
+                start_row_offset=t_row_set.startRowOffset,
```
jprakash-db: Why was the order of the other variables changed?

tests/unit/test_cloud_fetch_queue.py
```diff
@@ -13,6 +13,31 @@
 
 @pytest.mark.skipif(pyarrow is None, reason="PyArrow is not installed")
 class CloudFetchQueueSuite(unittest.TestCase):
+    def create_queue(self, schema_bytes=None, result_links=None, description=None, **kwargs):
```
jprakash-db, line 16: nit: doesn't seem related to this PR

src/databricks/sql/auth/common.py
vikrantpuppala: error thrown here

src/databricks/sql/auth/common.py
vikrantpuppala: error thrown here

src/databricks/sql/common/unified_http_client.py
vikrantpuppala: 1. status and headers are basic attributes that are set when the HTTPResponse object is created and remain accessible even after close() is called.
2. data is a special property that may need to read from the underlying connection stream. This is why it needs to be cached before the connection is closed.

tests/unit/test_cloud_fetch_queue.py
vikrantpuppala: just created a helper function to cut down the verbosity of creating a queue in each test (adding http client everywhere was increasing the verbosity)

src/databricks/sql/telemetry/telemetry_client.py
vikrantpuppala: I was facing some issues while testing: https://github.com/databricks/databricks-sql-python/pull/673#discussion_r2266234232, if this is a concern i can try to look for other ways

src/databricks/sql/client.py
vikrantpuppala: http_clients are actually per connection. we just weren't storing it there, made the change now

src/databricks/sql/auth/common.py
vikrantpuppala: this is because, we have client context created at two places: client.py and auth.py, this could be deduped but let's keep it for a later PR to simplify? don't want to introduce more changes into this PR

src/databricks/sql/common/unified_http_client.py
```diff
@@ -0,0 +1,216 @@
...
+        Args:
+            method: HTTP method (HttpMethod.GET, HttpMethod.POST, HttpMethod.PUT, HttpMethod.DELETE)
+            url: URL to request
+            headers: Optional headers dict
+            **kwargs: Additional arguments passed to urllib3 request
+
+        Yields:
+            urllib3.HTTPResponse: The HTTP response object
+        """
+        logger.debug("Making %s request to %s", method, url)
```
vikrantpuppala: @jprakash-db does there have to be some safe logging for these logs?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -180,7 +185,9 @@ def __init__(
         self._driver_connection_params = None
         self._host_url = host_url
         self._executor = executor
-        self._http_client = TelemetryHttpClient.get_instance()
+
+        # Use the provided HTTP client directly
```
jprakash-db: Nit: don't need this comment, if its in function call it is assumed to be used

src/databricks/sql/common/unified_http_client.py
jprakash-db: Plz dont' log the presigned url, or just log the host

src/databricks/sql/auth/oauth.py
```diff
@@ -348,17 +337,22 @@ def refresh(self) -> Token:
...
                     oauth_response.access_token,
                     oauth_response.token_type,
                     oauth_response.refresh_token,
                 )
             else:
                 raise Exception(
-                    f"Failed to get token: {response.status_code} {response.text}"
+                    f"Failed to get token: {response.status} {response.data.decode('utf-8')}"
                 )
+        finally:
```
jprakash-db: Is this closing necessary? I believe it is already closed inside the execute

Issue #471, opened 2024-11-20
TCP connection not closing properly leading to dangling CLOSE_WAIT state and system overload


anugrahasinha:
Testcase 1
```
Python 3.10.15 (main, Sep  7 2024, 18:35:33) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from databricks import sql
>>> sql.__version__
'3.1.2'
>>> with sql.connect(server_hostname = "<HIDE>", http_path = "<HIDE>", access_token = "<HIDE>") as connection:
...     with connection.cursor() as cursor:
...             cursor.execute("select * from table")
...             column_names = [desc[0] for desc in cursor.description]
...             query_out = [dict(zip(column_names, row)) for row in cursor.fetchall()]
...             
... 
<databricks.sql.client.Cursor object at 0x7f88551c14e0>
>>> query_out
[{"sometime" : "something}]


On the sidelines, the network connection for this process (after with clause gets finished), remain in ESTABLISHED state and then moves to CLOSE_WAIT state indefinitely. 

Every 1.0s: netstat -apn | grep 1070099                                                                                          asinha-vm-ubuntu-20-04-1: Wed Nov 20 02:43:13 2024

tcp       25      0 10.5.220.4:58078        so:me:ip:add:443      CLOSE_WAIT  1070099/python3
```

Testcase 2
```
Python 3.10.15 (main, Sep  7 2024, 18:35:33) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from databricks import sql
>>> sql.__version__
'3.1.2'
>>> connection = sql.connect(server_hostname = "<HIDE>", http_path = "<HIDE>", access_token = "<HIDE>")
>>> ## at this point a TCP connection got made as ESTABLISHED state ##
>>> cursor = connection.cursor()
>>> cursor.execute("select * from table")
<databricks.sql.client.Cursor object at 0x7f9d65db26b0>
>>> column_names = [desc[0] for desc in cursor.description]
>>> query_out = [dict(zip(column_names, row)) for row in cursor.fetchall()]
>>> cursor.close()
>>> connection.close()
>>> # even after close being called, the connection remains in ESTABLISHED state (for a long time) and then finally moves to CLOSE_WAIT state, which never changes.

Every 1.0s: netstat -apn | grep 1074808                                                                                          asinha-vm-ubuntu-20-04-1: Wed Nov 20 02:49:36 2024

[...]

PR #674, opened 2025-08-09
[471] - Close underlying HTTP Client on closing `Connection`

Varun0157:
(Addresses #471). 

## What type of PR is this?
- [x] Bug Fix

## Description
Here, we resolve a few resource leaks with regards to TCP connections remaining open even after the `Connection` is closed: 
- Currently, on session closure, the thrift backend invokes `self._transport.close()` which simply releases a connection back to the underlying pool. It does not close the pool itself. This leads to the TCP connection not being closed properly and a dangling `CLOSE_WAIT` state as pointed out in the above issue. 
Thus, we rename `_transport.close()` to the more appropriate `_transport.release_connection()` and implement a new `_transport.close()` that actually closes the underlying pool. 
We invoke this method on SEA backend session closure as well. 
- Currently, the `close()` method of the `TelemetryClient` does not close the unified HTTP client instantiated by it. We add code to close this as well.

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [x] Manually
    - using the same scripts mentioned in the description of #471 - the connection is now closed properly and there is no dangling `CLOSE_WAIT`.
    - testing with an invalid token - before the addition of the `http_client.close()` in the `TelemetryClient`, a similar resource leak as pointed out above was observed, that has now been mitigated.  
- [ ] N/A

#471


src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -359,6 +359,7 @@ def close(self):
         """Flush remaining events before closing"""
         logger.debug("Closing TelemetryClient for connection %s", self._session_id_hex)
         self._flush()
+        self._http_client.close()
```
vikrantpuppala, line 438: Great catch!
the issue here would be that _flush is async, it is going to start a bunch of futures to perform the POST /telemetry-ext but then the main thread is going to close the http_client which is then going to cause all those futures to fail. We need to have to do something smarter here to ensure that all futures are completes in the background before closing this http client (also in the background)

vikrantpuppala (commented):
Thanks for making these fixes! could you update from the latest main (just made significant changes to use a unified http client)

src/databricks/sql/telemetry/telemetry_client.py
Varun0157: I have merged the updated `main` into my branch. 

Thanks for the catch, I didn't realise `_flush` was async. 
Regarding the above: waiting for the futures to complete would essentially make the `TelementryClient`'s `close()` synchronous, right? Is this an acceptable tradeoff? It seems necessary if we want to prevent the TCP connection from remaining open. 

I'll go ahead and implement this change by adding some state that tracks each future and waits for them to complete before invoking the `close()` of the HTTP Client, but we can revert it or iterate on it if it seems like the telemetry changes begin to hamper the main process.

vikrantpuppala: we wouldn't want to make the close() synchronous, I was trying this out, but the delay of this being sync is just too much. ideally we would like to perform async close of http_client based on future completion

src/databricks/sql/backend/sea/backend.py
```diff
@@ -302,6 +300,23 @@ def close_session(self, session_id: SessionId) -> None:
...
+
+        Raises:
+            ValueError: If the session ID is invalid
+            OperationalError: If there's an error closing the session
+        """
+
+        logger.debug("SeaDatabricksClient.close_session(session_id=%s)", session_id)
+
+        self._close_session(session_id)
+        self._http_client.close()
```
jayantsing-db: why is a session and http client closed together? I feel it is easier to imagine the backend module as a function module for interacting with SQL EXEC. So, with that perspective, there should be minimal state in this and the backend should receive resources by higher level classes and those higher level classes should take care of closing the resources (for example, the connection/cursor should pass a http client to this and close the http client. Same goes for session: cursor maintains the session obtained functionally through this backend and should take care of closing the session). Please let me know if this doesn't look okay.

jayantsing-db (changes requested):

src/databricks/sql/backend/sea/backend.py
Varun0157: I agree with the philosophy overall, but I think passing the HTTP client into the backend from a higher-level object is likely infeasible right now since both backends have distinct, custom HTTP clients that expose different methods and are processed differently. Until we unify them, it makes sense to make the backend responsible for it's own HTTP Client instantiation, as a result of which it should be responsible for closing it as well. 

Both HTTP Clients do have largely the same functionality, so unifying them (for the most part) could be a separate PR, at which point we can pass in the HTTP client from the `Session` class. 
Let me know if I should do it in this PR instead. 

Note that the pattern of instantiating the required HTTP client in the constructor of the concrete backend is common across Thrift and SEA.

Varun0157:
Thoughts? 
@jayantsing-db / @vikrantpuppala / @jprakash-db

PR #678, opened 2025-08-19
Fix compatibility with urllib3<2 and add CI actions to improve dependency checks

vikrantpuppala:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Other

## Description
Add CI actions to improve dependency checks
## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [x] Manually
- [ ] N/A

### DIFF
--- a/.github/workflows/code-quality-checks.yml
+++ b/.github/workflows/code-quality-checks.yml
@@ -47,7 +47,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .venv
-          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ github.event.repository.name }}-${{ hashFiles('**/poetry.lock') }}
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ matrix.dependency-version }}-${{ github.event.repository.name }}-${{ hashFiles('**/poetry.lock') }}
       #----------------------------------------------
       # install dependencies if cache does not exist
       #----------------------------------------------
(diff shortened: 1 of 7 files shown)


.github/workflows/code-quality-checks.yml
```diff
@@ -50,15 +60,47 @@ jobs:
       - name: Install library
         run: poetry install --no-interaction
       #----------------------------------------------
+      # override with custom dependency versions
```
jprakash-db, line 63: The cache key you are using, doesn't that also need to be updated with the `dependency-version` in the file name? Other wise both the dependency version end up using the same cache which is wrong

scripts/dependency_manager.py
```diff
@@ -0,0 +1,234 @@
...
+                else:
+                    # For other packages, use minor version constraint
+                    # e.g., 1.2.5 becomes >=1.2.5,<1.3.0
+                    next_minor = int(minor) + 1
+                    upper_bound = f"{major}.{next_minor}.0"
+                    return f"{package_name}>={min_version},<{upper_bound}"
+            else:
+                # If version doesn't have minor version, just use exact version
+                return f"{package_name}=={min_version}"
+
```
jprakash-db, line 105: Are these cases handled , there are also things based on python versions - 
```
pandas = [
    { version = ">=1.2.5,<2.3.0", python = ">=3.8,<3.13" },
    { version = ">=2.2.3,<2.3.0", python = ">=3.13" }
]
```

vikrantpuppala: yes, although since the cicd only runs for <=3.13 right now, it is easier to do so

.github/workflows/code-quality-checks.yml
vikrantpuppala: makes sense, done

jprakash-db (approved):

samikshya-db (approved):

PR #680, opened 2025-08-19
Add documentation for proxy support

vikrantpuppala:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Add documentation for proxy support
## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -1,0 +1,232 @@
+# Proxy Support
+
+The Databricks SQL Connector supports connecting through HTTP and HTTPS proxy servers with various authentication methods. This feature automatically detects system proxy configuration and handles proxy authentication transparently.
+
+## Quick Start
+
+The connector automatically uses your system's proxy configuration when available:
+
+```python
+from databricks import sql
+
+# Basic connection - uses system proxy automatically
+with sql.connect(
+    server_hostname="your-workspace.cloud.databricks.com",
+    http_path="/sql/1.0/endpoints/your-endpoint-id",
+    access_token="your-token"
+) as connection:
+    # Your queries here...
+```
+
+For advanced proxy authentication (like Kerberos), specify the authentication method:
+
+```python
+with sql.connect(
+    server_hostname="your-workspace.cloud.databricks.com",
+    http_path="/sql/1.0/endpoints/your-endpoint-id", 
+    access_token="your-token",
+    _proxy_auth_method="negotiate"  # Enable Kerberos proxy auth
+) as connection:
+    # Your queries here...
+```
+
+## Proxy Configuration
+
+### Environment Variables
+
+The connector follows standard proxy environment variable conventions:
+
+| Variable | Description | Example |
+|----------|-------------|---------|
+| `HTTP_PROXY` | Proxy for HTTP requests | `http://proxy.company.com:8080` |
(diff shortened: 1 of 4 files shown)


docs/proxy.md
```diff
@@ -0,0 +1,232 @@
...
+- **HTTPS Proxies**: Encrypted proxy connections
+- **Authentication**: Basic, Negotiate/Kerberos
+- **Bypass Rules**: Full support for NO_PROXY patterns
+
+### Connection Architecture
+
+The connector uses a unified HTTP client that maintains:
+- **Direct Pool Manager**: For non-proxy connections
+- **Proxy Pool Manager**: For proxy connections  
+- **Per-Request Routing**: Automatic selection based on target host
```
jprakash-db, line 230: Is this part relevant for the proxy.md ? Don't think architecture details are needed

examples/proxy_authentication.py
```diff
@@ -0,0 +1,153 @@
...
+    elif success_count > 0:
+        print("⚠ Some proxy authentication methods failed")
+        print("This may be normal depending on your proxy configuration")
+        return 0
+    else:
+        print("✗ All proxy authentication methods failed")
+        if not has_proxy:
+            print("Consider checking your proxy configuration")
+        return 1
+
```
jprakash-db, line 151: Wouldn't it be better to print after each proxy method whether that was successful or not. This end summary gives a vague picture

docs/proxy.md
```diff
@@ -0,0 +1,232 @@
...
+2. **Proxy bypass rules** - Honor `NO_PROXY` and system bypass settings  
+3. **Target host** - Check if the specific host should use proxy
+
+### Connection Pooling
+
+The connector maintains separate connection pools for direct and proxy connections, allowing efficient handling of mixed proxy/direct traffic.
+
+### SSL/TLS with Proxy
+
+HTTPS connections through HTTP proxies use the CONNECT method for SSL tunneling. The connector handles this automatically while preserving all SSL verification settings.
```
jprakash-db, line 143: Can the user add there custom SSL certificates using the usual `_ssl_cert` arguments in the connection or this follows a different flow?

docs/proxy.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Per-Request Proxy Decisions
+
+The connector automatically makes per-request decisions about proxy usage based on:
+
+1. **System proxy configuration** - Detected from environment variables
+2. **Proxy bypass rules** - Honor `NO_PROXY` and system bypass settings  
+3. **Target host** - Check if the specific host should use proxy
+
+### Connection Pooling
```
jprakash-db, line 137: Is this relevant for a readme? Looks like architecture details

docs/proxy.md
vikrantpuppala: ssl arguments can be used as is, no changes there

docs/proxy.md
vikrantpuppala: I guess ok to have additional details incase anyone is interested in to read all of this

examples/proxy_authentication.py
vikrantpuppala: So folks would need to comment out any auth methods that they do not use because the proxy would only support one of the auth methods

PR #688, opened 2025-08-26
[PECOBLR-681] added new session conf to enable metric view metadata

shivam2680:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Add new session conf to set spark conf to enable metric view metadata

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

Tested using custom dbr image. Verified:
1. cursor.columns() => returns <data_type> measure for measure columns
2. cursor.tables() => returns METRIC_VIEW for metric views on UC

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -200,6 +200,12 @@ class Connection:
                     STRUCT is returned as Dict[str, Any]
                     ARRAY is returned as numpy.ndarray
                 When False, complex types are returned as a strings. These are generally deserializable as JSON.
+            :param enable_metric_view_metadata: `bool`, optional (default is False)
+                When True, enables metric view metadata support by setting the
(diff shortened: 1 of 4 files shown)


src/databricks/sql/client.py
```diff
@@ -248,6 +254,15 @@ def read(self) -> Optional[OAuthToken]:
             access_token_kv = {"access_token": access_token}
             kwargs = {**kwargs, **access_token_kv}
 
+        # Handle enable_metric_view_metadata parameter
```
jprakash-db: nit: don't add obvious comments

jprakash-db (approved):

src/databricks/sql/client.py
shivam2680: removed.

PR #691, opened 2025-09-05
Add Token Federation Support for Databricks SQL Python Driver

madhav-db:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
This PR implements token federation functionality for the databricks-sql-python driver, enabling seamless integration with external Identity Providers (IdPs) like Azure AD, Okta, and others. 
 Token federation allows users to authenticate with external IdPs and automatically exchange those tokens for Databricks in-house tokens when needed. This enables:
  - Workflow-Wide (M2M) Token Federation: Service principal authentication with external IdPs
  - Account-Wide (U2M) Token Federation: User authentication through external IdPs

Flow:
  1. Token Analysis: Checks if the external token's issuer differs from the Databricks host
  2. Automatic Exchange: If different, exchanges the token via /oidc/v1/token endpoint using OAuth 2.0 token exchange flow
  3. Caching: Caches exchanged tokens until expiry
  4. Fallback: Uses the original external token if exchange fails
## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

Extensive testing was performed covering:

  1. M2M (Machine-to-Machine) Token Federation
    - External service principal tokens from Azure AD
    - Automatic token exchange with Databricks workspace
    - Authentication as service principal in Databricks
  2. U2M (User-to-Machine) Token Federation
    - Browser-based OAuth flow with automatic token handling
    - Pre-obtained user tokens from external IdPs
    - Authentication as actual users in Databricks
  3. Token Lifecycle Management
    - Token caching with proper expiry handling
    - Automatic refresh when tokens expire
    - Graceful fallback when exchange fails
  4. Cross-Cloud Compatibility
    - Tested with GCP Databricks workspace using Azure AD tokens
    - Tested with Azure Databricks workspace
    - Verified issuer-based exchange decision logic

### DIFF
--- a/tests/unit/test_token_federation.py
+++ b/tests/unit/test_token_federation.py
@@ -4,9 +4,8 @@ import json
 import jwt
 from datetime import datetime, timedelta
 
-from databricks.sql.auth.token_federation import TokenFederationProvider
+from databricks.sql.auth.token_federation import TokenFederationProvider, Token
 from databricks.sql.auth.auth_utils import (
-    Token,
     parse_hostname,
     decode_token,
     is_same_host,


src/databricks/sql/auth/auth.py
```diff
@@ -60,6 +64,17 @@ def get_auth_provider(cfg: ClientContext, http_client):
             )
         else:
             raise RuntimeError("No valid authentication settings!")
+    
+    # Wrap with token federation if enabled
+    if cfg.enable_token_federation and base_provider:
```
jprakash-db: don't think the `and base_provider` check is needed. there will always be a base provider

src/databricks/sql/auth/auth.py
```diff
@@ -114,5 +129,8 @@ def get_python_sql_connector_auth_provider(hostname: str, http_client, **kwargs)
         else redirect_port_range,
         oauth_persistence=kwargs.get("experimental_oauth_persistence"),
         credentials_provider=kwargs.get("credentials_provider"),
+        # Token federation parameters
```
jprakash-db: nit: comment not needed. It is self explanatory

src/databricks/sql/auth/auth.py
```diff
@@ -114,5 +129,8 @@ def get_python_sql_connector_auth_provider(hostname: str, http_client, **kwargs)
...
+        enable_token_federation=kwargs.get("enable_token_federation", False),
```
jprakash-db: Our expectation is for token federation to be default, and only in case of error we fallback to the provided auth. Don't think we need to give an enable token_federation option

src/databricks/sql/auth/common.py
```diff
@@ -37,6 +37,9 @@ def __init__(
         tls_client_cert_file: Optional[str] = None,
         oauth_persistence=None,
         credentials_provider=None,
+        # Token federation parameters
+        enable_token_federation: bool = False,
```
jprakash-db: Same comment. Token federation should always be on, fallback to original auth if error is there.

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+        
+        Args:
+            hostname: The Databricks workspace hostname
+            external_provider: The external authentication provider
+            http_client: HTTP client for making requests
+            identity_federation_client_id: Optional client ID for token federation
+        """
+        self.hostname = self._normalize_hostname(hostname)
+        self.external_provider = external_provider
+        self.http_client = http_client or requests.Session()
```
jprakash-db: Don't ever fallback to `requests.Session()`. We should always have a http client passed. It will break in proxy setups

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+            raise ValueError("Authorization header is missing")
+            
+        parts = auth_header.split(" ", 1)
+        if len(parts) != 2:
+            raise ValueError("Invalid Authorization header format")
+            
+        return parts[0], parts[1]
+        
+    def _is_same_host(self, url1: str, url2: str) -> bool:
+        """Check if two URLs have the same host."""
```
jprakash-db: can we move this to a util file?

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+        try:
+            host1 = urlparse(url1).netloc
+            host2 = urlparse(url2).netloc
+            return host1 == host2
+        except Exception as e:
+            logger.debug(f"Failed to parse URLs: {e}")
+            return False
+            
+    def _normalize_hostname(self, hostname: str) -> str:
```
jprakash-db: This can also be moved to a util file

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+        
+        # Calculate expiry time
+        if "expires_in" in token_info:
+            expires_in = int(token_info["expires_in"])
+            # Set expiry with a 1-minute buffer
+            self._cached_token_expiry = datetime.now() + timedelta(seconds=expires_in - 60)
+        else:
+            # Try to get expiry from JWT
+            try:
+                decoded = jwt.decode(
```
jprakash-db: Why to use 2 methods to get expiry time, just use the expiry from the JWT that is the most reliable and always available one

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+                )
+                exp = decoded.get("exp")
+                if exp:
+                    self._cached_token_expiry = datetime.fromtimestamp(exp) - timedelta(minutes=1)
+                else:
+                    # Default to 1 hour if no expiry info
+                    self._cached_token_expiry = datetime.now() + timedelta(hours=1)
+            except:
+                # Default to 1 hour if we can't decode
+                self._cached_token_expiry = datetime.now() + timedelta(hours=1)
```
jprakash-db: Why are you handling these cases?  It is never possible for a JWT to not have expiry. Can't decode JWT is anyway a root error and there is no need to process

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        self.hostname = self._normalize_hostname(hostname)
+        self.external_provider = external_provider
+        self.http_client = http_client or requests.Session()
+        self.identity_federation_client_id = identity_federation_client_id
+        
+        self._cached_token = None
+        self._cached_token_expiry = None
+        self._external_headers = {}
+
```
jprakash-db: Plz create a new object called Token and move the token specific things, like check if expired etc inside it. This will have clear separation of concern and easy maintability

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+        """Check if the token should be exchanged based on issuer."""
+        try:
+            # Decode JWT without verification to check issuer
+            decoded = jwt.decode(access_token, options={"verify_signature": False})
+            issuer = decoded.get("iss", "")
+            
+            # Check if issuer host is different from Databricks host
+            return not self._is_same_host(issuer, self.hostname)
+        except Exception as e:
+            logger.debug(f"Failed to decode JWT token: {e}")
```
jprakash-db: use lazy logging

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+        token_type, access_token = self._extract_token_from_header(auth_header)
+        
+        # Check if token exchange is needed
+        if self._should_exchange_token(access_token):
+            try:
+                exchanged_token = self._exchange_token(access_token)
+                self._cache_token(exchanged_token)
+                return exchanged_token
+            except Exception as e:
+                logger.warning(f"Token exchange failed, using external token: {e}")
```
jprakash-db: plz use lazy logging in all the places

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+        # Prepare the token exchange request
+        data = {
+            "grant_type": self.TOKEN_EXCHANGE_GRANT_TYPE,
+            "subject_token": access_token,
+            "subject_token_type": self.TOKEN_EXCHANGE_SUBJECT_TYPE,
+            "scope": "sql",
+            "return_original_token_if_authenticated": "true",
+        }
+        
+        # Add client_id if provided
```
jprakash-db: plz don't add obvious comments

src/databricks/sql/auth/token_federation.py
```diff
@@ -0,0 +1,226 @@
...
+            
+        headers = {
+            "Content-Type": "application/x-www-form-urlencoded",
+            "Accept": "*/*",
+        }
+        
+        # Encode data as URL-encoded form
+        body = urlencode(data)
+        
+        # Make the token exchange request using UnifiedHttpClient API
```
jprakash-db: same, plz remove comments that add no meaningful info

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+import json
+import jwt
+from datetime import datetime, timedelta
+import requests
+
+from databricks.sql.auth.token_federation import TokenFederationProvider, ExternalTokenProvider
+from databricks.sql.auth.authenticators import AccessTokenAuthProvider
+
+
+class TestTokenFederationProvider(unittest.TestCase):
```
jprakash-db: Plz use pytest, we are actively moving away from unittest

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+    
+    def setUp(self):
```
jprakash-db: use Fixtures

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+            ("https://test.databricks.com/", "https://test.databricks.com/"),
+            ("test.databricks.com/", "https://test.databricks.com/"),
+        ]
+        
+        for input_hostname, expected in test_cases:
+            provider = TokenFederationProvider(
+                hostname=input_hostname,
+                external_provider=self.external_provider,
+            )
+            self.assertEqual(provider.hostname, expected)
```
jprakash-db: is this even a token federation test? This function should be a util and the test should in test_utils or somewhere else

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+            
+        # Invalid header - empty
+        with self.assertRaises(ValueError):
+            self.provider._extract_token_from_header("")
+            
+    def test_is_same_host(self):
+        """Test host comparison."""
+        test_cases = [
+            ("https://test.databricks.com", "https://test.databricks.com", True),
+            ("https://test.databricks.com", "https://test.databricks.com:443", False),
```
jprakash-db: isn't test.databricks.com and test.databricks.com:443 the same host?

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+            result = self.provider._should_exchange_token("mock-jwt-token")
+            self.assertTrue(result)
+            
+    def test_should_exchange_token_same_issuer(self):
+        """Test that token exchange is not triggered for same issuer."""
+        # Create a mock JWT with same issuer
+        token_payload = {
+            "iss": "https://test.databricks.com",
+            "aud": "databricks",
+            "exp": int((datetime.now() + timedelta(hours=1)).timestamp()),
```
jprakash-db: too much boiler plate. These are just the same tests with different parameters. Use parameterized tests

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+            headers2 = {}
+            self.provider.add_headers(headers2)
+            self.assertEqual(headers2["Authorization"], "Bearer external-token-2")
+            
+            # External provider should be called twice
+            self.assertEqual(self.external_provider.add_headers.call_count, 2)
+
+
+class TestExternalTokenProvider(unittest.TestCase):
+
```
jprakash-db: This TestExternalTokenProvider belongs in the `test_auth.py`

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+        """Test successful token exchange."""
+        access_token = "external-token-123"
+        
+        # Mock successful response (UnifiedHttpClient style)
+        mock_response = Mock()
+        mock_response.data = json.dumps({
+            "access_token": "databricks-token-456",
+            "token_type": "Bearer",
+            "expires_in": 3600,
+        }).encode('utf-8')
```
jprakash-db: move this mock response to a function, it is being copied multiple times

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+            "iss": "https://login.microsoftonline.com/tenant-id/",
+            "aud": "databricks",
+            "exp": int((datetime.now() + timedelta(hours=1)).timestamp()),
+        }
+        
+        # Mock successful token exchange
+        mock_response = Mock()
+        mock_response.data = json.dumps({
+            "access_token": "databricks-token-456",
+            "token_type": "Bearer",
```
jprakash-db: Plz use parameterized tests, the test_add_headers_with_token_exchange and test_add_headers_without_token_exchange are just the same test with different arguments. Too much boiler plate

tests/unit/test_token_federation.py
```diff
@@ -0,0 +1,325 @@
...
+            provider.add_headers(request_headers)
+            
+        # Credentials provider should only be called once
+        credentials_provider.assert_called_once()
+        
+        # Header factory should be called 3 times
+        self.assertEqual(header_factory.call_count, 3)
+
+
+if __name__ == "__main__":
```
jprakash-db: same comment, try to avoid unittest and use pytest

src/databricks/sql/auth/auth_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        # Handle port differences (e.g., example.com vs example.com:443)
+        host1_without_port = host1.split(":")[0]
+        host2_without_port = host2.split(":")[0]
+        return host1_without_port == host2_without_port
+    except Exception as e:
+        logger.debug("Failed to parse URLs: %s", e)
+        return False
+
+
+class Token:
```
jprakash-db: Token is a class not a util

PR #697, opened 2025-10-03
Added a workflow to parallelise the E2E tests. Updated E2E tests to c…

msrathore-db:
…reate new table names for each run to avoid issue in parallelisation

## What type of PR is this?

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## Description
Added a workflow to parallelise the E2E tests. Updated E2E tests to create new table names for each run to avoid issue in parallelisation
## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

### DIFF
--- a/tests/e2e/test_parameterized_queries.py
+++ b/tests/e2e/test_parameterized_queries.py
@@ -4,6 +4,7 @@ from decimal import Decimal
 from enum import Enum
 from typing import Dict, List, Type, Union
 from unittest.mock import patch
+from uuid import uuid4
 
 import time
 import numpy as np
@@ -130,9 +131,13 @@ class TestParameterizedQueries(PySQLPytestTestCase):
         Note that this fixture doesn't clean itself up. So the table will remain
         in the schema for use by subsequent test runs.
         """
+        
+        # Generate unique table name to avoid conflicts in parallel execution
+        table_name = f"pysql_e2e_inline_param_test_table_{str(uuid4()).replace('-', '_')}"
+        self.inline_table_name = table_name
 
-        query = """
-            CREATE TABLE IF NOT EXISTS pysql_e2e_inline_param_test_table (
+        query = f"""
+            CREATE TABLE IF NOT EXISTS {table_name} (
             null_col INT,
             int_col INT,
             bigint_col BIGINT,
@@ -179,9 +184,10 @@ class TestParameterizedQueries(PySQLPytestTestCase):
         :paramstyle:
             This is a no-op but is included to make the test-code easier to read.
         """
-        INSERT_QUERY = f"INSERT INTO pysql_e2e_inline_param_test_table (`{target_column}`) VALUES (%(p)s)"
-        SELECT_QUERY = f"SELECT {target_column} `col` FROM pysql_e2e_inline_param_test_table LIMIT 1"
-        DELETE_QUERY = "DELETE FROM pysql_e2e_inline_param_test_table"
+        table_name = self.inline_table_name
+        INSERT_QUERY = f"INSERT INTO {table_name} (`{target_column}`) VALUES (%(p)s)"
+        SELECT_QUERY = f"SELECT {target_column} `col` FROM {table_name} LIMIT 1"
+        DELETE_QUERY = f"DELETE FROM {table_name}"
 
         with self.connection(extra_params={"use_inline_params": True}) as conn:
             with conn.cursor() as cursor:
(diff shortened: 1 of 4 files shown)


.github/workflows/coverage-check-parallel.yml
```diff
@@ -0,0 +1,417 @@
...
+          path: .venv
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ github.event.repository.name }}-${{ hashFiles('**/poetry.lock') }}
+
+      - name: Install dependencies
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction --no-root
+
+      - name: Install library
+        run: poetry install --no-interaction --all-extras
+
```
jprakash-db: why is the poetry setup part being done multiple times, each part basically has the same setup

.github/workflows/coverage-check-parallel.yml
```diff
@@ -0,0 +1,417 @@
...
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+          installer-parallel: true
+
+      - name: Load cached venv
+        id: cached-poetry-dependencies
+        uses: actions/cache@v4
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ github.event.repository.name }}-${{ hashFiles('**/poetry.lock') }}
```
jprakash-db: Again we are creating multiple environments , which is just too much overhead.
Simpler would be create a first job called setup that installs the poetry and activates the venv
Then find all the test files and execute them in matrix fashion within the setup done first
Don't think it is efficient to create the entire venv setup for each test

.github/workflows/coverage-check-parallel.yml
```diff
@@ -0,0 +1,417 @@
...
+          if-no-files-found: warn  
+
+  common-tests:
+    runs-on: ubuntu-latest
+    environment: azure-prod
+    needs: discover-tests
+    if: ${{ needs.discover-tests.outputs.common-test-files != '[]' }}
+    strategy:
+      matrix:
+        test_file: ${{ fromJson(needs.discover-tests.outputs.common-test-files) }}
```
jprakash-db: why are you running each single test is parallel. This is just too much overhead. When using a matrix you are basically spinning up a VM for each test and this is just wrong

tests/e2e/test_parameterized_queries.py
```diff
@@ -179,9 +187,15 @@ def _inline_roundtrip(self, params: dict, paramstyle: ParamStyle, target_column)
...
-        DELETE_QUERY = "DELETE FROM pysql_e2e_inline_param_test_table"
+        if not hasattr(self, 'inline_table_name'):
+            table_name = f"pysql_e2e_inline_param_test_table_{str(uuid4()).replace('-', '_')}"
+            self.inline_table_name = table_name
+            self._create_inline_table(table_name)
+        
+        table_name = self.inline_table_name
+        INSERT_QUERY = f"INSERT INTO {table_name} (`{target_column}`) VALUES (%(p)s)"
+        SELECT_QUERY = f"SELECT {target_column} `col` FROM {table_name} LIMIT 1"
+        DELETE_QUERY = f"DELETE FROM {table_name}"
```
jprakash-db, line 198: There is a incomplete resource cleanup here. If the Insert or Select query fails then the Delete query will never be executed and so for every new run we will end up creating new garbage tables

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,136 @@
+name: Code Coverage
+
```
jprakash-db, line 2: If this is satisfying the usecase, why have you not deleted this file -.github/workflows/coverage-check.yml

msrathore-db: Will remove both coverage-check and integrations workflow since the new workflow satisfies the needs. Just kept the workflow to ensure the coverage percentage is same for the new workflow as well.

.github/workflows/coverage-check-parallel.yml
msrathore-db: Updated.

.github/workflows/coverage-check-parallel.yml
msrathore-db: Updated.

.github/workflows/coverage-check-parallel.yml
msrathore-db: Updated.

jprakash-db (approved):

tests/e2e/test_parameterized_queries.py
msrathore-db: updated

PR #698, opened 2025-10-09
Update native.py: use dict instead of Dict

wyattscarpenter:
The capital letter version from typing has been deprecated since Python 3.9 https://docs.python.org/3/library/typing.html#typing.Dict

## What type of PR is this?

- [x] Refactor

## How is this tested?

Type check in the CI

### DIFF
--- a/src/databricks/sql/parameters/native.py
+++ b/src/databricks/sql/parameters/native.py
@@ -13,7 +13,7 @@ from databricks.sql.thrift_api.TCLIService.ttypes import (
 import datetime
 import decimal
 from enum import Enum, auto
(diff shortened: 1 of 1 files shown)


src/databricks/sql/parameters/native.py
```diff
@@ -13,7 +13,7 @@
 import datetime
 import decimal
 from enum import Enum, auto
-from typing import Dict, List, Union
+from typing import, List, Union
```
wyattscarpenter:
```suggestion
from typing import List, Union
```

amarvin:
This package is supposed to still support Python 3.8 though.

PR #704, opened 2025-11-03
feat: Add multi-statement transaction support

jayantsing-db:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description

Implement PEP 249-compliant transaction control with extensions for manual commit/rollback operations. This enables atomic multi-table operations with REPEATABLE_READ isolation semantics.

Core API additions:
- connection.autocommit property for enabling/disabling auto-commit mode
- connection.commit() to commit active transactions
- connection.rollback() to rollback active transactions
- connection.get_transaction_isolation() returns current isolation level
- connection.set_transaction_isolation() validates isolation level
- TransactionError exception for transaction-specific failures

Implementation details:
- Added autocommit state caching in Session with optional server query
- Added TRANSACTION_ISOLATION_LEVEL_REPEATABLE_READ constant
- All transaction operations include proper error handling and telemetry
- Supports fetch_autocommit_from_server connection parameter

Testing:
- Unit tests covering all transaction methods and error scenarios
- e2e integration tests validating transaction behavior including multi-table atomicity, sequential transactions, and isolation semantics

Documentation:
- Comprehensive TRANSACTIONS.md guide with examples and best practices
- Updated README.md with basic usage and reference to detailed docs

Requires MST-enabled Databricks SQL warehouse and Delta tables with 'delta.feature.catalogOwned-preview' table property.

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [x] Manually
- [ ] N/A


README.md
```diff
@@ -67,6 +67,37 @@ or to a Databricks Runtime interactive cluster (e.g. /sql/protocolv1/o/123456789
...
+    cursor.execute("INSERT INTO table1 VALUES (1, 'a')")
+    cursor.execute("INSERT INTO table2 VALUES (2, 'b')")
+    connection.commit()  # Commit both inserts atomically
+except Exception as e:
+    connection.rollback()  # Rollback on error
+    raise
+
+cursor.close()
+connection.close()
+```
```
vikrantpuppala: let's remove example code from the README and keep it in the examples folder

TRANSACTIONS.md
vikrantpuppala, line 1: This documents the behaviour of MSTs and are not driver specific, I feel like it would be better if we just point to a single documentation page that covers all of these scenarios and link that in all drivers

TRANSACTIONS.md
```diff
@@ -0,0 +1,368 @@
...
+    connection.rollback()  # Discard the partial transaction
+    
+    # Now you can start a fresh transaction
+    cursor.execute("INSERT INTO error_log VALUES (1, 'Query failed')")
+    connection.commit()
+```
+
+## Querying Server State
+
+By default, the `autocommit` property returns a cached value for performance. If you need to query the server each time (for example, if you're debugging or the state might change externally):
```
vikrantpuppala: "the state might change externally" might not be good to mention this

tests/e2e/test_transactions.py
```diff
@@ -0,0 +1,597 @@
...
+
+    # ==================== BASIC AUTOCOMMIT TESTS ====================
+
+    def test_default_autocommit_is_true(self):
+        """Test that new connection defaults to autocommit=true."""
+        assert (
+            self.connection.autocommit is True
+        ), "New connection should have autocommit=true by default"
+
+    def test_set_autocommit_to_false(self):
```
vikrantpuppala, line 142: can remove this test since below test also does this

TRANSACTIONS.md
jayantsing-db: done.

TRANSACTIONS.md
jayantsing-db: I agree. But I cannot find any good public facing doc regarding interactive transactions with auto-commit in drivers. Maybe because the feature is in PrPr. Can I add this temporarily as a usage guide?

README.md
jayantsing-db: done

tests/e2e/test_transactions.py
jayantsing-db: This is very slightly different from the below in the sense that the below test checks for a happy path, the state change from auto-commit true to auto-commit false.

[In case you are wondering why the number of tests in PySQL is low: these are intentionally a subset of the tests in the JDBC PR. I believe the testing workspace for transactions isn’t enabled yet. I will copy over the remaining tests once this is integrated into automated CI/CD.]

README.md
vikrantpuppala: can we add it to the examples? https://github.com/databricks/databricks-sql-python/tree/main/examples

jayantsing-db: done

kacpermuda:
Hey @jayantsing-db, out of curiosity, is AUTOCOMMIT configuration with SQL generally available on Databricks already? We've encountered an error in Airflow after this PR (details [here](https://github.com/apache/airflow/issues/58386)) saying `Configuration AUTOCOMMIT is not available.` when trying to set autocommit to either False or True.

PR #705, opened 2025-11-04
Circuit breaker changes using pybreaker

nikhilsuri-db:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
This PR introduces a circuit breaker pattern to the telemetry system to prevent cascading failures and improve system resilience. The implementation uses the pybreaker library to monitor telemetry request failures and automatically open the circuit when failure rates exceed configurable thresholds, blocking further requests to protect downstream services. When the circuit is open, telemetry requests are temporarily blocked until the system recovers, at which point the circuit transitions to half-open for testing and eventually closes when normal operation resumes. The circuit breaker configuration is centralized with immutable settings, includes comprehensive logging for monitoring, and maintains full backward compatibility with existing telemetry functionality.

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [ ] Manually
- [ ] N/A

📦 Package Versions Verified
databricks-sql-connector: 4.2.1 (your local version)
pybreaker:                1.4.1 ✨ (successfully installed)
dbt-databricks:           1.11.0 ✅
databricks-sqlalchemy:    2.0.8 ✅
databricks-sdk:           0.67.0 ✅
sqlalchemy:               2.0.44 ✅
**LLM GENERATED**

https://docs.google.com/document/d/1ftRvby9bwDZzE3s1tOb4hJ4Pd9USiXskb9cDw-uQNPM/edit?usp=sharing


src/databricks/sql/telemetry/circuit_breaker_manager.py
```diff
@@ -0,0 +1,138 @@
...
+)
+LOG_CIRCUIT_BREAKER_CLOSED = (
+    "Circuit breaker closed for %s - telemetry requests will be allowed"
+)
+LOG_CIRCUIT_BREAKER_HALF_OPEN = (
+    "Circuit breaker half-open for %s - testing telemetry requests"
+)
+
+
+class CircuitBreakerStateListener(CircuitBreakerListener):
```
nikhilsuri-db, line 43: Only used for logging purposed for now

src/databricks/sql/telemetry/circuit_breaker_manager.py
```diff
@@ -0,0 +1,138 @@
...
+from typing import Dict, Optional, Any
+from dataclasses import dataclass
+
+import pybreaker
+from pybreaker import CircuitBreaker, CircuitBreakerError, CircuitBreakerListener
+
+logger = logging.getLogger(__name__)
+
+# Circuit Breaker Configuration Constants
+MINIMUM_CALLS = 20
```
nikhilsuri-db: Hardcoded values as do not want user to configure them anywhere in the driver

pyproject.toml
```diff
@@ -26,6 +26,7 @@ pyarrow = [
     { version = ">=18.0.0", python = ">=3.13", optional=true }
 ]
 pyjwt = "^2.0.0"
+pybreaker = "^1.0.0"
```
nikhilsuri-db, line 29: @vikrantpuppala Can we be sure that adding new library won't break any other client usage?

src/databricks/sql/auth/common.py
```diff
@@ -51,6 +51,7 @@ def __init__(
         pool_connections: Optional[int] = None,
         pool_maxsize: Optional[int] = None,
         user_agent: Optional[str] = None,
+        telemetry_circuit_breaker_enabled: Optional[bool] = None,
```
nikhilsuri-db, line 54: Will make it true in next PR post end to end testing

src/databricks/sql/common/unified_http_client.py
```diff
@@ -264,7 +264,22 @@ def request_context(
             yield response
         except MaxRetryError as e:
             logger.error("HTTP request failed after retries: %s", e)
-            raise RequestError(f"HTTP request failed: {e}")
+
+            # Try to extract HTTP status code from the MaxRetryError
```
nikhilsuri-db: setting http_code here which will be used by CircuitBreaker to consider regression in telemetry endpoint

src/databricks/sql/telemetry/circuit_breaker_manager.py
```diff
@@ -0,0 +1,194 @@
...
+        if new_state_name == CIRCUIT_BREAKER_STATE_OPEN:
+            logger.warning(LOG_CIRCUIT_BREAKER_OPENED, cb.name)
+        elif new_state_name == CIRCUIT_BREAKER_STATE_CLOSED:
+            logger.info(LOG_CIRCUIT_BREAKER_CLOSED, cb.name)
+        elif new_state_name == CIRCUIT_BREAKER_STATE_HALF_OPEN:
+            logger.info(LOG_CIRCUIT_BREAKER_HALF_OPEN, cb.name)
+
+
+@dataclass(frozen=True)
+class CircuitBreakerConfig:
```
nikhilsuri-db: Single object to store CB config

src/databricks/sql/telemetry/circuit_breaker_manager.py
```diff
@@ -0,0 +1,194 @@
...
+            name=f"{config.name}-{host}",
+        )
+
+        # Add state change listeners for logging
+        breaker.add_listener(CircuitBreakerStateListener())
+
+        return breaker
+
+    @classmethod
+    def _create_noop_circuit_breaker(cls) -> CircuitBreaker:
```
nikhilsuri-db: In cases where client config setup failed or used by test cases.

src/databricks/sql/telemetry/telemetry_push_client.py
```diff
@@ -0,0 +1,202 @@
...
+            host,
+        )
+
+    def _create_mock_success_response(self) -> BaseHTTPResponse:
+        """
+        Create a mock success response for when circuit breaker is open.
+
+        This allows telemetry to fail silently without raising exceptions.
+        """
+        from unittest.mock import Mock
```
samikshya-db: umm, using test imports in production code is intended here?

tests/unit/test_circuit_breaker_http_client.py
```diff
@@ -0,0 +1,218 @@
...
+            CircuitBreakerManager,
+            CircuitBreakerConfig,
+            DEFAULT_MINIMUM_CALLS as MINIMUM_CALLS,
+            DEFAULT_RESET_TIMEOUT as RESET_TIMEOUT,
+        )
+        import time
+
+        # Clear any existing state
+        CircuitBreakerManager._instances.clear()
+        CircuitBreakerManager.initialize(CircuitBreakerConfig())
```
samikshya-db: Do we need to add a similar initialization in main function too?

src/databricks/sql/auth/common.py
```diff
@@ -83,6 +84,7 @@ def __init__(
         self.pool_connections = pool_connections or 10
         self.pool_maxsize = pool_maxsize or 20
         self.user_agent = user_agent
+        self.telemetry_circuit_breaker_enabled = bool(telemetry_circuit_breaker_enabled)
```
samikshya-db, line 87: should we not add this to build_client_context too?

src/databricks/sql/telemetry/circuit_breaker_manager.py
```diff
@@ -0,0 +1,181 @@
...
+
+        Args:
+            host: The hostname for the circuit breaker
+
+        Returns:
+            New CircuitBreaker instance
+        """
+        config = cls._config
+        if config is None:
+            raise RuntimeError("CircuitBreakerManager not initialized")
```
samikshya-db: we shouldn't be throwing errors at all, I suppose?

src/databricks/sql/common/unified_http_client.py
```diff
@@ -264,7 +264,31 @@ def request_context(
...
+            if (
+                hasattr(e, "reason")
+                and e.reason is not None
+                and hasattr(e.reason, "response")
+                and e.reason.response is not None
+            ):
+                # The reason may contain a response object with status
+                http_code = getattr(e.reason.response, "status", None)
+            elif (
+                hasattr(e, "response")
```
samikshya-db: nit : these ~30 lines feels like it would live more happily in a helper

src/databricks/sql/common/unified_http_client.py
```diff
@@ -264,7 +264,31 @@ def request_context(
             yield response
         except MaxRetryError as e:
             logger.error("HTTP request failed after retries: %s", e)
-            raise RequestError(f"HTTP request failed: {e}")
+
+            # Try to extract HTTP status code from the MaxRetryError
+            http_code = None
+            if (
+                hasattr(e, "reason")
```
samikshya-db: great to see the defensive approach with urlLib object structure ! we have had several compatibility issue with this library before (I also remember a sev1 once), also can you double check if this is the expected structure once?

src/databricks/sql/telemetry/circuit_breaker_manager.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        Get or create a circuit breaker instance for the specified host.
+
+        Args:
+            host: The hostname for which to get the circuit breaker
+
+        Returns:
+            CircuitBreaker instance for the host
+        """
+        if not cls._config:
```
samikshya-db: Can we have this inside lock too?

src/databricks/sql/common/unified_http_client.py
samikshya-db: (I meant across versions. see [internal doc](https://docs.google.com/document/d/11vT9z-RCx-jwk0QxArg-j8wljxNF-16VqXAH7LdQKSs/edit?tab=t.0) for more context on the type of issues we get)

src/databricks/sql/auth/common.py
nikhilsuri-db: Why would we want this to be configurable from user connection params? We should not be giving any option to user to modify this flag.

src/databricks/sql/telemetry/telemetry_push_client.py
nikhilsuri-db: Moved this to a MockResponse Object instead Mock. Added test cases around  that also to assert this is parsed as expected.

src/databricks/sql/common/unified_http_client.py
nikhilsuri-db: Yes, miss from my sight. Thanks

src/databricks/sql/common/unified_http_client.py
nikhilsuri-db: Got it, the sev1 was that we started using a new method from the library which was not present in older versions of it. Here we are checking before accessing any attribute so should be fine. Anything else you think we need to worry about here?

src/databricks/sql/common/unified_http_client.py
```diff
@@ -28,6 +28,42 @@
 logger = logging.getLogger(__name__)
 
 
+def _extract_http_status_from_max_retry_error(e: MaxRetryError) -> Optional[int]:
```
nikhilsuri-db, line 31: When `total` is zero, urllib3 raises -> MaxRetryError
 We have configured the retry for 429 & 503 error codes in our retry.py

src/databricks/sql/auth/common.py
samikshya-db: This is a pattern we follow across drivers, add a private/undocumented flag just in case something breaks. We do not document all the flags that we offer. Once released, it is difficult to control if things go wrong.

src/databricks/sql/telemetry/circuit_breaker_manager.py
```diff
@@ -0,0 +1,112 @@
...
+
+import pybreaker
+from pybreaker import CircuitBreaker, CircuitBreakerError, CircuitBreakerListener
+
+from databricks.sql.exc import TelemetryNonRateLimitError
+
+logger = logging.getLogger(__name__)
+
+# Circuit Breaker Constants
+MINIMUM_CALLS = 20  # Number of failures before circuit opens
```
samikshya-db, line 21: are these numbers in-line with OSS JDBC too?

src/databricks/sql/telemetry/telemetry_push_client.py
```diff
@@ -0,0 +1,201 @@
...
+        Raises:
+            TelemetryRateLimitError: For 429/503 status codes (circuit breaker counts)
+            TelemetryNonRateLimitError: For other errors (circuit breaker excludes)
+        """
+        try:
+            response = self._delegate.request(method, url, headers, **kwargs)
+
+            # Check for rate limiting or service unavailable
+            if response.status in [429, 503]:
+                logger.warning(
```
samikshya-db, line 123: nit : all telemetry related errors/warning should not catch user's attention. They need to have the lower levels of logging

src/databricks/sql/auth/common.py
samikshya-db: Could you manually test this PR? (The initialization fix in your commit addressing a prior comment where the code would not have worked I suppose)

Here from manual testing : I mean atleast know that CB is coming into picture when this value is true.

src/databricks/sql/auth/common.py
nikhilsuri-db: As part of writing the e2e test, I have added the flag in context setup now

src/databricks/sql/telemetry/circuit_breaker_manager.py
nikhilsuri-db: Yes

src/databricks/sql/auth/common.py
nikhilsuri-db: Added a e2e test with patching the telemetry response and asserting the CB behaviour.

tests/e2e/test_circuit_breaker.py
```diff
@@ -0,0 +1,348 @@
...
+        """Get connection details from pytest fixture"""
+        self.arguments = connection_details.copy()
+
+    def test_circuit_breaker_opens_after_rate_limit_errors(self):
+        """
+        Verify circuit breaker opens after 429/503 errors and blocks subsequent calls.
+        """
+        request_count = {"count": 0}
+
+        def mock_rate_limited_request(*args, **kwargs):
```
samikshya-db: let's move this out of e2e, I can't think of having an end to end test without mocking.

tests/unit/test_telemetry_request_error_handling.py
```diff
@@ -0,0 +1,160 @@
...
+        """Test that RequestError with http-code=429 raises TelemetryRateLimitError."""
+        # Create RequestError with http-code in context
+        request_error = RequestError("HTTP request failed", context={"http-code": 429})
+        mock_delegate.request.side_effect = request_error
+
+        # Should raise TelemetryRateLimitError (circuit breaker counts it)
+        with pytest.raises(TelemetryRateLimitError):
+            client.request(HttpMethod.POST, "https://test.com", {})
+
+    def test_request_error_with_http_code_503_triggers_rate_limit_error(
```
samikshya-db: @pytest.mark.parametrize("status_code,expected_error", [
       (429, TelemetryRateLimitError),
       (503, TelemetryRateLimitError),
       (500, TelemetryNonRateLimitError),
   ]) - can we just do something like this?
   
   One thing about LLMs is they generate a lot of verbose/bulky tests. Can we fix that across this file?

tests/e2e/test_circuit_breaker.py
```diff
@@ -0,0 +1,348 @@
...
+import pytest
+from pybreaker import STATE_OPEN, STATE_CLOSED, STATE_HALF_OPEN
+from urllib3 import HTTPResponse
+
+import databricks.sql as sql
+from databricks.sql.telemetry.circuit_breaker_manager import CircuitBreakerManager
+
+
+@pytest.fixture(autouse=True)
+def aggressive_circuit_breaker_config():
```
samikshya-db, line 27: spent some time checking the unit tests and this e2e test. This is not adding something that we are not already testing tbh - or am I missing something? In which case we can remove this altogether.

samikshya-db (approved):
LGTM overall. I have some concerns with the unit/e2e tests added : added comments around it. Please push the PR once all changes are made. Thanks.

tests/unit/test_telemetry_request_error_handling.py
nikhilsuri-db: Agreed, was able to reduce the code by 37% keeping the  same test coverage

tests/e2e/test_circuit_breaker.py
nikhilsuri-db: I disagree, this E2E test verifies the full integration path from SQL query execution through telemetry generation to circuit breaker enforcement with real Databricks connections. It ensures that when the circuit breaker opens after rate-limit errors, subsequent telemetry requests are actually blocked from hitting Databricks, and that circuit recovery works correctly with real timing.

Can you elaborate why you think the tests are not adding anything to the existing test coverage?

tests/e2e/test_circuit_breaker.py
nikhilsuri-db: Only telemetry client is being mocked, everything else in a query execution life cycle is actually being triggered. Even the python code of telemetry is being executed and not wrapped around any mock.

tests/e2e/test_circuit_breaker.py
samikshya-db: > Only telemetry client is being mocked : 

[c1] we're intercepting calls before they ever reach the circuit breaker layer. This means none of the responses are occurring naturally. This means we are testing mock behavior. 
Circuit breaker transitions are already tested in `test_circuit_breaker_manager.py`, rate limit and error propogation are also tested in unit tests. I do not see a single use-case not being already tested via unit tests.

tests/e2e/test_circuit_breaker.py
samikshya-db: Same as c1. These test that telemetry events are created but not something different from unit tests in circuit breaker testing. `test_telemetry_circuit_breaker_integration.py` already has the testing for this.

tests/unit/test_telemetry_request_error_handling.py
samikshya-db: Thanks, looking forward to this across the test files.

nikhilsuri-db: I meant across all test files introduced ad part of this PR, was able to reduce 37% in total.

tests/e2e/test_circuit_breaker.py
nikhilsuri-db: I checked the unit tests - they test circuit breaker logic in isolation with mock functions (breaker.call(failing_func)), and telemetry client with direct API calls (_send_with_unified_client()). None of them execute real SQL queries (cursor.execute("SELECT 1")) or verify that telemetry is actually generated during query execution. The E2E test is the only place we validate the full journey: Real Query → Telemetry Generation → Circuit Breaker Enforcement. It also does not test that Circuit breaker state persists across multiple real SQL queries

tests/e2e/test_circuit_breaker.py
nikhilsuri-db: I think what is replaceable is `test_telemetry_circuit_breaker_integration.py` not the e2e tests. We can remove these tests, anyway these are units are incorrectly named as integration tests.

tests/e2e/test_circuit_breaker.py
samikshya-db: 1. we already have an e2e for telemetry with real queries `test_concurrent_telemetry.py`
2. What state transitions are you talking about other than ones in ` test_circuit_breaker_manager.py`?

I mean, i am ok with adding the e2e - I just don't want any redundant code in our repo - as long that it is not the case - I am good tbh. I have already approved the PR anyway.

nikhilsuri-db: I could generate a summary of use cases that will not be tested if we remove these test cases

```
# SCENARIO: You remove test_circuit_breaker.py

# What could break without detection:

1. Bug: Circuit breaker not invoked during SQL query execution
   - test_concurrent_telemetry.py: Won't catch (no circuit breaker testing)
   - test_circuit_breaker_manager.py: Won't catch (mocks, no real queries)
   - Result: Circuit breaker silently broken in production ❌

2. Bug: Circuit breaker opens but telemetry still sent
   - test_concurrent_telemetry.py: Won't catch (no error scenarios)
   - test_circuit_breaker_manager.py: Won't catch (no telemetry integration)
   - Result: Telemetry spam during outages ❌

3. Bug: 500 errors trigger circuit breaker (should only be 429/503)
   - test_concurrent_telemetry.py: Won't catch (no error scenarios)
   - test_circuit_breaker_manager.py: Won't catch (generic error testing)
   - Result: Circuit breaker too aggressive ❌

4. Bug: Circuit breaker config not passed through connection → telemetry
   - test_concurrent_telemetry.py: Won't catch (circuit breaker not tested)
   - test_circuit_breaker_manager.py: Won't catch (no connection flow)
   - Result: Circuit breaker never enabled ❌
   ```

samikshya-db: 1 and 4 is good. Although `(mocks, no real queries) and no telemetry integration` is ambiguous as the current e2e has mocks too. But 2, 3 still covered in tests (test_request_circuit_breaker_open , test_telemetry_request_error_handling)

PR #708, opened 2025-11-18
basic e2e test for force telemetry verification

nikhilsuri-db:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/.github/workflows/daily-telemetry-e2e.yml
+++ b/.github/workflows/daily-telemetry-e2e.yml
@@ -1,32 +1,42 @@
-name: Integration Tests
+name: Daily Telemetry E2E Tests
 
 on:
-  push:            
-    branches:
-      - main
-  pull_request:   
+  schedule:
+    - cron: '0 0 * * 0'  # Run every Sunday at midnight UTC
+  
+  workflow_dispatch:  # Allow manual triggering
+    inputs:
+      test_pattern:
+        description: 'Test pattern to run (default: tests/e2e/test_telemetry_e2e.py)'
+        required: false
+        default: 'tests/e2e/test_telemetry_e2e.py'
+        type: string
 
 jobs:
-  run-e2e-tests:
+  telemetry-e2e-tests:
     runs-on: ubuntu-latest
     environment: azure-prod
+    
     env:
       DATABRICKS_SERVER_HOSTNAME: ${{ secrets.DATABRICKS_HOST }}
       DATABRICKS_HTTP_PATH: ${{ secrets.TEST_PECO_WAREHOUSE_HTTP_PATH }}
       DATABRICKS_TOKEN: ${{ secrets.DATABRICKS_TOKEN }}
       DATABRICKS_CATALOG: peco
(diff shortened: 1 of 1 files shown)


tests/e2e/test_telemetry_e2e.py
```diff
@@ -0,0 +1,903 @@
+"""
+E2E test for telemetry - verifies telemetry behavior with different scenarios
```
samikshya-db, line 2: we shouldn't have this run for every PR push, that looks like a costly operation.

tests/e2e/test_telemetry_e2e.py
```diff
@@ -0,0 +1,903 @@
...
+        Default behavior:
+        - enable_telemetry defaults to False
+        - force_enable_telemetry defaults to False
+        - Telemetry will NOT be sent (NoopTelemetryClient used)
+        """
+        (
+            captured_events,
+            captured_futures,
+            export_wrapper,
+            callback_wrapper,
```
samikshya-db: let's use @pytest.mark.parametrize across this file, and follow DRY principle. This file is too bulky - this is a major drawback of LLMs. Test files should be readable, short and crisp.

nikhilsuri-db: Thanks for the suggestions. Was able to reduce the test size by 67%

tests/e2e/test_telemetry_e2e.py
nikhilsuri-db: How do you measure the cost of this? I ran some evaluation on my local, comparing with other integrations test as well. Seems like a ver light weight e2e test suite with only 12 queries getting executed and takes only 1 minute to cover all test cases. Given we are not using very heavy or large data set query, I am not that worried about the runtime cost for this execution. 
Let me know if you still think otherwise?

samikshya-db: These run against Azure Production (per workflow), not a test environment. 12 queries run across this test : for every PR push that too for a feature which is not core driver functionality is costly to me.  

we can do something like weekly run, or run only when certain telemetry paths are touched?

nikhilsuri-db: I do not think this is not a core functionality. We need to give telemetry equal ( if not higher ) position in our driver feature set. 

> we can do something like weekly run, or run only when certain telemetry paths are touched?

Do we have such setup or configuration for such sophisticated setup?

samikshya-db: This is nothing sophisticated imo, [we do this on JDBC](https://github.com/databricks/databricks-jdbc/blob/main/.github/workflows/vulnerabilityCatcher.yml#L4).

nikhilsuri-db: Updated to move it for daily run and remove it from normal e2e test suite execution

Issue #702, opened 2025-10-30
Auth token warning using oauth M2M


septponf:
This was originally posted in dbt-databricks project and I was asked to cross post i here.

Related to:  https://github.com/databricks/dbt-databricks/issues/1246

When running dbt in an Azure Devops Agent (for continues integration reasons) I get this warning:

```
09:13:52  Concurrency: 6 threads (target='test')
09:13:52  
09:13:53  databricks-sql-connector adapter: Token exchange failed, using external token: %s
WARNING:databricks.sql.auth.token_federation:Token exchange failed, using external token: 'access_token'
```
It still works, the project runs, but I would like to understand what might be happening here.
I am not able to reproduce this on my local machine using the same M2M profile.
Maybe you can point me in the right direction?

Profile using an azure service principal:
```
    test:
      host: the_host
      http_path: the_path
      catalog: the_catalog
      schema: test
      threads: 6
      type: databricks
      auth_type: oauth
      azure_client_id: "{{ env_var('DATABRICKS_CLIENT_ID') }}"
      azure_client_secret: "{{ env_var('DATABRICKS_CLIENT_SECRET') }}"
```

Python 3.11.13
Running with dbt=1.10.8
Registered adapter: databricks=1.10.9

The Devops Azure Agent runs an ubuntu-24.04 Image
Version: 20250929.60.1

msrathore-db:
The logs are regarding token exchange. Currently we try to exchange external Idp tokens for internal Databricks tokens and silently fail if it is not possible

calbert82uhah:
I get the same error message when I run 
(...)
```
        return sql.connect(
            server_hostname = st.secrets["DATABRICKS_SERVER_HOST"],
            http_path = st.secrets["DATABRICKS_HTTP_PATH"],
            access_token = entra_token
        )
```

`entra_token ` is a valid Entra Id token.
The connection works afterwards, but I do not understand why I get this warning on the console.

lorogno:
Still relevant!

PR #710, opened 2025-11-19
Fix token federation warnings by making it opt-in and improving error handling [Issue #702]

madhav-db:
## What type of PR is this?
Fixes token federation warnings that were appearing for users not using identity federation. Token federation is now opt-in and only enabled when `identity_federation_client_id` is explicitly provided. 

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Other

## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/src/databricks/sql/auth/auth.py
+++ b/src/databricks/sql/auth/auth.py
@@ -65,8 +65,8 @@ def get_auth_provider(cfg: ClientContext, http_client):
         else:
             raise RuntimeError("No valid authentication settings!")
 
-    # Always wrap with token federation (falls back gracefully if not needed)
-    if base_provider:
+    # Wrap with token federation only if explicitly enabled via identity_federation_client_id
+    if base_provider and cfg.identity_federation_client_id:
         return TokenFederationProvider(
             hostname=cfg.hostname,
             external_provider=base_provider,
--- a/src/databricks/sql/auth/token_federation.py
+++ b/src/databricks/sql/auth/token_federation.py
@@ -15,6 +15,24 @@ from databricks.sql.common.http import HttpMethod
 logger = logging.getLogger(__name__)
 
 
+class TokenFederationError(Exception):
+    """Base exception for token federation errors."""
+
+    pass
+
+
+class TokenExchangeNotAvailableError(TokenFederationError):
+    """Raised when token exchange endpoint is not available (404)."""
+
+    pass
+
+
+class TokenExchangeAuthenticationError(TokenFederationError):
+    """Raised when token exchange fails due to authentication issues (401/403)."""
+
+    pass
+
+
 class Token:
     """
     Represents an OAuth token with expiration management.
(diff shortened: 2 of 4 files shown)


src/databricks/sql/auth/auth.py
```diff
@@ -65,8 +65,8 @@ def get_auth_provider(cfg: ClientContext, http_client):
         else:
             raise RuntimeError("No valid authentication settings!")
 
-    # Always wrap with token federation (falls back gracefully if not needed)
-    if base_provider:
+    # Wrap with token federation only if explicitly enabled via identity_federation_client_id
+    if base_provider and cfg.identity_federation_client_id:
```
jprakash-db, line 69: For account wide token federation, this client id is not present? IIUC, this will completely skip the token federation

src/databricks/sql/auth/token_federation.py
```diff
@@ -143,9 +181,33 @@ def _get_token(self) -> Token:
...
+            except TokenExchangeNotAvailableError:
+                logger.debug(
+                    "Token exchange endpoint not available. Using external token directly. "
+                    "This is expected when token federation is not configured for this workspace."
+                )
+            except TokenExchangeAuthenticationError as e:
+                logger.warning(
+                    "Token exchange failed due to authentication error. Using external token directly. "
+                    "Error: %s. If this persists, verify your identity_federation_client_id configuration.",
+                    e,
```
jprakash-db, line 197: use lazy logging

src/databricks/sql/auth/token_federation.py
```diff
@@ -15,6 +15,24 @@
...
+
+    pass
+
+
+class TokenExchangeAuthenticationError(TokenFederationError):
+    """Raised when token exchange fails due to authentication issues (401/403)."""
+
+    pass
+
+
```
jprakash-db, line 35: Why are separate error classes being created for standard Http errors? Like 404, 401, 403 etc are just Http errors, so don't think need to create something different for them, just catch and log the normal errors

src/databricks/sql/auth/token_federation.py
```diff
@@ -184,15 +259,45 @@ def _exchange_token(self, access_token: str) -> Token:
...
-
-        return Token(
-            token_response["access_token"], token_response.get("token_type", "Bearer")
-        )
+        try:
+            response = self.http_client.request(
+                HttpMethod.POST, url=token_url, body=body, headers=headers
+            )
+
+            # Check response status code
```
jprakash-db, line 267: Can you use the response with the with context, to handler auto close.

src/databricks/sql/auth/token_federation.py
```diff
@@ -184,15 +259,45 @@ def _exchange_token(self, access_token: str) -> Token:
...
+            if response.status == 404:
+                raise TokenExchangeNotAvailableError(
```
jprakash-db, line 269: No need to manually handle everything just do `response.raise_for_status`,  it auto handles all these basic errors

src/databricks/sql/auth/token_federation.py
```diff
@@ -184,15 +259,45 @@ def _exchange_token(self, access_token: str) -> Token:
...
+                    "Token exchange endpoint not found. Token federation may not be enabled for this workspace."
+                )
+            elif response.status in (401, 403):
+                error_detail = (
+                    response.data.decode() if response.data else "No error details"
+                )
+                raise TokenExchangeAuthenticationError(
```
jprakash-db, line 276: Same comment, basic http errors don't need to be handled, just do `response.raise_for_status`

src/databricks/sql/auth/token_federation.py
```diff
@@ -72,8 +90,20 @@ class TokenFederationProvider(AuthProvider):
...
+    Token federation is useful for:
+    - Cross-cloud authentication scenarios
+    - Unity Catalog access across Azure subscriptions
+    - Service principal authentication with external identity providers
+
+    The provider automatically detects when token exchange is needed by comparing the
+    token issuer with the Databricks workspace hostname. If exchange fails, it gracefully
+    falls back to using the external token directly.
+
+    Note: Token federation must be explicitly enabled by providing the
```
jprakash-db, line 105: How does it work with account wide token federation then?

samikshya-db:
@madhav-db / @jprakash-db  - do we plan to merge this?

PR #715, opened 2025-11-24
perf: Optimize telemetry latency logging to reduce overhead

samikshya-db:
## Changes
1. **Eliminated extractor pattern** - Replaced wrapper classes (TelemetryExtractor, CursorExtractor, ResultSetDownloadHandlerExtractor) with direct attribute access functions
2. **Feature flag early exit** - Checks cached `telemetry_enabled` flag to skip heavy work when telemetry is disabled

## Performance Impact
Internal doc : https://docs.google.com/document/d/1sV_N3IXhzH5aAoAOOhc4sA49ZdINGii_wXvYNyZdVEo/edit?tab=t.0#heading=h.ezkw4po8o1mx 

~ 8% improvement across test cases.

## Behavior Changes
- The decorator now always logs latency at DEBUG level for debugging purposes, even when telemetry is disabled
- Exits early using cached `connection.telemetry_enabled` flag (avoids dictionary lookup + instance check on every operation)
- Only performs data extraction and object creation when telemetry is actually enabled

### DIFF
--- a/src/databricks/sql/telemetry/telemetry_client.py
+++ b/src/databricks/sql/telemetry/telemetry_client.py
@@ -415,9 +415,9 @@ class TelemetryClientFactory:
     It uses a thread pool to handle asynchronous operations and a single flush thread for all clients.
     """
 
-    _clients: Dict[str, BaseTelemetryClient] = (
-        {}
-    )  # Map of session_id_hex -> BaseTelemetryClient
+    _clients: Dict[
+        str, BaseTelemetryClient
+    ] = {}  # Map of session_id_hex -> BaseTelemetryClient
     _executor: Optional[ThreadPoolExecutor] = None
     _initialized: bool = False
     _lock = threading.RLock()  # Thread safety for factory operations
@@ -518,21 +518,21 @@ class TelemetryClientFactory:
                         session_id_hex,
                     )
                     if telemetry_enabled:
-                        TelemetryClientFactory._clients[session_id_hex] = (
-                            TelemetryClient(
-                                telemetry_enabled=telemetry_enabled,
-                                session_id_hex=session_id_hex,
-                                auth_provider=auth_provider,
-                                host_url=host_url,
-                                executor=TelemetryClientFactory._executor,
-                                batch_size=batch_size,
-                                client_context=client_context,
-                            )
+                        TelemetryClientFactory._clients[
+                            session_id_hex
+                        ] = TelemetryClient(
+                            telemetry_enabled=telemetry_enabled,
+                            session_id_hex=session_id_hex,
+                            auth_provider=auth_provider,
+                            host_url=host_url,
+                            executor=TelemetryClientFactory._executor,
+                            batch_size=batch_size,
+                            client_context=client_context,
                         )
                     else:
-                        TelemetryClientFactory._clients[session_id_hex] = (
-                            NoopTelemetryClient()
-                        )
+                        TelemetryClientFactory._clients[
+                            session_id_hex
+                        ] = NoopTelemetryClient()
         except Exception as e:
             logger.debug("Failed to initialize telemetry client: %s", e)
             # Fallback to NoopTelemetryClient to ensure connection doesn't fail


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "databricks-sql-connector"
-version = "4.2.0"
+version = "4.2.1"
```
nikhilsuri-db, line 3: Why the upgrade part of this PR?

src/databricks/sql/backend/sea/backend.py
```diff
@@ -157,7 +157,7 @@ def __init__(
             "_use_arrow_native_complex_types", True
         )
 
-        self.use_hybrid_disposition = kwargs.get("use_hybrid_disposition", True)
```
nikhilsuri-db: I think this is not the right view of diff from master branch?

src/databricks/sql/common/feature_flag.py
```diff
@@ -165,8 +165,9 @@ def get_instance(cls, connection: "Connection") -> FeatureFlagsContext:
             cls._initialize()
             assert cls._executor is not None
 
-            # Use the unique session ID as the key
-            key = connection.get_session_id_hex()
+            # Cache at HOST level - share feature flags across connections to same host
```
nikhilsuri-db, line 168: Can you please share your findings on what improved post this change? Do we see a memory drop after caching it at host level?

src/databricks/sql/common/feature_flag.py
```diff
@@ -165,8 +165,9 @@ def get_instance(cls, connection: "Connection") -> FeatureFlagsContext:
...
+            # Feature flags are per-host, not per-session
+            key = connection.session.host
```
nikhilsuri-db, line 170: Is this similar to how we cache feature flag values in Java?

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -143,11 +157,10 @@ def log_latency(statement_type: StatementType = StatementType.NONE):
     data about the operation, including latency, statement information, and
     execution context.
 
-    The decorator automatically:
-    - Measures execution time using high-precision performance counters
-    - Extracts telemetry information from the method's object (self)
-    - Creates a SqlExecutionEvent with execution details
-    - Sends the telemetry data asynchronously via TelemetryClient
+    OPTIMIZATIONS APPLIED:
```
nikhilsuri-db: nit; remove `OPTIMIZATIONS APPLIED:`

src/databricks/sql/telemetry/latency_logger.py
```diff
@@ -143,11 +157,10 @@ def log_latency(statement_type: StatementType = StatementType.NONE):
...
+    - Uses time.monotonic() instead of time.perf_counter() for faster timing
```
nikhilsuri-db: Reading python and some online articles, it says both of them use same lock internally so there should not be any performance gaps & the doc prefers to use time,perf_counter for short durations capturing https://peps.python.org/pep-0418/

Can you share the details of your finding with and without this specific change?

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -109,18 +110,21 @@ def get_auth_flow(auth_provider):
 
     @staticmethod
     def is_telemetry_enabled(connection: "Connection") -> bool:
+        # Fast path: force enabled - skip feature flag fetch entirely
```
nikhilsuri-db, line 118: Nice!!! We should always aim for early exit from a method 👍

src/databricks/sql/common/feature_flag.py
nikhilsuri-db: Can we test this change via unit tests?

nikhilsuri-db (commented):
Few comments to understand intention of each change but LGTM overall.

src/databricks/sql/telemetry/latency_logger.py
samikshya-db: Yes, this is correct actually. 0.93939384 seconds gets rounded off the 0.939 itself - i would prefer not using the precision which is not required. I will edit the text above.

src/databricks/sql/common/feature_flag.py
samikshya-db: > Is this similar to how we cache feature flag values in Java?

Yes, this is the key we use in java now (updated after the RL issue). Let me also add UTs. Thanks

src/databricks/sql/common/feature_flag.py
samikshya-db: This is in-general optimisation - Replicating what we have in jdbc. I did not test memory changes post this

PR #723, opened 2025-12-26
Fixed the exception handler close() on _TelemetryClientHolder 

msrathore-db:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Other

## Description
Fixed the exception handler close() on _TelemetryClientHolder objects instead of accessing the client inside them.
## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [x] Manually
- [ ] N/A

### DIFF
--- a/src/databricks/sql/telemetry/telemetry_client.py
+++ b/src/databricks/sql/telemetry/telemetry_client.py
@@ -542,8 +542,8 @@ class TelemetryClientFactory:
         logger.debug("Handling unhandled exception: %s", exc_type.__name__)
 
         clients_to_close = list(cls._clients.values())
-        for client in clients_to_close:
-            client.close()
+        for holder in clients_to_close:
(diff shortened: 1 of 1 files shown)


nikhilsuri-db:
Should we add more configuration like `check_untyped_defs` `disallow_untyped_defs` `warn_return_any` `warn_unused_ignores` in mypy tool to identify such issues early or at least let the developer know?

samikshya-db:
Sorry, might not be aware of this - Can you give more context on this change and how was this caught?

msrathore-db:
> Sorry, might not be aware of this - Can you give more context on this change and how was this caught?

Was testing something with python connector but got an error from the server, which led to seeing this exception. The close() didn't exist in the class so made this small fix

Issue #709, opened 2025-11-18
ERROR:databricks.sql.common.unified_http_client:HTTP request error: Retry request would exceed Retry policy max retry duration of 300 seconds


jimsan6817:
My Code:
```
conn = sql.connect(server_hostname = server_hostname, http_path = http_path, access_token = access_token, _tls_no_verify=True)
cursor = conn.cursor()
select_sql="select * from tableA"
cursor.execute(select_sql)
```

Then console show error as follows:

cursor.execute(select_sql) start :  2025-11-18 09:11:09

> **2025-11-18 09:12:18,717 WARNING:urllib3.connectionpool:Retrying (DatabricksRetryPolicy(total=9, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7f3cb45eb610>, 'Connection to dbstoragejjgoq3xz6zawu.blob.core.windows.net timed out. (connect timeout=60)')': /jobs/8235627411498389/sql/2025-11-18/01/results_2025-11-18T01:11:13Z_4c427180-1208-4c20-924a-f36a81468ec2?sig=<REDACTED>&se=<REDACTED>&sv=<REDACTED>&spr=<REDACTED>&sp=<REDACTED>&sr=<REDACTED>
> 
> 2025-11-18 09:13:26,777 WARNING:urllib3.connectionpool:Retrying (DatabricksRetryPolicy(total=8, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7f3cb45eaf10>, 'Connection to dbstoragejjgoq3xz6zawu.blob.core.windows.net timed out. (connect timeout=60)')': /jobs/8235627411498389/sql/2025-11-18/01/results_2025-11-18T01:11:13Z_4c427180-1208-4c20-924a-f36a81468ec2?sig=<REDACTED>&se=<REDACTED>&sv=<REDACTED>&spr=<REDACTED>&sp=<REDACTED>&sr=<REDACTED>**

And Finally Got this error:

> **2025-11-18 09:19:18,996 ERROR:databricks.sql.common.unified_http_client:HTTP request error: Retry request would exceed Retry policy max retry duration of 300 seconds**

May I ask, what are the possible solutions to fix this kind of error?
--------------------------------------------------

> urllib3 v2.2.3
> databricks-sql-python v4.1.4

[...]

PR #724, opened 2025-12-30
created util method to normalise http protocol in http path

nikhilsuri-db:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Other

## Description
Problem
Feature flag and telemetry endpoints were double-prepending https:// when server_hostname already included the protocol, causing https://https://host.com URLs. This broke telemetry for dbt-databricks users who configure hostnames with https:// prefix.

Solution
Added normalize_host_with_protocol() utility to handle protocol normalization
Updated feature flag and telemetry clients to use: normalize_host_with_protocol(host) + "/path"
Handles hostnames with/without protocol and trailing slashes

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

https://github.com/databricks/databricks-sql-python/issues/709

### DIFF
--- a/src/databricks/sql/common/url_utils.py
+++ b/src/databricks/sql/common/url_utils.py
@@ -5,7 +5,7 @@ URL utility functions for the Databricks SQL connector.
 
 def normalize_host_with_protocol(host: str) -> str:
     """
-    Normalize a connection hostname by ensuring it has a protocol and removing trailing slashes.
+    Normalize a connection hostname by ensuring it has a protocol.
 
     This is useful for handling cases where users may provide hostnames with or without protocols
     (common with dbt-databricks users copying URLs from their browser).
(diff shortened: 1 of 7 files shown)


src/databricks/sql/common/url_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+def normalize_host_with_protocol(host: str) -> str:
+    """
+    Normalize a connection hostname by ensuring it has a protocol and removing trailing slashes.
+
+    This is useful for handling cases where users may provide hostnames with or without protocols
+    (common with dbt-databricks users copying URLs from their browser).
+
+    Args:
+        host: Connection hostname which may or may not include a protocol prefix (https:// or http://)
```
samikshya-db, line 14: nit : case insensitivity should be considered

src/databricks/sql/common/url_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Returns:
+        Normalized hostname with protocol prefix and no trailing slash
+
+    Examples:
+        normalize_host_with_protocol("myserver.com") -> "https://myserver.com"
+        normalize_host_with_protocol("https://myserver.com") -> "https://myserver.com"
+    """
+    # Remove trailing slash
+    host = host.rstrip("/")
```
samikshya-db, line 34: nit : check for NPE

src/databricks/sql/common/url_utils.py
```diff
@@ -0,0 +1,44 @@
+"""
+URL utility functions for the Databricks SQL connector.
+"""
+
+
+def normalize_host_with_protocol(host: str) -> str:
```
samikshya-db, line 6: Was rechecking this piece of code : auth_utils.py and thrift_backend - Has proper check on this already. should we use this util instead?  

Also, the sea flow looks incorrect at the moment : backend/sea/utils/http_client.py

nikhilsuri-db: Good point, we can consolidate the auth_util to url_utils 👍  but thrift_backend seems very inline modification, do not want to touch it in current PR scope. 


Regarding Sea - Fixed by using the new url_utils

PR #730, opened 2026-01-22
fix: Adding Python shutdown check in _close to fix sys.meta_path is None error

Bidek56:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Other

## Description
This PR adds Python shutdown check in _close function to avoid this error:
`ERROR:Attempt to close session raised a local exception: sys.meta_path is None, Python is likely shutting down`

## How is this tested?

- [X ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/src/databricks/sql/session.py
+++ b/src/databricks/sql/session.py
@@ -211,6 +211,7 @@ class Session:
                     "Attempt to close session raised an exception at the server: %s", e
                 )
         except Exception as e:
-            logger.error("Attempt to close session raised a local exception: %s", e)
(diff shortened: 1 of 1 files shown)


msrathore-db:
> @msrathore-db Please review this PR, it resolves the following error that we are seeing. Thx `Attempt to close session raised a local exception: sys.meta_path is None, Python is likely shutting down`

When are you observing this error. Can you provide repro steps?

PR #734, opened 2026-01-29
[PECOBLR-1735] Fix #729 and #731: Telemetry lifecycle management

### LINKED ISSUE #729: ERROR: Rapid fire connection error without exit
We have a CICD test running from a simple python script in a git action. Authenticating with a service principal pat token against a Databricks warehouse.

Since the release of the most recent version we have been experiencing extended intermittent failures.

The git action output prints out the error Service Principal does not have permission to use warehouse. This error repeats around every 20 seconds. Yesterday we were unable to cancel the git action while this connect to warehouse step was occurring and it repeated printing out that error for about two hours.

There are no loops in the script, it executes a single databricks query after initializing a single connection to a warehouse.

Once git eventually let us cancel the action we reran the action without changing any permissions or touching Databricks at all and it connected and executed as expected in under 30 seconds, no errors reported.

The issue has occurred again today but it did not prevent us cancelling the git action, and again today after restarting the git action it executed as expected.

Are there any changes we should be aware of that would allow us to prevent this happening in the future.

### LINKED ISSUE #731: Sporadic `'NoneType' object has no attribute 'request'` error from telemetry — usage question
### Summary

We're seeing sporadic errors logged from what appears to be the telemetry subsystem. Queries complete successfully, but we'd like to understand if we're missing a recommended usage pattern or if this is a known issue.

### Environment

- **databricks-sql-connector version**: 4.2.4
- **Python version**: 3.13
- **OS**: Linux (Ubuntu)
- **pyarrow installed**: No

### Error Message

```
ERROR:databricks.sql.common.unified_http_client:HTTP request error: 'NoneType' object has no attribute 'request'
```

**Frequency**: Sporadic, timing-dependent  
**Impact**: Non-fatal (queries complete successfully)

### Our Usage Pattern

We're using short-lived connections with the context manager:

```python
with sql.connect(
    server_hostname=host,
    http_path=http_path,
    access_token=token,
) as connection:
    with connection.cursor() as cursor:
        cursor.execute(query)
        return cursor.fetchall()
```

### Our Investigation

We traced through the code and believe the error originates from telemetry, since:

- We don't have pyarrow installed (Cloud Fetch disabled)
- We're not using staging operations
- Thrift RPC uses its own connection pools, not `UnifiedHttpClient`
- Telemetry is the only remaining `UnifiedHttpClient` consumer in our setup

We noticed that `TelemetryClient.close()` flushes events but doesn't close `_http_client`, and wondered if there might be a race condition when async telemetry tasks are still in-flight after the connection closes.

### Questions

1. **Is there a recommended pattern for short-lived connections that avoids this?** For example, should we be keeping connections open longer, or using a connection pool?

2. **Is disabling telemetry the intended workaround?** We've confirmed that `enable_telemetry=False` eliminates the error, but wanted to check if this is the right approach or if we're masking an underlying usage issue.

3. **Should `TelemetryClient.close()` be closing `_http_client`?** If this is indeed a bug, we're happy to submit a PR — just want to confirm we're understanding the intended lifecycle correctly.

### Workaround (current)

```python
with sql.connect(
    server_hostname=host,
    http_path=http_path,
    access_token=token,
    enable_telemetry=False,
) as connection:
    ...
```

Thanks for any guidance!

msrathore-db:
<!-- We welcome contributions. All patches must include a sign-off. Please see CONTRIBUTING.md for details -->


## What type of PR is this?
<!-- Check all that apply, delete what doesn't apply. -->

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Other

## Description
  Issue #729: Connection failures hang CICD jobs for 15+ minutes, cannot be cancelled                                                                                                                       
  - Telemetry forced on even with enable_telemetry=False                                                                                                                                                    
  - 900-second timeout causes long hangs                                                                                                                                                                    
  - Blocking shutdown prevents process exit                                                                                                                                                                 
                                                                                                                                                                                                            
  Issue #731: Race condition causes AttributeError: 'NoneType' object has no attribute 'request'                                                                                                            
  - Async telemetry executes after _http_client is garbage collected                                                                                                                                        
  - No null check before calling pool_manager.request()                                                                                                                                                     
                                                                                                                                                                                                            
  ## Changes                                                                                                                                                                                                   
                                                                                                                                                                                                            
  ### Issue #729 (3 fixes)                                                                                                                                                                                      
                                                                                                                                                                                                            
  1. Respect enable_telemetry parameter (telemetry_client.py:722-729, client.py:321)                                                                                                                        
    - Added parameter to connection_failure_log() with early return if disabled                                                                                                                             
    - Connection passes user's preference through                                                                                                                                                           
  2. Reduce timeout to 30s (telemetry_client.py:299)                                                                                                                                                        
    - Changed from 900s to 30s for faster failure                                                                                                                                                           
  3. Non-blocking shutdown (telemetry_client.py:707)                                                                                                                                                        
    - Changed executor.shutdown(wait=True) to wait=False                                                                                                                                                    
                                                                                                                                                                                                            
  ### Issue #731 (3 fixes)                                                                                                                                                                                      
                                                                                                                                                                                                            
  4. Null-safety check (unified_http_client.py:296-298)                                                                                                                                                     
    - Check if pool_manager is None before calling .request()                                                                                                                                               
    - Raise RequestError with clear message                                                                                                                                                                 
  5. Add __del__ cleanup (telemetry_client.py:439-451)                                                                                                                                                      
    - Close _http_client during garbage collection                                                                                                                                                          
    - Ensures eventual resource cleanup                                                                                                                                                                     
  6. Document lifecycle (telemetry_client.py:430-435)                                                                                                                                                       
    - Added docstring explaining why _http_client isn't closed in close()                                                                                                                                   
    - Import RequestError from databricks.sql.exc                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
  ## Impact                                                                                                                                                                                                    
                                                                                                                                                                                                            
  - Connection failures exit in <1s instead of 15+ minutes                                                                                                                                                  
  - CICD jobs no longer hang                                                                                                                                                                                
  - No more race condition AttributeError                                                                                                                                                                   
  - enable_telemetry=False now respected in all scenarios                                                                                                                                                   
                                                                                                                                                                                                            
  Backward Compatibility                                                                                                                                                                                    
                                                                                                                                                                                                            
  Fully backward compatible. No API changes, only behavior improvements.                                                                                                                                    
                                                                                                                                                                                                            
  Closes #729                                                                                                                                                                                               
  Closes #731   
## How is this tested?

- [ ] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

<!-- If Manually, please describe. -->

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -87,22 +87,6 @@ jobs:
             -v
 
       #----------------------------------------------
-      #    run serial tests (excluding telemetry) with coverage
-      #----------------------------------------------
-      - name: Run serial tests (excluding telemetry) with coverage
-        continue-on-error: false
-        run: |
-          poetry run pytest tests/e2e \
-            -m "serial" \
-            --ignore=tests/e2e/test_telemetry_e2e.py \
-            --ignore=tests/e2e/test_concurrent_telemetry.py \
-            --cov=src \
-            --cov-append \
-            --cov-report=xml \
-            --cov-report=term \
-            -v
-
-      #----------------------------------------------
       #    run telemetry tests with coverage (isolated)
       #----------------------------------------------
       - name: Run telemetry tests with coverage (isolated)


src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -295,7 +296,7 @@ def _send_telemetry(self, events):
                 url,
                 data=request.to_json(),
                 headers=headers,
-                timeout=900,
+                timeout=30,
```
samikshya-db: There are retries etc that incorporate backoff, the push is anyway async, we should not be changing this in my opinion.

src/databricks/sql/telemetry/telemetry_client.py
```diff
@@ -689,13 +719,19 @@ def connection_failure_log(
...
+        # Respect user's telemetry preference - don't force-enable
+        if not enable_telemetry:
+            logger.debug("Telemetry disabled, skipping connection failure log")
+            return
+
         UNAUTH_DUMMY_SESSION_ID = "unauth_session_id"
 
         TelemetryClientFactory.initialize_telemetry_client(
-            telemetry_enabled=True,
+            telemetry_enabled=enable_telemetry,
```
samikshya-db: nit : this change is not required.

src/databricks/sql/telemetry/telemetry_client.py
msrathore-db: Sure. Reverting this change

src/databricks/sql/telemetry/telemetry_client.py
msrathore-db: Makes sense. Could've kept for defensive approach but reverted

PR #735, opened 2026-02-04
[ES-1717039] Fix 60 seconds delay in gov cloud connections + Fix PR check failures in the repo

samikshya-db:
## What type of PR is this?

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Other

## Description

Customer Issue : There was delay of 60 seconds when user from gov-cloud connects to the driver. This was because connector service is still not deployed to gov-cloud. And, the retry mechanism indicated feature flag endpoint to retry 404 too.  There were 5 retries with ~10 second wait for each. 

- Bug 1 : There are different behavior for `404` error code in each command types- in case of `OTHER` (i.e., commands other than execute, getoeration, close etc.) - we retry even for 404! Hence - In case of featureFlag endpoint, we were retrying for 404 too. This is not expected and we should not retry for 404.
- Bug 2 : [Unsure if this is a bug, I am not fixing this in the PR, but will start a discussiong with the team]. backoff calculation should have been 1s, 2s, 4s, 8s, etc. but in our code we force a minumum 10 second wait, this causes ~60 second delay rather than a ~25 second delay.

Also, note : this PR fixes github action failures because of a flaky new poetry version

## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/.github/workflows/publish-test.yml
+++ b/.github/workflows/publish-test.yml
@@ -64,7 +64,12 @@ jobs:
       - name: Build package
         run: poetry build
       #----------------------------------------------
+      # Configure test-pypi repository
+      #----------------------------------------------
+      - name: Configure test-pypi repository
+        run: poetry config repositories.testpypi https://test.pypi.org/legacy/
+      #----------------------------------------------
       # Attempt push to test-pypi
       #----------------------------------------------
-      - name: Publish to pypi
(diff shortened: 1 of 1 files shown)


vikrantpuppala (approved):
thank you!

Issue #741, opened 2026-02-24
dbsql_parameter_from_primitive maps Python float to FloatParameter (32-bit) instead of DoubleParameter (64-bit), causing silent precision loss


Shubhambhusate:
**Description:**

Python's native float type is a 64-bit IEEE 754 double-precision number. However, the dbsql_parameter_from_primitive() function in src/databricks/sql/parameters/native.py (line 662) maps it to FloatParameter, which casts the value to a Databricks SQL FLOAT (32-bit single-precision). This causes silent precision loss when using parameterized queries with float values.

The correct mapping should be DoubleParameter, which corresponds to Databricks SQL DOUBLE (64-bit) and matches Python's native float precision.

**Steps to reproduce:**

```
from databricks import sql
from databricks.sql.parameters import DoubleParameter

test_val = 12345.678901234

with connection.cursor() as cursor:
    cursor.execute("DROP TABLE IF EXISTS test_precision")
    cursor.execute("CREATE TABLE test_precision (val DOUBLE)")

    # BUG: raw float silently loses precision (uses FloatParameter → 32-bit)
    cursor.execute("INSERT INTO test_precision (val) VALUES (?)", [test_val])
    cursor.execute("SELECT val FROM test_precision")
    stored_bug = cursor.fetchone()[0]

    print(f"Input:  {test_val}")       # 12345.678901234
    print(f"Stored: {stored_bug}")     # 12345.6787109375 (precision lost!)
    print(f"Match:  {test_val == stored_bug}")  # False
```

**Expected Behavior**

Passing a Python float to a parameterized query should preserve 64-bit precision by default, since Python's float is inherently 64-bit.

**Actual Behavior**

Precision is silently truncated to 32-bit because dbsql_parameter_from_primitive() returns FloatParameter instead of DoubleParameter.

jprakash-db:
Closing the issue, since the PR is merged

PR #742, opened 2026-02-24
Fix float inference to use DoubleParameter (64-bit) instead of FloatP…

Shubhambhusate:
…arameter (32-bit)

## What type of PR is this?

- [ ] Bug Fix

## Description

dbsql_parameter_from_primitive() in src/databricks/sql/parameters/native.py currently maps Python's float type to FloatParameter, which binds values as Databricks SQL FLOAT (32-bit single-precision). However, Python's float is a 64-bit IEEE 754 double-precision number, so this mapping silently truncates precision.
This PR changes the inference to use DoubleParameter instead, which binds values as Databricks SQL DOUBLE (64-bit) and correctly preserves the full precision of Python floats.

Before (buggy):
```
cursor.execute("INSERT INTO t (val) VALUES (?)", [12345.678901234])
# Stored as 12345.0 — precision lost silently
```

After (fixed):

```
cursor.execute("INSERT INTO t (val) VALUES (?)", [12345.678901234])
# Stored as 12345.678901234 — full precision preserved
```

## How is this tested?

- [ ] Unit tests
- [ ] Manually

All 55 unit tests in tests/unit/test_parameters.py pass, including test_inference[DoubleParameter-Primitive.FLOAT]. 

Manually verified that parameterized queries with raw Python floats now preserve full 64-bit precision when inserted into DOUBLE columns.

### DIFF
--- a/tests/unit/test_parameters.py
+++ b/tests/unit/test_parameters.py
@@ -295,6 +295,7 @@ class TestDbsqlParameter:
             (BigIntegerParameter, Primitive.BIGINT),
             (BooleanParameter, Primitive.BOOL),
             (DateParameter, Primitive.DATE),
+            (DoubleParameter, Primitive.DOUBLE),
             (DoubleParameter, Primitive.FLOAT),
             (VoidParameter, Primitive.NONE),
             (TimestampParameter, Primitive.TIMESTAMP),


Shubhambhusate:
Issue link: https://github.com/databricks/databricks-sql-python/issues/741

Shubhambhusate:
Hi @msrathore-db Can you please review the PR

Shubhambhusate:
Hi @jprakash-db Can you please review this PR?

tests/unit/test_parameters.py
```diff
@@ -295,7 +295,7 @@ def test_tspark_param_ordinal(self):
             (BigIntegerParameter, Primitive.BIGINT),
             (BooleanParameter, Primitive.BOOL),
             (DateParameter, Primitive.DATE),
-            (FloatParameter, Primitive.FLOAT),
+            (DoubleParameter, Primitive.FLOAT),
```
gopalldb, line 299: we can also add to the test:
(DoubleParameter, Primitive.DOUBLE)

gopalldb (commented):
A minor comment on test, will also let Jothi also take a look

tests/unit/test_parameters.py
Shubhambhusate: Please let me know if you want to me make the suggested changes in the code

PR #743, opened 2026-02-26
[PECO-AI] PECOBLR-1968: Add warnings in logs if someone uses a connection param that is not defined in python sql connector

deeksha-db:
**PECO Agent Workflow** (Partner Ecosystem)

**Jira Ticket:** https://databricks.atlassian.net/browse/PECOBLR-1968

## Changes
Added a `KNOWN_KWARGS` frozenset constant to `client.py` listing all supported `**kwargs` for `Connection.__init__`. In the constructor, right after the debug log, the code computes `set(kwargs.keys()) - KNOWN_KWARGS` and emits a `logger.warning` naming any unrecognized parameter(s). This is a purely additive, backward-compatible change — unknown kwargs continue to be passed through; they just now generate a visible warning.

## Test Cases Written
- `test_unknown_connection_param_issues_warning` — unknown kwarg triggers `logger.warning`
- `test_unknown_connection_param_warning_names_the_param` — warning message includes the param name
- `test_known_connection_pa...

*Summary:* Added a `KNOWN_KWARGS` frozenset constant to `client.py` listing all supported `**kwargs` for `Connection.__init__`. In the constructor, right after the debug log, the code computes `set(kwargs.keys()) - KNOWN_KWARGS` and emits a `logger.warning` naming any unrecognized parameter(s). This is a purely additive, backward-compatible change — unknown kwargs continue to be passed through; they just now generate a visible warning.

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -93,6 +93,78 @@ NO_NATIVE_PARAMS: List = []
 # Transaction isolation level constants (extension to PEP 249)
 TRANSACTION_ISOLATION_LEVEL_REPEATABLE_READ = "REPEATABLE_READ"
 
+# All supported **kwargs for Connection.__init__. Used to warn on unknown params.
+KNOWN_KWARGS = frozenset(
+    [
+        # Authentication
+        "access_token",
+        "auth_type",
+        "username",
+        "password",
+        "credentials_provider",
+        "_use_cert_as_auth",
+        "oauth_client_id",
+        "oauth_redirect_port",
+        "oauth_scopes",
+        "experimental_oauth_persistence",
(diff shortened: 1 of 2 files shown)


src/databricks/sql/client.py
```diff
@@ -93,6 +93,78 @@
 # Transaction isolation level constants (extension to PEP 249)
 TRANSACTION_ISOLATION_LEVEL_REPEATABLE_READ = "REPEATABLE_READ"
 
+# All supported **kwargs for Connection.__init__. Used to warn on unknown params.
+KNOWN_KWARGS = frozenset(
```
samikshya-db, line 97: we should add a test `test_known_kwargs_is_complete`

PR #749, opened 2026-03-04
[QI-3367] Allow specifiying query tags as a dict upon connection creation

jiabin-hu:
## What type of PR is this?

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Other

## Description
For session level query tags,
1. adding the option to specify session query tags as a dict rather than a serialized string, to be consistent w/ statement level
2. if specified both as dict param, and serialized string as session param, dict param takes precedence, overwriting everything in session param
## How is this tested?

- [x] Unit tests
- [ ] E2E Tests
- [ ] Manually
- [ ] N/A

### DIFF
--- a/examples/query_tags_example.py
+++ b/examples/query_tags_example.py
@@ -17,7 +17,7 @@ Example: {"team": "engineering", "application": "etl", "priority": "high"}
 Special cases:
 - If a value is None, only the key is included (no colon or value)
 - Special characters (comma, colon and backslash) in values are automatically escaped
-- Backslashes in keys are automatically escaped; other special characters in keys are not escaped
+- Backslashes in keys are automatically escaped; other special characters in keys are not allowed
 """
 
 print("=== Query Tags Example ===\n")


src/databricks/sql/client.py
```diff
@@ -281,6 +283,15 @@ def read(self) -> Optional[OAuthToken]:
...
             ] = "true"
 
+        if query_tags is not None:
+            if session_configuration is None:
+                session_configuration = {}
+            serialized = serialize_query_tags(query_tags)
+            if serialized:
+                session_configuration["QUERY_TAGS"] = serialized
+            else:
+                session_configuration.pop("QUERY_TAGS", None)
```
sreekanth-db, line 293: should we add some debug log here about serialization failure ?

jiabin-hu: this is not a failure , it's when user say `query_tags = {}` (empty dict)

PR #756, opened 2026-03-18
Add `_retry_server_directed_only` mode for Retry-After header compliance

sd-db:
## Summary

- Adds a new opt-in `_respect_server_retry_after_header` connection parameter that restricts retries to only occur when the server includes a `Retry-After` header in the response
- Prevents duplicate side effects for non-idempotent `ExecuteStatement` operations where the server has not explicitly signaled that retry is safe
- Threads the parameter through `ClientContext`, all three `DatabricksRetryPolicy` construction sites (Thrift, SEA, UnifiedHttpClient), and the retry policy's `should_retry`/`is_retry` methods
- Default behavior (retry without requiring the header) is unchanged

## Test plan

- [x] 8 new unit tests covering server-directed-only mode in `tests/unit/test_retry.py`
- [x] Fixed mock fixture in `tests/unit/test_unified_http_client.py` for the new attribute
- [x] Full unit test suite passes (`poetry run python -m pytest tests/unit`)

### DIFF
--- a/src/databricks/sql/auth/common.py
+++ b/src/databricks/sql/auth/common.py
@@ -47,7 +47,7 @@ class ClientContext:
         retry_stop_after_attempts_duration: Optional[float] = None,
         retry_delay_default: Optional[float] = None,
         retry_dangerous_codes: Optional[List[int]] = None,
-        retry_server_directed_only: Optional[bool] = None,
+        respect_server_retry_after_header: Optional[bool] = None,
         proxy_auth_method: Optional[str] = None,
         pool_connections: Optional[int] = None,
         pool_maxsize: Optional[int] = None,
@@ -81,7 +81,7 @@ class ClientContext:
         )
         self.retry_delay_default = retry_delay_default or 5.0
         self.retry_dangerous_codes = retry_dangerous_codes or []
-        self.retry_server_directed_only = bool(retry_server_directed_only)
+        self.respect_server_retry_after_header = bool(respect_server_retry_after_header)
         self.proxy_auth_method = proxy_auth_method
         self.pool_connections = pool_connections or 10
         self.pool_maxsize = pool_maxsize or 20
--- a/src/databricks/sql/auth/common.py
+++ b/src/databricks/sql/auth/common.py
@@ -47,6 +47,7 @@ class ClientContext:
         retry_stop_after_attempts_duration: Optional[float] = None,
         retry_delay_default: Optional[float] = None,
         retry_dangerous_codes: Optional[List[int]] = None,
+        retry_server_directed_only: Optional[bool] = None,
         proxy_auth_method: Optional[str] = None,
         pool_connections: Optional[int] = None,
         pool_maxsize: Optional[int] = None,
@@ -80,6 +81,7 @@ class ClientContext:
         )
         self.retry_delay_default = retry_delay_default or 5.0
         self.retry_dangerous_codes = retry_dangerous_codes or []
+        self.retry_server_directed_only = bool(retry_server_directed_only)
         self.proxy_auth_method = proxy_auth_method
         self.pool_connections = pool_connections or 10
         self.pool_maxsize = pool_maxsize or 20
--- a/src/databricks/sql/auth/retry.py
+++ b/src/databricks/sql/auth/retry.py
@@ -94,6 +94,7 @@ class DatabricksRetryPolicy(Retry):
         stop_after_attempts_duration: float,
         delay_default: float,
         force_dangerous_codes: List[int],
+        server_directed_only: bool = False,
         urllib3_kwargs: dict = {},
     ):
         # These values do not change from one command to the next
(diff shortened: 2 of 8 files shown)


src/databricks/sql/auth/common.py
```diff
@@ -80,6 +81,7 @@ def __init__(
         )
         self.retry_delay_default = retry_delay_default or 5.0
         self.retry_dangerous_codes = retry_dangerous_codes or []
+        self.retry_server_directed_only = bool(retry_server_directed_only)
```
jprakash-db: Not needed to perform `bool (value),` we have already mentioned the types so users will get a hint on the type while using the connector

src/databricks/sql/auth/common.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         retry_stop_after_attempts_duration: Optional[float] = None,
         retry_delay_default: Optional[float] = None,
         retry_dangerous_codes: Optional[List[int]] = None,
+        retry_server_directed_only: Optional[bool] = None,
```
jprakash-db: Why optional ? Can be directly assigned a default value of False right?

src/databricks/sql/auth/retry.py
```diff
@@ -388,6 +393,12 @@ def should_retry(self, method: str, status_code: int) -> Tuple[bool, str]:
         if not self._is_method_retryable(method):
             return False, "Only POST requests are retried"
 
+        # In server_directed_only mode, only retry when the server explicitly signals
+        # it's safe via a Retry-After header. This prevents duplicate side effects for
+        # non-idempotent operations.
+        if self.server_directed_only and not has_retry_after:
```
jprakash-db: nit: can we give the variable a better name, like server_directed_only doesn't seem right. a better name could be `enforce_retry_after`, `respect_server_retry_after`

src/databricks/sql/backend/thrift_backend.py
```diff
@@ -191,6 +191,9 @@ def __init__(
                 " This behaviour is deprecated and will be removed in a future release."
             )
         self.force_dangerous_codes = kwargs.get("_retry_dangerous_codes", [])
+        self._retry_server_directed_only = kwargs.get(
+            "_retry_server_directed_only", False
```
jprakash-db: nit: a better name could be `enforce_server_retry_after`

tests/unit/test_retry.py
```diff
@@ -84,6 +84,114 @@ def test_excessive_retry_attempts_error(self, t_mock, retry_policy):
                 # Internally urllib3 calls the increment function generating a new instance for every retry
                 retry_policy = retry_policy.increment()
 
+    @pytest.fixture()
+    def server_directed_retry_policy(self) -> DatabricksRetryPolicy:
```
jprakash-db: There is already a fixture called `retry_policy` why can't we simply override it with kwargs for additional arguments ?

tests/unit/test_retry.py
```diff
@@ -84,6 +84,114 @@ def test_excessive_retry_attempts_error(self, t_mock, retry_policy):
...
+        server_directed_retry_policy.command_type = CommandType.OTHER
+        should_retry, msg = server_directed_retry_policy.should_retry(
+            "POST", 503, has_retry_after=False
+        )
+        assert should_retry is False
+        assert "server_directed_only" in msg
+
+    def test_server_directed_only__overrides_dangerous_codes(self):
+        """force_dangerous_codes=[500] + no Retry-After → no retry in server_directed_only mode"""
+        policy = DatabricksRetryPolicy(
```
jprakash-db: plz, use the common fixture with just kwargs override

src/databricks/sql/auth/retry.py
sd-db: `respect_server_retry_after_header` seems better (_unambiguous about our intentions_); I will do the change.

src/databricks/sql/backend/thrift_backend.py
sd-db: I have cleaned up this specific piece of code.

tests/unit/test_retry.py
sd-db: Totally agree, good catch !! Updated

tests/unit/test_retry.py
sd-db: Updated

src/databricks/sql/auth/common.py
sd-db: I think `None` is much more sensible default. Ideally for optional params default should always be `None`
- It denotes that the params is unset, vs something can be set to False (_not the case here specifically but can be for other params_)
- Secondly this is also the pattern throughout the codebase (which imo makes sense)

src/databricks/sql/auth/common.py
sd-db: Since default is `None` when a param is not set, this is required type conversion

jprakash-db (approved):

sd-db:
E2E/Code Coverage tests can't run against fork PRs. I have ran these locally to verify all fine.

PR #772, opened 2026-04-13
Optimize CI: consolidate workflows, fix caching, speed up e2e tests (47min → 15min)

vikrantpuppala:
## Summary
- **Consolidate 3 workflows into 1**: Delete `integration.yml` and `daily-telemetry-e2e.yml` — coverage workflow already runs all e2e tests. Add `push: main` trigger. Run all tests (including telemetry) in a single pytest invocation with `--dist=loadgroup` for xdist_group isolation.
- **Fix pyarrow cache**: Remove `cache-path: .venv-pyarrow` — poetry always creates `.venv`, so the cache was never saved ("Path does not exist" error). 3.14 PyArrow jobs dropped from **18min → 3min** once cache populated.
- **Fix 3.14 post-test DNS hang**: Add `enable_telemetry=False` to unit test dummy connection args. Unit tests using `server_hostname="foo"` triggered real HTTP calls — on protected runners this caused an 8-min process hang. 3.14 unit tests dropped from **17min → 2min**.
- **Better xdist distribution**: Split `TestPySQLLargeQueriesSuite` into 3 separate classes and split lz4 on/off into separate parametrized cases so xdist distributes slow tests across 4 workers.
- **Use 4 workers**: `-n 4` instead of `-n auto` (2 CPUs). E2e tests are network-bound (waiting on warehouse), not CPU-bound.
- **Reduce test sizes**: Large result set tests 300MB → 100MB. `test_long_running_query` threshold 3min → 1min, starting scale_factor 1 → 50.

## Results (measured)
| Metric | Before | After |
|---|---|---|
| E2E workflows per PR | 3 | 1 |
| Coverage wall-clock | **47 min** | **15 min** |
| Integration workflow | 40 min | deleted |
| 3.14 unit tests | 17m38s | 2m46s |
| 3.14 PyArrow tests | 18m26s | 3m21s |
| 3.14 linting | 15m46s | 1m27s |
| Total warehouse compute per PR | ~85 min | ~15 min |

## Test plan
- [x] All 34 CI checks pass
- [x] Coverage workflow runs all tests including telemetry (870 passed, 25 skipped)
- [x] 3.14 pyarrow cache saves and hits on subsequent runs
- [x] 3.14 jobs no longer have post-test DNS hang
- [x] LargeQueriesSuite tests distributed across multiple xdist workers

SKIP_COVERAGE_CHECK = CI workflow changes only, no source code coverage impact

This pull request was AI-assisted by Isaac.


.github/workflows/code-coverage.yml
```diff
@@ -32,25 +37,16 @@ jobs:
...
           poetry run pytest tests/unit tests/e2e \
-            -m "not serial" \
-            -n auto \
+            -n 4 \
+            --dist=loadgroup \
             --cov=src \
             --cov-report=xml \
             --cov-report=term \
             -v
-      - name: Run telemetry tests with coverage (isolated)
```
vikrantpuppala, line 45: This is consolidated into the single pytest invocation above that runs all tests together

.github/workflows/daily-telemetry-e2e.yml
vikrantpuppala, line 1: Now that coverage runs all tests including telemetry on every PR and push to main, the weekly schedule is redundant for post-merge coverage.

.github/workflows/integration.yml
vikrantpuppala, line 1: run as part of coverage

tests/e2e/common/large_queries_mixin.py
```diff
@@ -138,3 +148,8 @@ def test_long_running_query(self, extra_params):
                 print("Took {} s with scale factor={}".format(duration, scale_factor))
                 # Extrapolate linearly to reach 3 min and add 50% padding to push over the limit
                 scale_factor = math.ceil(1.5 * scale_factor / current_fraction)
+
+
+# Keep backward-compatible alias that combines all three
```
jprakash-db: What do you mean by backward compatible ? These are just tests there is no meaning for backward compatible

tests/e2e/test_driver.py
```diff
@@ -138,24 +142,36 @@ def assertEqualRowValues(self, actual, expected):
...
             return [row]
         else:
             return None
 
+
+class TestPySQLLargeWideResultSet(PySQLPytestTestCase, _LargeQueryRowHelper, LargeWideResultSetMixin):
+    pass
+
+
+class TestPySQLLargeNarrowResultSet(PySQLPytestTestCase, _LargeQueryRowHelper, LargeNarrowResultSetMixin):
```
jprakash-db: This doesn't make sense you are inheriting and mixing for no real point. You can just create a TestPySQLLargeNarrowResultSet and just define the function. A mixin is supposed to be something that is like a helper or sort of something to plug and play randomly. Here its a complete test

tests/e2e/common/large_queries_mixin.py
```diff
@@ -138,3 +148,8 @@ def test_long_running_query(self, extra_params):
...
+class LargeQueriesMixin(LargeWideResultSetMixin, LargeNarrowResultSetMixin, LongRunningQueryMixin):
```
jprakash-db: Also this LargeQueriesMixin doesn't seem to be used anywhere

tests/e2e/test_driver.py
```diff
@@ -138,24 +142,36 @@ def assertEqualRowValues(self, actual, expected):
                 assert act[i] == exp[i]
 
 
-class TestPySQLLargeQueriesSuite(PySQLPytestTestCase, LargeQueriesMixin):
+class _LargeQueryRowHelper:
```
jprakash-db: Why is this a _ , its a class name. Just follow simple camle case

jprakash-db (approved):
Overall LGTM. Thanks for making the changes

PR #775, opened 2026-04-21
[PECOBLR-2461] Add comprehensive MST transaction E2E tests

vikrantpuppala:
## Summary
- Rewrites `tests/e2e/test_transactions.py` with 42 MST tests across 5 categories
- Tests are parallelization-friendly (each test uses its own unique Delta table)
- Full suite runs in ~2 minutes on 4 workers (`pytest -n 4 --dist=loadgroup`)
- Covers gaps identified in the [MST + xDBC Metadata RPCs audit](https://docs.google.com/document/d/1WSX5imeH8lwiWB6hrKJbL-D6iROxZ7EPWhdxzu-ObTo/edit)

## Test categories
- **TestMstCorrectness (18)** — commit/rollback/isolation/multi-table atomicity/repeatable reads/write conflict/parameterized DML/auto-start after commit/rollback/close-connection-implicit-rollback
- **TestMstApi (6)** — DB-API-specific: autocommit getter/setter, isolation level (supported + unsupported), commit without active txn, setAutoCommit during active txn
- **TestMstMetadata (6)** — `cursor.columns/tables/schemas/catalogs` inside a txn + two freshness tests asserting Thrift metadata RPCs are non-transactional (they see concurrent DDL that the txn should not see)
- **TestMstBlockedSql (9)** — MSTCheckRule enforcement. The server allowlists only `SELECT / INSERT / MERGE / UPDATE / DELETE / DESCRIBE TABLE` inside MST. 8 statements throw + abort txn (SHOW COLUMNS/TABLES/SCHEMAS/CATALOGS/FUNCTIONS, DESCRIBE QUERY, DESCRIBE TABLE EXTENDED, `information_schema`). 1 statement succeeds (DESCRIBE TABLE basic — explicitly allowed by server).
- **TestMstExecuteVariants (2)** — executemany commit + rollback

## Parallelisation strategy
- Each test derives a unique Delta table name from its node id so 4 parallel workers don't collide.
- Tests that spawn concurrent connections to the same table (repeatable reads, write conflict, freshness) use `xdist_group` so the concurrent connections within a single test don't conflict with other tests on different workers.

## Test plan
- [x] Run full suite locally against dogfood warehouse (older DBR) — 42/42 pass
- [x] Verify CI's unit test suite passes with the rebased dependency_manager fix (#776)
- [ ] Verify e2e CI (test-with-coverage) passes

## Related
- JDBC client-side equivalent: databricks/databricks-jdbc#1366
[...]

PR #788, opened 2026-05-15
Test fixes for MST metadata expectations and User-Agent env leak

vikrantpuppala:
## Summary

Two independent test-only fixes bundled together because both were uncovered while investigating CI red on `test-with-coverage`.

### 1. Flip `TestMstMetadata` to assert blocked behavior

The four `tests/e2e/test_transactions.py::TestMstMetadata::test_cursor_*_in_mst` tests, added in #775, were written speculatively expecting Thrift `Get{Catalogs,Schemas,Tables,Columns}` RPCs to succeed inside an active multi-statement transaction. The server's MST guard is intentionally designed to reject these (`"... is not supported within a multi-statement transaction."`), matching the small allowlist pattern documented in `TestMstBlockedSql`.

- Renamed the four `*_in_mst` tests to `*_blocked` and rewrote them to assert the expected `DatabaseError` via a shared `_assert_metadata_rpc_blocked` helper.
- Deleted the two `*_non_transactional_after_concurrent_*` freshness tests outright — their premise ("Thrift RPCs bypass MST and see concurrent DDL") is no longer valid, so there's nothing meaningful left to assert.
- Updated the `TestMstMetadata` and `TestMstBlockedSql` docstrings to reflect actual server behavior.

One subtle difference vs `TestMstBlockedSql`'s SQL-form blocks: the Get* rejection happens at the MST guard before reaching the transaction, so the active txn remains usable after a blocked Get* (whereas SQL-form blocks abort the txn). The shared helper documents this.

### 2. Fix `test_useragent_header` environment-sensitivity

`tests/unit/test_session.py::test_useragent_header` strictly asserts `("User-Agent", "<name>/<version>") in http_headers`. When run inside any AI-agent shell that `databricks.sql.common.agent.KNOWN_AGENTS` knows about (e.g. Claude Code sets `CLAUDECODE=1`), `session.py:68-70` appends ` agent/<product>` to the header and the `in` check fails. CI runners don't set agent env vars, so this passed there — it's an environment-sensitivity bug rather than a code regression.

The fix mirrors the pattern in `tests/unit/test_agent_detection.py:42-49`: `monkeypatch.delenv` every known agent env var before the assertion. Iterating `KNOWN_AGENTS` keeps the cleared list automatically in sync if new agents are added to `src/databricks/sql/common/agent.py`.

## Test plan

- [x] `pytest tests/e2e/test_transactions.py::TestMstMetadata` against dogfood: 4 passed in 33.67s
- [x] `pytest tests/unit/test_session.py::TestSession::test_useragent_header` passes inside a Claude Code shell (was failing before the fix)
- [x] Full local unit suite: 601 passed, 4 skipped
- [ ] CI `test-with-coverage` (e2e + coverage) — verifying via this PR

### DIFF
--- a/tests/e2e/test_transactions.py
+++ b/tests/e2e/test_transactions.py
@@ -29,6 +29,7 @@ import uuid
 import pytest
 
 import databricks.sql as sql
+from databricks.sql.exc import DatabaseError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


msrathore-db (approved):
LGTM

PR #793, opened 2026-05-19
feat(backend/kernel): match Thrift backend's user-visible surface

vikrantpuppala:
## Summary

Three small parity fixes that close **34 of the 66 diffs** the Thrift vs Kernel comparator surfaced against Dogfood today (see [`databricks-driver-test#303`](https://github.com/databricks/databricks-driver-test/pull/303) for the full diff catalog). None of these are data-correctness changes — they're contract-level differences that affect code reading the cursor's description, catching server errors by class, or calling `get_columns` with no catalog.

### 1. PEP-249 `description.null_ok` is always `None` (closes 29 diffs)

`description_from_arrow_schema` previously took `null_ok` from the Arrow `field.nullable` bit. The Thrift backend has always reported `None` for this slot (PEP 249 permits either). Hardcode `None` so the kernel backend is a drop-in for code that reads `cursor.description[i][6]`.

### 2. `SqlError` maps to `ServerOperationError` (closes 3 diffs)

Server-side SQL failures (syntax error, missing object, etc.) used to wrap into the generic `DatabaseError` on the kernel backend. The Thrift backend raises `ServerOperationError` for the same shape. `ServerOperationError` is a `DatabaseError` subclass, so existing catches of the base class are unaffected — but code that catches the specific subclass now works equivalently.

### 3. `get_columns(catalog_name=None)` accepted (closes 2 diffs)

Previously rejected at the connector layer with a `ProgrammingError`. The kernel's `list_columns` now issues `SHOW COLUMNS IN ALL CATALOGS` server-side when catalog is `None` (per [`databricks-sql-kernel#33`](https://github.com/databricks/databricks-sql-kernel/pull/33)); the response carries `catalogName` per row so `TABLE_CAT` is correctly attributed without client-side enumeration. Matches Thrift's `getColumns(null, …)` behaviour.

## Out of scope (separate PRs)

- **INTERVAL columns crashing on fetch** (11 diffs) — needs a kernel-side post-decode cast since `pyarrow.compute.cast(month_interval → utf8)` is not implemented. Separate kernel PR.
- **`_use_arrow_native_complex_types=False` not honoured** (2 diffs) — pre-existing gap shared with the native SEA backend; needs separate work that fixes both.

## Test plan

- [x] `python -m pytest tests/unit/` — 715 passed, 4 skipped (matches prior baseline; updated kernel-backend tests to reflect the new contract).
- [x] `black src/databricks/sql/backend/kernel/ tests/unit/test_kernel_*.py` clean.
- [ ] Re-running the comparator end-to-end after merge to confirm the 34-diff drop.

This pull request and its description were written by Claude Code.

### DIFF
--- a/src/databricks/sql/backend/kernel/client.py
+++ b/src/databricks/sql/backend/kernel/client.py
@@ -517,14 +517,13 @@ class KernelDatabricksClient(DatabricksClient):
     ) -> "ResultSet":
         if self._kernel_session is None:
             raise InterfaceError("get_columns requires an open session.")
-        if not catalog_name:
-            # Kernel's list_columns requires a catalog (SEA `SHOW
-            # COLUMNS` cannot span catalogs). Surface the constraint
-            # explicitly rather than letting the kernel error.
-            raise ProgrammingError(
-                "get_columns requires catalog_name on the kernel backend."
-            )
         try:
+            # `catalog_name=None` is supported: the kernel issues
+            # `SHOW COLUMNS IN ALL CATALOGS` server-side and the
+            # response carries `catalogName` per row, so each result
+            # row's `TABLE_CAT` is correctly attributed. Matches the
+            # Thrift backend's `getColumns(null, …)` behaviour from
+            # the user's perspective.
             stream = self._kernel_session.metadata().list_columns(
                 catalog=catalog_name,
                 schema_pattern=schema_name,
--- a/src/databricks/sql/backend/kernel/type_mapping.py
+++ b/src/databricks/sql/backend/kernel/type_mapping.py
@@ -85,9 +85,11 @@ def description_from_arrow_schema(schema: pyarrow.Schema) -> List[Tuple]:
     """Build a PEP 249 ``description`` list from a pyarrow Schema.
 
     Each tuple is ``(name, type_code, display_size, internal_size,
-    precision, scale, null_ok)``. ``null_ok`` is taken from
-    ``field.nullable``; the other four are not reported by the
-    kernel today.
+    precision, scale, null_ok)``. PEP 249 allows ``null_ok`` to be
+    either a bool or ``None``; the Thrift backend always reports
+    ``None``, so we match that here for drop-in parity. The actual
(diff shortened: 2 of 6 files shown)


src/databricks/sql/backend/kernel/client.py
```diff
@@ -517,14 +517,13 @@ def get_columns(
     ) -> "ResultSet":
         if self._kernel_session is None:
             raise InterfaceError("get_columns requires an open session.")
-        if not catalog_name:
-            # Kernel's list_columns requires a catalog (SEA `SHOW
-            # COLUMNS` cannot span catalogs). Surface the constraint
```
msrathore-db, line 522: This command is not yet supported in all supported DBRs. Have we verified this?

src/databricks/sql/backend/kernel/type_mapping.py
```diff
@@ -85,24 +85,56 @@ def description_from_arrow_schema(schema: pyarrow.Schema) -> List[Tuple]:
     """Build a PEP 249 ``description`` list from a pyarrow Schema.
 
     Each tuple is ``(name, type_code, display_size, internal_size,
-    precision, scale, null_ok)``. ``null_ok`` is taken from
-    ``field.nullable``; the other four are not reported by the
-    kernel today.
+    precision, scale, null_ok)``. PEP 249 allows ``null_ok`` to be
```
msrathore-db, line 88: Nullability is correctly sent by the SEA driver. Why are we changing it to None? This field was also fixed for one of the recent DBR versions

src/databricks/sql/backend/kernel/client.py
vikrantpuppala: this is SEA so DBSQL check is only required

src/databricks/sql/backend/kernel/type_mapping.py
vikrantpuppala: Thrift parity

msrathore-db (commented):
LGTM other than the 2 comments

msrathore-db (approved):

PR #794, opened 2026-05-20
fix: close unfetched async commands

### LINKED ISSUE #791: Cursor.close() leaks server-side handle for async commands that were never fetched
## Summary

`Cursor.close()` does not free the server-side statement handle for async-submitted commands when `get_execution_result()` was never called. The handle leaks until the session closes (where backend-specific sweeps may or may not catch it).

## Repro

```python
import databricks.sql

with databricks.sql.connect(...) as conn:
    cur = conn.cursor()
    cur.execute("SELECT * FROM huge_table", async_op=True)
    # User decides not to wait for the result.
    cur.close()
    # Server-side ExecutedAsyncStatement still tracked until conn.close().
```

## Root cause

`Cursor.close()` at `src/databricks/sql/client.py:1713-1718` only nulls `active_command_id` and closes `active_result_set` — but for the async-submit-without-fetch case, `active_result_set` is `None`, so `backend.close_command(active_command_id)` is never invoked.

```python
def close(self) -> None:
    """Close cursor"""
    self.open = False
    self.active_command_id = None
    if self.active_result_set:
        self._close_and_clear_active_result_set()
```

## Scope

Affects all three backends: Thrift, SEA, and the new kernel backend. Each one tracks the async handle differently (Thrift via Thrift-server-side state; SEA via the SEA statement; kernel via `_async_handles` dict), but the shared `Cursor.close()` path doesn't notify any of them about the unfetched async handle.

## Suggested fix

Single change in shared `Cursor.close()`:

```python
def close(self) -> None:
    self.open = False
    if self.active_result_set:
        self._close_and_clear_active_result_set()
    elif self.active_command_id is not None:
        # async submission that never had get_execution_result called —
        # the result-set close path didn't fire, so issue an explicit
        # close_command to free the server-side handle.
        try:
            self.backend.close_command(self.active_command_id)
        except Exception as exc:
            logger.warning("close_command on cursor close failed: %s", exc)
    self.active_command_id = None
```

Safe across all backends: Thrift's `close_command` sends `TCloseOperation` (valid call), SEA's sends a cancel/close, kernel's is already tolerant of unknown ids (no-op).

## Surfaced by

Review of PR #787 (kernel backend). Documented as P0 #1 in https://github.com/databricks/databricks-sql-python/pull/787#issuecomment-4475165482 — but the leak is pre-existing on Thrift and SEA too, not specific to the kernel backend.

nanookclaw:
## Summary

`Cursor.close()` now frees an active async command id when no result set has been fetched yet. This covers the `execute_async()` followed by `close()` path from #791, where `active_command_id` is set but `active_result_set` is still `None`.

When a result set exists, close still delegates to `active_result_set.close()` so that path can perform its existing backend cleanup without a duplicate `close_command` call. The cursor clears `active_command_id` in a `finally` block either way.

## Tests

- `PYTHONPATH=src uv run --with pytest --with thrift --with pandas --with lz4 --with requests --with oauthlib --with openpyxl --with urllib3 --with python-dateutil --with pyjwt --with pybreaker --python 3.12 --env-file test.env.example python -m pytest tests/unit/test_client.py -k 'close_closes_active_command_without_result_set or close_delegates_to_active_result_set_without_double_closing_command or closing_result_set_hard_closes_commands or closing_connection_closes_commands'`
- `uv run --with black==22.3.0 --python 3.12 black --check src/databricks/sql/client.py tests/unit/test_client.py`
- `git diff --check`

Closes #791

### DIFF
--- a/src/databricks/sql/client.py
+++ b/src/databricks/sql/client.py
@@ -1713,9 +1713,13 @@ class Cursor:
     def close(self) -> None:
         """Close cursor"""
         self.open = False
-        self.active_command_id = None
-        if self.active_result_set:
-            self._close_and_clear_active_result_set()
+        try:
+            if self.active_result_set:
+                self._close_and_clear_active_result_set()
+            elif self.active_command_id is not None:
+                self.backend.close_command(self.active_command_id)
+        finally:
+            self.active_command_id = None
 
     @property
     def query_id(self) -> Optional[str]:
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -280,6 +280,33 @@ class ClientTestSuite(unittest.TestCase):
             cursor.close = mock_close
         mock_close.assert_called_once_with()
 
+    def test_close_closes_active_command_without_result_set(self):
+        mock_backend = Mock()
+        mock_command_id = Mock(spec=CommandId)
+        cursor = client.Cursor(Mock(), mock_backend)
+        cursor.active_command_id = mock_command_id
+
+        cursor.close()
+
+        mock_backend.close_command.assert_called_once_with(mock_command_id)
+        self.assertIsNone(cursor.active_command_id)
+        self.assertIsNone(cursor.active_result_set)
+
+    def test_close_delegates_to_active_result_set_without_double_closing_command(self):
+        mock_backend = Mock()
+        mock_result_set = Mock()
+        mock_command_id = Mock(spec=CommandId)
+        cursor = client.Cursor(Mock(), mock_backend)
+        cursor.active_result_set = mock_result_set
+        cursor.active_command_id = mock_command_id
+
+        cursor.close()
+
+        mock_result_set.close.assert_called_once_with()
+        mock_backend.close_command.assert_not_called()
+        self.assertIsNone(cursor.active_command_id)
+        self.assertIsNone(cursor.active_result_set)
+
     def dict_product(self, dicts):
         """
         Generate cartesion product of values in input dictionary, outputting a dictionary


src/databricks/sql/client.py
```diff
@@ -1713,9 +1713,13 @@ def cancel(self) -> None:
...
         """Close cursor"""
         self.open = False
-        self.active_command_id = None
-        if self.active_result_set:
-            self._close_and_clear_active_result_set()
+        try:
+            if self.active_result_set:
+                self._close_and_clear_active_result_set()
+            elif self.active_command_id is not None:
+                self.backend.close_command(self.active_command_id)
```
vikrantpuppala, line 1720: - ResultSet.close() (result_set.py:184-187) wraps the same call in try/except RequestError, swallowing CursorAlreadyClosedError. The new branch has neither the guard nor the except — only a finally that clears active_command_id but does not contain the raise.
- Connection._close() (client.py:547-550) runs for cursor in self._cursors: cursor.close() with the try/except covering only the subsequent session.close(). A raise from any cursor aborts the loop → remaining cursors never closed, session.close() never runs.
- Blast radius: conn.close() / with teardown when an async cursor's handle is already gone server-side (post-cancel, expired session) or a transient network blip → whole-session leak + raw stack trace out of close(). This is the exact scenario the PR targets, so it's a live regression, not hypothetical.
- Fix: mirror ResultSet.close() — swallow RequestError/CursorAlreadyClosedError (+ a catch-all) and log. This is what issue #791's suggested fix specified and the PR dropped.

vikrantpuppala: ResultSet.close() gates the RPC on status != CLOSED and not has_been_closed_server_side and connection.open (result_set.py:180-182). The new branch fires unconditionally, sending a pointless (and, per F1, unguarded) close_command RPC even when the session is already closed server-side. Gating on self.connection.open would avoid the round-trip and much of F1's trigger surface.

tests/unit/test_client.py
vikrantpuppala, line 1: Both added tests use bare Mock() backends whose close_command never raises, so they only prove the happy case. The one genuinely new invariant — active_command_id still cleared (via finally) and close() staying best-effort when close_command raises — is entirely unverified. Fix: add a test with mock_backend.close_command.side_effect = RequestError(...), asserting close() does not raise and active_command_id is None.

vikrantpuppala (commented):
Thanks for the contribution, would be great if we can address the comments and get this in

PR #329, opened 2024-01-23
[sqlalchemy] Add table and column comment support

susodapop:
## Description

This PR consolidates the changes from #306 and #308 after review + tests + refactor which I placed into a staging branch to coordinate reviews and testing.

PR #799, opened 2026-05-22
ci: add label-gated proxy-test dispatch to databricks-driver-test

vikrantpuppala:
## Summary

Wires this repo into the proxy-based integration test infrastructure in [databricks/databricks-driver-test](https://github.com/databricks/databricks-driver-test). Mirrors the pattern used by [databricks-odbc#55](https://github.com/databricks/databricks-odbc/pull/55) and the ADBC drivers (C# / Rust).

The driver-test repo already runs proxy-based integration tests against the **canonical C# implementation** on every PR to it. With this change, the same harness can be triggered against this repo's PRs:

1. A maintainer reviews a PR and adds the `integration-test` label.
2. This workflow dispatches a `python-pr-test` event to `databricks/databricks-driver-test` with the PR number + head SHA.
3. The driver-test repo installs the connector at the dispatched SHA via `pip install git+https://...@<sha>`, runs the Python proxy test suite ([databricks-driver-test#329](https://github.com/databricks/databricks-driver-test/pull/329), `tests/python/`, ~59 tests in Phase 1) against the pecotesting warehouse with mitmproxy in the loop, and posts a check run back to this PR.

## Security

| Concern | Mitigation |
|---|---|
| Untrusted PR code running with workspace credentials | Label-gated: only maintainers can apply `integration-test`. |
| Re-running after a malicious follow-up commit | Auto-removed on `synchronize`: pushing new commits clears the label, forcing maintainer re-review before tests re-run. A comment is posted explaining what happened. |
| Bypassing the gate via the merge queue | `merge_group` is trusted at that point — dispatches automatically. |

The dispatch token comes from a GitHub App (`INTEGRATION_TEST_APP_ID` / `INTEGRATION_TEST_PRIVATE_KEY`), not a personal PAT, so it can be scoped to just `databricks-driver-test` and rotated independently.

## Companion PRs

- [databricks-driver-test#322](https://github.com/databricks/databricks-driver-test/pull/322) — design doc (merged).
- [databricks-driver-test#329](https://github.com/databricks/databricks-driver-test/pull/329) — `tests/python/` scaffolding + Phase 1 tests + receiving CI workflow. **Must merge before this PR is useful** — without it, the dispatch goes nowhere.

## Test plan

- [ ] After merge: create the `integration-test` label in this repo (one-time manual step — same as we did for `databricks-odbc`).
- [ ] Add the label to an open PR and verify the `python-pr-test` dispatch appears in the [driver-test Actions tab](https://github.com/databricks/databricks-driver-test/actions/workflows/python-proxy-tests.yml).
- [ ] Verify the check run shows up on this PR (in-progress → success / failure).
- [ ] Push a new commit to the test PR and verify the label is auto-removed with a comment explaining why.

## Out of scope

- The companion driver-test PR ([#329](https://github.com/databricks/databricks-driver-test/pull/329)) is responsible for the actual test suite. This PR only adds the dispatch trigger.
- Phase 1 ports 2 of 17 C# spec files (~59 of ~174 tests). Subsequent phases land in follow-up PRs on the driver-test repo without changes here.

This pull request and its description were AI-assisted by Isaac.

### DIFF
--- a/.github/workflows/trigger-integration-tests.yml
+++ b/.github/workflows/trigger-integration-tests.yml
@@ -1,0 +1,409 @@
+name: Trigger Integration Tests
+
+# Dispatches the proxy-based integration test suite in
+# databricks/databricks-driver-test to run against this PR's commit.
+#
+# Mirrors the canonical pattern in adbc-drivers/databricks. The model:
+#
+#   - On a normal PR event (open / push / reopen / non-IT label) we
+#     post a `success` Python Proxy Tests check immediately so the
+#     required check doesn't block the PR. The real tests are gated
+#     in the merge queue.
+#   - When a maintainer adds the `integration-test` label we dispatch
+#     the suite as a preview — useful for catching regressions before
+#     merge queue time.
+#   - Pushing new commits to the PR auto-removes the label so a
+#     subsequent labelled run requires a fresh maintainer review.
+#   - On the `merge_group` event the suite runs as the real required
+#     gate. Only PRs whose tests dispatch (or auto-pass when no driver
+#     files changed) can proceed to `main`.
+#
+# Required external setup (outside this workflow):
+#
+#   1. `integration-test` label exists in this repo (one-off; created
+#      separately).
+#   2. `INTEGRATION_TEST_APP_ID` / `INTEGRATION_TEST_PRIVATE_KEY` repo
+#      secrets installed for the dispatcher GitHub App (write access
+#      to databricks/databricks-driver-test).
+#   3. Merge queue enabled on `main` branch protection AND
+#      `Python Proxy Tests` listed as a required status check. Without
+#      this the merge-queue job is dead code and ITs run only on
+#      explicit label.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+  merge_group:  # Trigger when added to merge queue
+
+jobs:
+  # =============================================================================
+  # Security: Auto-remove label when new commits are pushed
+  # =============================================================================
+  remove-label-on-new-commit:
+    if: github.event_name == 'pull_request' && github.event.action == 'synchronize'
+    runs-on:
+      group: databricks-protected-runner-group
+      labels: linux-ubuntu-latest
+    permissions:
+      pull-requests: write
+      issues: write
+    steps:
+      - name: Check if integration-test label exists
+        id: check-label
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          script: |
+            const labels = context.payload.pull_request.labels.map(l => l.name);
+            const hasLabel = labels.includes('integration-test');
+            console.log(`integration-test label exists: ${hasLabel}`);
+            return hasLabel;
+
+      - name: Remove integration-test label
+        if: steps.check-label.outputs.result == 'true'
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          script: |
+            try {
+              await github.rest.issues.removeLabel({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                name: 'integration-test'
+              });
+              console.log('Removed integration-test label');
+            } catch (error) {
+              if (error.status === 404) {
+                console.log('Label already removed or does not exist');
+              } else {
+                throw error;
+              }
+            }
+
+      - name: Comment on PR about label removal
+        if: steps.check-label.outputs.result == 'true'
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          script: |
+            const pr = context.payload.pull_request;
+            const isFromFork = pr.head.repo.full_name !== pr.base.repo.full_name;
+            const repoType = isFromFork ? '**fork PR**' : 'PR';
+
+            const body = [
+              'Integration test approval reset.',
+              '',
+              `New commits were pushed to this ${repoType}. The \`integration-test\` label has been automatically removed for security.`,
+              '',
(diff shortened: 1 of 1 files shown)


msrathore-db (approved):
LGTM. Just sanitise the title

.github/workflows/trigger-integration-tests.yml
```diff
@@ -0,0 +1,157 @@
...
+          token: ${{ steps.app-token.outputs.token }}
+          repository: databricks/databricks-driver-test
+          event-type: python-pr-test
+          client-payload: |
+            {
+              "pr_number": "${{ github.event.pull_request.number }}",
+              "commit_sha": "${{ github.event.pull_request.head.sha }}",
+              "pr_repo": "${{ github.repository }}",
+              "pr_url": "${{ github.event.pull_request.html_url }}",
+              "pr_title": "${{ github.event.pull_request.title }}",
```
msrathore-db: **JSON injection risk in `pr_title`.** `${{ github.event.pull_request.title }}` is interpolated raw into the `client-payload` JSON. A PR title containing `"`, `\`, or a newline will break the JSON (causing the dispatch to fail), and a crafted title (e.g. `foo", "extra_field": "bar`) can append fields to the dispatched payload.

Recommend adding a sanitize step before this one and using its output:

```yaml
- name: Sanitize PR title
  id: sanitize
  uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
  with:
    result-encoding: string
    script: |
      const title = context.payload.pull_request.title || '';
      return title.replace(/[\\"\n\r\t]/g, ' ').substring(0, 200);
```

Then reference it as `"pr_title": "${{ steps.sanitize.outputs.result }}"`.

.github/workflows/trigger-integration-tests.yml
```diff
@@ -0,0 +1,157 @@
...
+        uses: peter-evans/repository-dispatch@ff45666b9427631e3450c54a1bcbee4d9ff4d7c0 # v3.0.0
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+          repository: databricks/databricks-driver-test
+          event-type: python-pr-test
+          client-payload: |
+            {
+              "commit_sha": "${{ github.event.merge_group.head_sha }}",
+              "pr_repo": "${{ github.repository }}"
+            }
```
msrathore-db, line 389: **Sparse merge-queue payload hurts traceability.** This payload only sends `commit_sha` and `pr_repo`, while the labeled-PR dispatch above sends `pr_number`, `pr_url`, `pr_title`, and `pr_author`. As a result, runs that come through the merge queue can't be traced back to the originating PR in the driver-test repo's reporting / check-run titles.

The PR number is recoverable from `github.event.merge_group.head_ref`, which is shaped like `refs/heads/gh-readonly-queue/<base>/pr-<N>-<sha>`. Suggest extracting it and filling in the same fields:

```yaml
- name: Extract PR number
  id: extract-pr
  run: |
    REF="${{ github.event.merge_group.head_ref }}"
    if [[ $REF =~ pr-([0-9]+) ]]; then
      echo "pr_number=${BASH_REMATCH[1]}" >> $GITHUB_OUTPUT
    else
      echo "Error: failed to extract PR number from merge group ref: '$REF'" >&2
      exit 1
    fi
```

Then pass `pr_number`, a constructed `pr_url`, and placeholder `pr_title`/`pr_author` (e.g. `"Merge queue validation"` / `"merge-queue"`) in the payload to match the labeled-PR shape.

vikrantpuppala:
Thanks for the review @msrathore-db — both comments uncovered a deeper issue.

Investigating your point about the merge-queue payload prompted me to look at the canonical pattern in [adbc-drivers/databricks](https://github.com/adbc-drivers/databricks/blob/main/.github/workflows/trigger-integration-tests.yml) (the established gold standard for cross-repo IT dispatch in this org). It turns out the original version of this PR — which I'd copied from `databricks-odbc#55` — was significantly simpler than the canonical pattern, and the merge-queue traceability gap was a symptom of a bigger missing piece: **the merge-queue job is supposed to be the *real* gate**, not a nice-to-have.

I've rewritten the workflow from scratch to match the canonical 4-job pattern (the ADBC version has 5 — extra one is for a second driver in the same repo).

### What's new (vs the original 2-job version)

- **`skip-integration-tests-pr`** — On any PR event that *isn't* a labeled event, post a synthetic `success` `Python Proxy Tests` check. Without this, making `Python Proxy Tests` a required check would block every PR until someone applied the label.
- **`merge-queue-python`** — Real gate on `merge_group`. Detects whether driver source files (or workflow / pyproject / lockfile) actually changed; if not, auto-pass; if yes, dispatch with the full traceability payload extracted from the queue ref. Posts a `failure` check if dispatch itself errors.
- **`trigger-tests-pr`** — Same dispatch path as before, but now also has changed-paths detection (auto-pass when no driver files touched), title sanitization (your F1), and a dispatch-failure handler.

### Both your comments addressed

- **F1 — JSON injection in `pr_title`**: sanitized via `actions/github-script` with `result-encoding: string`, applied in both PR-label and merge-queue dispatch paths.
- **F2 — Sparse merge-queue payload**: PR number is now extracted from the merge-queue head ref (`pr-<N>-<sha>` pattern); full payload (`pr_url`, placeholder `pr_title`/`pr_author`) matches the labeled-PR shape so the driver-test repo's check-run titles don't have to special-case merge-queue runs.

### Operational follow-ups (outside this PR)

For this to deliver value, after merge someone needs to:

1. Enable merge queue on `main` branch protection.
2. Mark `Python Proxy Tests` as a required status check.
3. Install `INTEGRATION_TEST_APP_ID` / `INTEGRATION_TEST_PRIVATE_KEY` secrets (the GitHub App with write access to `databricks-driver-test`).

The `integration-test` label has already been created.

Until step 1+2 are done, the merge-queue job is dead code and ITs run only on explicit label — same effective behaviour as the original version. The label path is fully usable in that interim state.

Re-pushed via `--force-with-lease`. PR should be ready for another look.

PR #800, opened 2026-05-22
test: smoke-test proxy IT dispatch workflow (DO NOT MERGE)

vikrantpuppala:
## Purpose

End-to-end smoke test of the IT dispatch workflow added in #799. **Not for merge** — closing after the smoke test confirms the workflow behaves correctly.

## Diff

A 5-line docstring addition in \`src/databricks/sql/backend/kernel/__init__.py\` — touches \`src/\` so the changed-paths filter in \`trigger-tests-pr\` fires.

### Phase 1 — PR opened (this commit)
- [x] \`skip-integration-tests-pr\` runs and posts a synthetic \`success\` \`Python Proxy Tests\` check.
- [x] \`trigger-tests-pr\` and \`merge-queue-python\` skip (no label, no merge_group).
- [ ] Verify \`Python Proxy Tests\` shows up as **green** in the check rollup.

### Phase 2 — apply \`integration-test\` label
- [ ] \`trigger-tests-pr\` runs.
  - \`Detect changed driver paths\` → \`python: true\` (we touched \`src/\`).
  - \`Sanitize PR title\` → strips quote/newline/etc. from the title.
  - \`Dispatch Python tests to internal repo\` → either succeeds (if secrets are installed on the App + this repo) or fails with a 403/401 (if secrets aren't installed yet → expected, will be addressed separately).
- [ ] If dispatch succeeds: comment posted on PR; driver-test repo's \`python-proxy-tests.yml\` workflow runs against this branch's HEAD.
- [ ] If dispatch fails: \`Fail check on dispatch error\` step posts a \`failure\` check on \`Python Proxy Tests\` so we know the gate didn't silently pass.

### Phase 3 — push a new commit
- [ ] \`remove-label-on-new-commit\` runs.
  - Removes the \`integration-test\` label.
  - Posts the "Integration test approval reset" comment.

### Phase 4 — cleanup
- Close PR. No merge.

## Out of scope

This PR doesn't verify the **merge queue** path (\`merge-queue-python\` job). That requires merge queue to be enabled on \`main\` branch protection (operational step, separate from this PR).

This pull request and its description were AI-assisted by Isaac.

PR #801, opened 2026-05-22
ci(integration-tests): use github.token for check-run posters

vikrantpuppala:
## Summary

Follow-up to #799 (cross-repo IT dispatch). The dispatch failure handlers and auto-pass steps in \`trigger-integration-tests.yml\` were posting check-runs with \`steps.public-token.outputs.token\` — itself an App-token-generating step.

That created a **silent-failure trap**: if the \`INTEGRATION_TEST_APP_ID\` / \`INTEGRATION_TEST_PRIVATE_KEY\` secrets are missing or rotated:
1. App-token step fails.
2. Failure handler also fails (no token to authenticate the check-run API call).
3. The gate sits green from the earlier \`skip-integration-tests-pr\` job's synthetic-success check.

Exactly the silent-pass anti-pattern the failure handler exists to prevent.

## How I found it

Exercised the dispatch end-to-end on a draft PR before the App secrets were installed (#800, closed). The dispatch failed as expected, but the failure handler also failed with \`Input required and not supplied: github-token\`, and \`Python Proxy Tests\` stayed green from Phase 1's stale synthetic-success check. Full smoke-test write-up is in the closing comment on #800.

## The fix

Use the default workflow token \`\${{ github.token }}\` for all check-posting steps. The default token has \`checks: write\` because each job already declares the permission. The App token is still used (correctly) for the actual \`peter-evans/repository-dispatch\` call, where cross-repo write access is required.

Since \`steps.public-token\` is no longer referenced anywhere, the App-token generation step itself is also removed to keep the workflow tidy.

## Note on the canonical pattern

The same latent bug exists in [adbc-drivers/databricks's trigger-integration-tests.yml](https://github.com/adbc-drivers/databricks/blob/main/.github/workflows/trigger-integration-tests.yml) — that's where I copied the App-token pattern from. Not fixing upstream here, but worth flagging for the ADBC team.

## Diff

| Change | Lines |
|---|---|
| Failure handlers (×2): \`steps.public-token.outputs.token\` → \`\${{ github.token }}\` | -2 +2 |
| Auto-pass / no-driver-changes steps (×2): same swap | -2 +2 |
| Remove dead App-token-generation steps (×2) | -16 |
| Comments explaining the choice | +12 |
| **Total** | **+20 / -22** (-2 net) |

## Test plan

- [x] \`actionlint\` clean (only the expected \`linux-ubuntu-latest\` runner-label warning that matches every other workflow in this repo).
- [x] YAML parses; all 4 jobs intact.
- [ ] Smoke-test will be straightforward once the App secrets are installed — the next missing-secret state should now show a red \`Python Proxy Tests\` check via the failure handler, instead of staying silently green.

This pull request and its description were AI-assisted by Isaac.

### DIFF
--- a/.github/workflows/trigger-integration-tests.yml
+++ b/.github/workflows/trigger-integration-tests.yml
@@ -194,15 +194,6 @@ jobs:
           owner: databricks
           repositories: databricks-driver-test
 
-      - name: Generate GitHub App Token (public repo)
-        id: public-token
-        uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
-        with:
-          app-id: ${{ secrets.INTEGRATION_TEST_APP_ID }}
-          private-key: ${{ secrets.INTEGRATION_TEST_PRIVATE_KEY }}
-          owner: databricks
-          repositories: databricks-sql-python
-
       - name: Sanitize PR title
         id: sanitize
         uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
@@ -235,7 +226,11 @@ jobs:
         if: steps.changed.outputs.python != 'true'
         uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
         with:
-          github-token: ${{ steps.public-token.outputs.token }}
+          # Default workflow token, not the App token — same rationale
+          # as the failure handler below. We don't want a missing-secret
+          # state to silently swallow the green check for path-filtered
+          # no-op runs.
+          github-token: ${{ github.token }}
           script: |
             await github.rest.checks.create({
               owner: context.repo.owner,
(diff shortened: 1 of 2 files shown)


msrathore-db (approved):

vikrantpuppala:
## Added: fix for pre-existing e2e test flakiness

The `test-with-coverage` job is failing on both this PR and `main` (commit fb55001c) due to **pre-existing flakiness** unrelated to the workflow change in this PR. Two failure modes:

### 1. Telemetry-executor pollution of retry-test mocks (FIXED in 68f4c605)

`test_oserror_retries`, `test_retry_max_count_not_exceeded`, and `test_retry_exponential_backoff` assert exact `urllib3` call counts (`call_count == 6`) while patching `_validate_conn` / `_get_conn` at the global urllib3 layer. The shared `TelemetryClientFactory` executor — populated by *prior* tests in the same xdist worker — keeps firing background telemetry pushes that land on the same mocked urllib3 layer, inflating counts. Captured logs show extra calls to `/telemetry-ext` from background pushes.

The fix adds an `_isolated_from_telemetry()` context manager around those three tests that drains the factory and routes any new `initialize_telemetry_client` call to `NoopTelemetryClient` for the duration of the test. Factory state is fully restored on exit.

### 2. `test_telemetry_flags` workspace-level `/telemetry-ext` 429s (NOT fixed here)

`test_telemetry_flags[*-2-*]` failed with `Expected 2 responses, got 0` because the test fleet's `/telemetry-ext` endpoint was returning sustained 429s, and the unified-HTTP-client's retry stack gave up. This is a workspace/infra-level rate-limit issue, not a test logic bug — leaving it for a separate change so this PR stays focused.

PR #802, opened 2026-05-23
fix(utils): import ProgrammingError instead of bare `exc.` reference

vikrantpuppala:
## Summary

`ParamEscaper.escape_args` and `ParamEscaper.escape_item` both raise `exc.ProgrammingError(...)` (lines 551 and 609 of `utils.py`), but `exc` is never imported. On any unsupported parameter shape the user sees `NameError: name 'exc' is not defined` instead of a clean PEP-249 `ProgrammingError`.

Surfaced via the python-comparator audit harness running the `INLINE_PARAMS` connection config: both backends raised `NameError: name 'exc' is not defined`, which the comparator's class+message match treated as parity — a false-positive that hid both the real driver bug and the underlying caller-side type mismatch.

## Fix

Import `ProgrammingError` directly from `databricks.sql.exc` (matching the pattern used in `session.py`, `client.py`, `result_set.py`, etc.) and replace the two `exc.ProgrammingError(...)` sites with bare `ProgrammingError(...)`.

## Test plan

- [x] Manual repro: `pe = ParamEscaper(); pe.escape_args(object())` now raises `ProgrammingError`, not `NameError`. Same for `pe.escape_item(object())`.
- [ ] Existing test suite — no behaviour change beyond the exception class on a previously-unreachable error path.

### DIFF
--- a/src/databricks/sql/utils.py
+++ b/src/databricks/sql/utils.py
@@ -19,6 +19,7 @@ except ImportError:
     pyarrow = None
 
 from databricks.sql import OperationalError
+from databricks.sql.exc import ProgrammingError
 from databricks.sql.cloudfetch.download_manager import ResultFileDownloadManager
 from databricks.sql.thrift_api.TCLIService.ttypes import (
     TRowSet,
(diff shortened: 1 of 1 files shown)


vikrantpuppala:
Folded into #803 as part of a bundle of five comparator-parity fixes that share the same audit baseline. The exc.ProgrammingError import fix is item #5 in the updated #803 description. Closing this in favour of the bundled review.

PR #817, opened 2026-05-30
Extract SPOG org-id from cluster http_path for non-Thrift requests

msrathore-db:
## Summary

- Fixes silent telemetry loss on SPOG (custom-URL) hosts when connecting to an all-purpose cluster via an http_path like `sql/protocolv1/o/<workspace-id>/<cluster-id>`.
- `Session._extract_spog_headers` now extracts the workspace ID from the `/o/<wsid>/` path segment as a fallback when `?o=<wsid>` is not present, and sets it as the `x-databricks-org-id` header so the workspace-scoped HTTP transports (telemetry, feature flags, SEA) can route correctly.
- Priority order preserved: explicit caller header (case-insensitive) ▶ `?o=` in `http_path` ▶ `/sql/protocolv1/o/<wsid>/` path segment.

## Why

On a SPOG host the workspace identity has to be in either the URL or the `x-databricks-org-id` header so PoPP can route a request to the right workspace. For all-purpose cluster Thrift this is free — the workspace ID is in the `/o/<wsid>/` segment of the http_path, so PoPP routes Thrift via `routing_reason=workspace-id` and the session opens fine without `?o=`.

Connection-scoped HTTP clients used for telemetry, feature flags, and the SEA backend talk to different paths (`/telemetry-ext`, `/api/...`) that do not carry the workspace ID. The previous extraction only looked at `?o=` in the query string, so on a cluster http_path without `?o=` no `x-databricks-org-id` header was ever attached. PoPP fell back to default (account) routing on those endpoints and responded with a 303 redirect to `/login` — silently dropping every telemetry batch.

## What changes

`src/databricks/sql/session.py`
- New compiled regex `Session._CLUSTER_PATH_ORG_ID_RE` matching `(?:^|/)sql/protocolv1/o/(\d+)/[^/?]+`.
- `_extract_spog_headers` now checks the caller-header guard first, then `?o=`, then the cluster path segment. Empty `http_path` returns `{}` as before. Header-name comparison for the caller guard is case-insensitive, matching HTTP semantics.
- The log line reports which source produced the workspace ID.

`tests/unit/test_session.py`
- Six new tests under the existing `TestSpogHeaders` class:
  - Cluster path without `?o=` → header extracted from `/o/<wsid>/`
  - Cluster path with leading `/` → same
  - Cluster path with `?o=` → query-param value wins
  - Cluster path + explicit caller header → caller wins
  - Mixed-case explicit caller header → caller still wins
  - Warehouse path without `?o=` → no header (regression guard: the new regex must not match warehouse paths)

## Test plan

- [x] `pytest tests/unit/test_session.py::TestSpogHeaders -v` — 12 tests passed before the mixed-case regression test was added.
- [x] `pytest tests/unit/test_session.py` — full file passed before the mixed-case regression test was added.
- [x] GitHub CI passed on the upstream-head PR for the current branch.
- [x] Behavior validated on the OSS JDBC equivalent fix against Prod SPOG (`peco.azuredatabricks.net`) all-purpose cluster: telemetry POST `/telemetry-ext` flipped from `HTTP 303 → /login` to `HTTP 200 OK` once the path-segment extraction populated `x-databricks-org-id`. Same shape of fix here.

## Out of scope

- The corresponding OSS JDBC driver fix is opened as databricks/databricks-jdbc#1475.
- The Node.js connector (`databricks/databricks-sql-nodejs`) already extracts the org ID from both query param and path segment (see `extractWorkspaceId`/`hasMalformedOrgParam` in `lib/DBSQLClient.ts`).
- The Go connector (`databricks/databricks-sql-go`) has a separate PR for the same fix.
- The C# ADBC driver (`adbc-drivers/databricks`) has the same parsing limitation in `PropertyHelper.ParseOrgIdFromProperties` and needs a corresponding fix in its own repo.

This pull request and its description were written with assistance from Claude Code.

### DIFF
--- a/src/databricks/sql/session.py
+++ b/src/databricks/sql/session.py
@@ -73,10 +73,10 @@ class Session:
         base_headers = [("User-Agent", self.useragent_header)]
         all_headers = (http_headers or []) + base_headers
 
-        # Extract ?o=<workspaceId> from http_path for SPOG routing.
-        # On SPOG hosts, the httpPath contains ?o=<workspaceId> which routes Thrift
-        # requests via the URL. For SEA, telemetry, and feature flags (which use
-        # separate endpoints), we inject x-databricks-org-id as an HTTP header.
+        # Extract workspace context from http_path for SPOG routing.
+        # On SPOG hosts, the http_path can contain either ?o=<workspaceId> or an
+        # all-purpose-compute /o/<workspaceId>/ path segment. For SEA, telemetry,
+        # and feature flags, we inject x-databricks-org-id as an HTTP header.
         self._spog_headers = self._extract_spog_headers(http_path, all_headers)
         if self._spog_headers:
             all_headers = all_headers + list(self._spog_headers.items())
@@ -197,8 +197,8 @@ class Session:
         if not http_path:
             return {}
 
-        # Caller already set the header — never override.
-        if any(k == "x-databricks-org-id" for k, _ in existing_headers):
+        # Caller already set the header; never override. Header names are case-insensitive.
+        if any(k.lower() == "x-databricks-org-id" for k, _ in existing_headers):
             logger.debug(
                 "SPOG header extraction: x-databricks-org-id already set by caller, "
                 "not extracting from http_path"
@@ -239,7 +239,7 @@ class Session:
         return {"x-databricks-org-id": org_id}
 
     def get_spog_headers(self):
-        """Returns SPOG routing headers (x-databricks-org-id) if ?o= was in http_path."""
+        """Returns extracted SPOG routing headers (x-databricks-org-id), if any."""
         return dict(self._spog_headers)
 
     def open(self):
--- a/src/databricks/sql/session.py
+++ b/src/databricks/sql/session.py
@@ -173,7 +173,8 @@ class Session:
 
     # All-purpose-compute Thrift http_path:
     # [/]sql/protocolv1/o/<workspace-id>/<cluster-id>[/...][?...]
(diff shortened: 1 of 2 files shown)


gopalldb:
🟡 P2 (Minor — fast-follow)

  - Asymmetric: ?o= query value injected unvalidated — session.py diff:84-86. The new path source is digit-only (\d+), but the query source assigns org_id = value with no validation. A control-char ?o= value
  gets stored and later raises ValueError deep in urllib3 when telemetry/FF fire (telemetry swallows it → silent degrade; no crash). Pre-existing, but this PR refactors this exact function and validates the
  new source while leaving the query source unvalidated. → add if value and value.isdigit(): to make the two sources symmetric and turn malformed input into a clean skip.
  - Nested-prefix regex false positive — (?:^|/)sql/protocolv1/o/(\d+)/[^/?]+ with re.search. Confirmed evil/sql/protocolv1/o/999/x → 999. Digit-only value (no injection) and http_path is operator-controlled,
  so low practical risk. Warehouse paths and incomplete/empty-cluster paths correctly don't match (verified). → anchor with ^/re.match if you want strict matching. Python's re does backtrack, but this pattern
  has no overlapping nested quantifiers → no ReDoS.
  - Test gaps — added tests cover happy paths + precedence, but not the nested-prefix false positive, non-numeric/CRLF ?o=, empty ?o=, or incomplete cluster paths. → add these to pin the digit/boundary
  behavior.

gopalldb (approved):
minor comments

Issue #107, opened 2023-04-18
SQLAlchemy Dialect Compliance Tests


unj1m:
I did a lot of work on the BigQuery SQLAlchemy dialect on 2021.

(I needed a BigQuery connection to develop/test that. :smile: )

[I leaned heavily](https://github.com/googleapis/python-bigquery-sqlalchemy/tree/main/tests/sqlalchemy_dialect_compliance) on the [SQLAlchemy Dialect Compliance test suite](https://github.com/zzzeek/sqlalchemy/blob/main/README.dialects.rst).

It doesn't look like y'all are using the compliance suite.  It might be a good way to provide a lot more testing of your dialect.

FWIW.

susodapop:
Thanks for the insight and links. Our focus when writing the initial SQLAlchemy dialect for this package was to functionally replace the community-made `sqlalchemy-databricks` pypi package, which is not being maintained. Full implementation of the SQLA api wasn't our target; although long-term that's desirable. For now, development has been driven by specific customer use-cases.

unj1m:
Would you like to have a PR that sets up the dialect-compliance tests?

unj1m:
Cool. I'm going to PyCon this week, so it may be a couple of weeks. :)

unj1m:
I started working on this.

With a clean checkout,  unit tests run fine.

When I ran the end-to-end tests, which by default only run the sqlalchemy tests, I got lots of errors like:

`Catalog 'None' plugin class not found: spark.sql.catalog.None is not defined`.

I then:

`export catalog=hive_metastore`

which got me past those errors and then I got a lot of:

`Database 'none' not found`.

which were fixed by:

`export schema=default`

I'll update `CONTRIBUTING.md` to mention setting `catalog` and `schema`.

Finally, I'm getting:

```
self = <databricks.sql.thrift_backend.ThriftBackend object at 0x7f28145012a0>
op_handle = TOperationHandle(operationId=THandleIdentifier(guid=b'RJ"M\x82\x19G\xea\x92\xb3\x80\xfa\xc1R\x02\x15', secret=b'\xb7\x...x8dM\xb4\x92\xcb\xa9\x83\x03i\xefC', executionVersion=None), operationType=0, hasResultSet=True, modifiedRowCount=None)
[...]

PR #819, opened 2026-06-02
feat(kernel): wire OAuth (M2M/U2M), TLS/mTLS, and Geometry on use_kernel path

vikrantpuppala:
## What

Wires the OAuth + TLS surface (added to the Rust kernel's pyo3 binding in databricks-sql-kernel) through the `use_kernel` connector path, plus Geometry/Geography result-type mapping. This brings the kernel backend toward parity with Thrift/SEA for auth + TLS.

Companion to the kernel-side PR (databricks-sql-kernel #107). `KERNEL_REV` is bumped to the kernel commit that ships the new `Session` kwargs.

## Changes

**OAuth — kernel-owned token lifecycle**
- `auth_bridge.kernel_auth_kwargs` now takes the *raw* connect() auth options. The OAuth secret is consumed during `auth_provider` construction and isn't recoverable from the built provider, so the bridge reads it from the original kwargs:
  - **M2M:** `oauth_client_id` + `oauth_client_secret` → kernel `auth_type='oauth-m2m'`. The kernel acquires + refreshes tokens via workspace OIDC client-credentials. `oauth_client_secret` is a **new `connect()` kwarg**, honored **only** on `use_kernel=True` (Thrift/SEA unaffected).
  - **U2M:** `auth_type` `databricks-oauth` / `azure-oauth` → kernel `auth_type='oauth-u2m'` (forwards optional `oauth_client_id` / `oauth_redirect_port`); the kernel runs the browser flow.
  - **Custom `credentials_provider`:** rejected with a clear `NotSupportedError`. It's an opaque token source with no extractable raw creds, so the kernel can't own the lifecycle — callers are pointed at `oauth_client_id` / `oauth_client_secret`.
- `session.py` forwards the raw auth kwargs into `KernelDatabricksClient`.

**TLS / mTLS**
- `client._kernel_tls_kwargs` translates the connector's `SSLOptions` → kernel `tls_*` `Session` kwargs: CA file → `tls_ca_cert` (PEM bytes), client cert/key files → `tls_client_cert` / `tls_client_key` (mTLS), and inverts `tls_verify`→`tls_skip_verify`, `tls_verify_hostname`→`tls_skip_hostname_verify`. A password-protected client key is rejected (no kernel surface for it yet). `ssl_options` is no longer accept-and-ignored on this path.

**Geometry**
- `type_mapping` recovers `GEOGRAPHY` / `GEOMETRY` result columns from the `databricks.type_name` metadata (they arrive as Arrow `Utf8`, like `VARIANT`), so PEP-249 `description` reports the precise type instead of `string`.

## Design note — why delegate raw creds rather than reuse the connector's OAuth provider

The connector's auth providers are opaque token sources that mint their own tokens; the OAuth `client_secret` is discarded after provider construction. Rather than run two OAuth stacks (or snapshot a non-refreshing token), we forward the *raw* credentials to the kernel and let it own acquisition + refresh — matching the kernel-owned model the pyo3 binding was built for. A custom `credentials_provider` (the connector's M2M path on Thrift/SEA) can't be delegated this way, so it's explicitly rejected on the kernel path.

## Tests
- **140 kernel unit tests pass** — new coverage: M2M/U2M routing, M2M-takes-precedence-over-PAT, scope normalization, custom-provider rejection, `SSLOptions`→`tls_*` translation (incl. mTLS cert/key, cert-only fallback, encrypted-key rejection, missing-file error), and GEOGRAPHY/GEOMETRY type recovery.
- `black` + `mypy` clean on the kernel backend.

## Caveat for reviewers
`KERNEL_REV` currently points at the kernel's `feat/pyo3-oauth-tls` branch HEAD (companion PR #107, not yet merged) so this PR's CI can build against the new `Session` kwargs. **Re-pin to the squash-merge SHA once #107 lands** before this merges.


gopalldb:
🔴 P0 (Must Fix)
  
  KERNEL_REV pins an unmerged branch HEAD. KERNEL_REV:1 points at 774569bc... which is 1 commit ahead of the kernel's main. CI only validates 40-hex format, not reachability — nothing catches a force-push or
  branch deletion that orphans this SHA. Must wait for the companion kernel PR to merge, then re-pin to the squash-merge SHA. Optional follow-up: add git merge-base --is-ancestor check in kernel-e2e CI.

  🟡 P1 (Should Fix)

  1. OAuth client_secret retained on long-lived connector (client.py:106, read at :178). Stored as self._auth_options in __init__, never scrubbed even after open_session() succeeds. The PR description's
  "secret is consumed during provider construction" contract is true on Thrift/SEA but false here. Trivially leaks via vars(conn), pickle, debugger inspection. Fix: scrub
  self._auth_options.pop("oauth_client_secret", None) in open_session's finally.
  2. tls_verify=False doesn't also imply skip-hostname (client.py:684-687). SSLOptions treats tls_verify=False as a superset of tls_verify_hostname=False; the bridge maps them independently. Depending on
  kernel interpretation, "fully disable TLS" intent either fails on hostname or over-disables. Fix: when tls_verify=False, emit both tls_skip_verify=True AND tls_skip_hostname_verify=True.
  3. credentials_provider + M2M kwargs silently picks M2M (auth_bridge.py:147-159). The rejection at :190 is unreachable when M2M kwargs are also present — M2M wins at step 1. PR description called this out as
   a focus area; implementation doesn't enforce it. Fix: detect ambiguity before resolving.
  4. auth_type='databricks-oauth' + oauth_client_secret silently uses M2M (same routing bug, different combo). User asked for U2M, gets M2M against the wrong principal — failure surfaces only as a confusing
  first-call 401.
  5. No secret-leak tests. Brand-new credential kwarg, zero tests asserting it's absent from caplog, exception __str__/__repr__, or vars(connector). _errors.py:wrap_kernel_exception:133 formats {exc!r} into an
   OperationalError — any kernel-side error embedding arg values would leak through and CI wouldn't notice.
  
  🟢 P2 (Nice-to-have)

  - Empty/unreadable CA file passes through (cryptic kernel error)
  - _normalize_scopes returns None for unknown types instead of raising
  - Geometry/Geography mapping depends on undocumented databricks.type_name stability
  - U2M oauth_scopes is forwarded but untested
  - auth_kwargs/tls_kwargs computed before the try-block — secret materialised in a local that's outside the scrub finally

vikrantpuppala:
Thanks @gopalldb — all P0/P1 addressed plus the actionable P2s, in `eed76ff6`:

**P0 — KERNEL_REV**
- Re-pinned to the squash-merge SHA `ec2288742cbac0cd9fab50da353e8405972eefe9` (now on kernel `main`), replacing the orphaned branch-HEAD SHA. No further re-pin needed before merge. (Good call on the `git merge-base --is-ancestor` CI check — filing that as a kernel-e2e follow-up.)

**P1**
1. **Secret retention** — `open_session`'s `finally` now scrubs `oauth_client_secret` from the long-lived `self._auth_options` (runs even on the failure path). You're right the "consumed during provider construction" contract didn't hold here.
2. **`tls_verify=False` ⊃ hostname** — now emits both `tls_skip_verify` **and** `tls_skip_hostname_verify`, matching `SSLOptions.create_ssl_context` (which sets `check_hostname=False` whenever `tls_verify` is False).
3. **`credentials_provider` + M2M** — now rejected up front with a clear `NotSupportedError` (was unreachable behind the step-1 M2M win).
4. **`databricks-oauth` + `oauth_client_secret`** — same ambiguity class, now rejected naming the conflict instead of silently routing M2M against the wrong principal.
5. **Secret-leak tests** — added: secret forwarded to the kernel but scrubbed from `_auth_options`, and absent from `repr(conn)` / `vars(conn)`; scrub verified on the raise path too.

**P2**
- Empty/whitespace CA/cert file → clear `ProgrammingError` instead of cryptic kernel parse error.
- `_normalize_scopes` raises on non-str/list/tuple instead of silently defaulting.
- Added a U2M `oauth_scopes` forwarding test.
- Left as-is with rationale: the `databricks.type_name` geometry dependency mirrors the existing VARIANT/complex handling (same metadata contract), and the pre-`try` materialization is covered by the new `_auth_options` scrub + local-dict scrub.

150 kernel unit tests pass; black + mypy clean.

gopalldb (approved):

PR #825, opened 2026-06-04
feat(kernel): wire CUJ-gap fixes — staging fail-loud, error context, sync cancel

vikrantpuppala:
## What

Connector half of the `use_kernel` API+CUJ gap audit fixes. The kernel half is **[databricks-sql-kernel#121](https://github.com/databricks/databricks-sql-kernel/pull/121)**; this PR wires it through and bumps `KERNEL_REV`.

> ⚠️ **`KERNEL_REV` is pinned to #121's branch HEAD (`f17a302`)** as a placeholder. **Must re-pin to #121's squash-merge SHA before merging** (orphan-SHA risk otherwise — same as prior connector PRs).

### 1. Staging fail-loud (`kernel/client.py`)
Volume/staging `PUT`/`GET`/`REMOVE` **silently no-op'd** on the kernel path — `KernelResultSet.is_staging_operation` is always `False`, so the connector's `_handle_staging_operation` never fired and **no file was transferred** (ETL silently ingests stale/missing data). Now detects the leading verb in `execute_command` and raises `NotSupportedError`, so it fails loud. (Real staging support tracked separately — needs a kernel staging marker.)

### 2. Error context (`_errors.py`)
Forward `display_message` / `diagnostic_info` / `error_details_json` (now exposed across the pyo3 boundary in #121) onto the re-raised PEP-249 exception, and populate `ServerOperationError.context` with `"diagnostic-info"` (the Spark stack trace) + `"operation-id"` — matching the Thrift backend so `except ServerOperationError as e: e.context["diagnostic-info"]` works identically on `use_kernel`.

### 3. Sync cancel wiring (`client.py`, `kernel/client.py`)
`cursor.cancel()` was a **silent no-op** for the default blocking `execute()` (the documented cross-thread shape) — `active_command_id` is `None` until execute returns. The kernel backend now registers a detached `StatementCanceller` (keyed by the cursor) before the blocking execute and exposes `cancel_running_cursor(cursor)`. `Cursor.cancel()` routes to that hook **via `getattr`** when there's no command id yet — opt-in, so the Thrift/SEA backends (which set `active_command_id` before blocking) are **unaffected**.

## Testing
- **Unit**: staging fail-loud + `_is_staging_statement`; sync-canceller registry + `cancel_running_cursor` + the `Cursor.cancel()` routing (incl. the warning-still-fires-when-nothing-in-flight case, and Thrift/SEA unaffected via `spec=`); error-context / diagnostic-info forwarding. All pass.
- **E2E (live, dogfood, `use_kernel=True`)**: tz-aware TIMESTAMP `+05:30` → UTC; scientific DECIMAL bind; staging PUT → `NotSupportedError`; server error exposes `context["diagnostic-info"]`; **cross-thread `cursor.cancel()` interrupts a running sync query** (cancel at 15s aborts a query that would otherwise run longer — the kernel publishes the statement id within the server wait window).

Depends on #121 (kernel). `black` (pinned 22.3.0) + unit suites green.


gopalldb:
🟠 P1 (Important — should fix)

  1. Staging fail-loud has a hole: comment-prefixed staging statements aren't detected → re-opens the silent-no-op data bug — kernel/client.py:85-90, 396. _is_staging_statement takes the first
  whitespace-delimited token without stripping SQL comments. I reproduced it: '-- upload\nPUT ... INTO ...' → False, '/* c */ PUT ...' → False. A comment-prefixed staging op (common in ETL) flows into
  stmt.execute(), and since kernel KernelResultSet.is_staging_operation is hardcoded False, _handle_staging_operation never fires — the exact silent stale/missing-data bug this PR exists to close. (Thrift
  classifies staging server-side via isStagingOperation, so it handles comment-prefixed staging that this heuristic misses.) → strip leading --…\n and /*…*/ comments before extracting the verb; add tests for
  both comment forms.
  2. Sync-cancel early-cancel window may propagate an exception out of Cursor.cancel() — kernel/client.py:432-443, 485-507. The canceller is registered before stmt.execute(), but the server statement id is
  populated mid-execute. A cancel() arriving in the first ms fires StatementCanceller.cancel() before the id exists; if that raises, cancel_running_cursor wraps it in _wrap_kernel_exception and propagates out
  of the public Cursor.cancel() — which PEP-249 callers expect to be tolerant. The e2e test sleeps 15s precisely to dodge this window, confirming it's unhandled. → confirm with the kernel team that
  StatementCanceller.cancel() is a safe no-op pre-id; if not, swallow/log (match cancel_command's tolerance). Untested.
  
  🟡 P2 (Minor)

  - id(cursor) reuse — adjudicated NOT exploitable. The live execute_command frame holds a strong ref to cursor for its whole duration, so the cursor can't be GC'd (and its id reused) while its entry is live;
  the finally dict.pop(..., None) always clears it and can't raise. The only collision is illegal same-cursor concurrent execute(), where a mis-fire cancels that cursor's own work, not an unrelated query. →
  optional: key by a per-execute monotonic token to be fully robust.
  - Canceller use-after-teardown — cancel_running_cursor reads under lock, releases, then calls .cancel(); the execute thread may stmt.close() concurrently. Python exceptions are caught, but a native pyo3
  panic (if the kernel doesn't guard cancel-after-close) wouldn't be. → confirm StatementCanceller holds its own ref and is safe post-close.
  - Older-wheel degradation — stmt.canceller() missing is swallowed by the registration try/except (cancel silently won't work — acceptable); new error attrs use getattr(..., None). Both graceful; note the
  KERNEL_REV bump means cancel/diagnostic features require the new kernel.
  - e2e sync-cancel test sleeps 15s — slow + timing-fragile.

vikrantpuppala:
Thanks @gopalldb — addressed in `4413040` (+ `KERNEL_REV` bumped to the amended kernel #121).

**P1 #1 — staging fail-loud missed comment-prefixed statements** ✅ Fixed. `_is_staging_statement` now calls `_strip_leading_sql_comments` first, which skips leading `-- …` line and `/* … */` block comments (multiple/mixed) before extracting the verb. So `-- upload\nPUT …` and `/* c */ PUT …` are correctly classified as staging and fail loud. Added unit cases for both comment forms, mixed comments, leading-whitespace, and the inverse (`-- PUT …\nSELECT 1` / `/* PUT */ SELECT 1` must NOT match), plus a live e2e (`test_comment_prefixed_staging_put_raises_not_supported`).

**P1 #2 — early-cancel could propagate out of cursor.cancel()** ✅ Fixed. `cancel_running_cursor` is now tolerant: it swallows + logs (`exc_info=True`) a canceller failure instead of re-raising via `_wrap_kernel_exception`, matching `cancel_command`'s tolerance and the PEP-249 expectation that `cursor.cancel()` doesn't raise. It still returns `True` (a canceller was present and attempted) so the Cursor doesn't emit the misleading "no executing command" warning. (Also: the kernel-side `StatementCanceller.cancel()` is a documented no-op before the id is observed, so the common early-cancel case doesn't even reach this path.) Added a unit test asserting a raising canceller is swallowed and `True` is returned.

**P2s** — all acknowledged:
- `id(cursor)` reuse: agreed not exploitable (live `execute_command` frame pins the cursor; `finally` always pops). Left as-is.
- Use-after-teardown: the kernel `StatementCanceller` holds its own `Arc<SessionInner>` + id slot (not the `Statement`), and `cancel_statement` against a closed/terminal id is a server no-op; the connector now swallows any error regardless. Safe.
- Older-wheel degradation: intentional + graceful (`getattr(..., None)`; canceller registration in try/except). The `KERNEL_REV` bump means these features require the new kernel — expected.
- e2e sleeps 15s: it's `#[ignore]`-style live-only and dodges the pre-id window deliberately; timing-fragile but only runs on the gated live job. Could shorten with a polling wait on `query_id` in a follow-up.

gopalldb (approved):

Issue #123, opened 2023-05-14
Delta has TINYINT now which needs to be added in?


Data-drone:
I got:

```
CompileError: (in table 'dimaccount', column 'taxstatus'): Compiler  can't render element of type TINYINT
```
When working with a sqlalchemy engine with the databricks dialect enabled connected to a SQL Warehouse endpoint.

See:  https://docs.databricks.com/sql/language-manual/sql-ref-datatypes.html#supported-data-types

susodapop:
Thanks for reporting. Fix incoming.

PR #120, opened 2023-05-09
Disable non_native_boolean_check_constraint

bkyryliuk:
Disable non_native_boolean_check_constraint as it is not supported by databricks.
It is true by default: https://github.com/zzzeek/sqlalchemy/blob/main/lib/sqlalchemy/engine/default.py#L162

Tested via monkeypatching out code

error msg:
```
File /opt/miniconda3/envs/dev/lib/python3.9/site-packages/databricks/sql/thrift_backend.py:484, in ThriftBackend._check_command_not_in_error_or_closed_state(self, op_handle, get_operations_resp)
    482 if get_operations_resp.operationState == ttypes.TOperationState.ERROR_STATE:
    483     if get_operations_resp.displayMessage:
--> 484         raise ServerOperationError(
    485             get_operations_resp.displayMessage,
    486             {
    487                 "operation-id": op_handle and op_handle.operationId.guid,
    488                 "diagnostic-info": get_operations_resp.diagnosticInfo,
    489             },
    490         )
    491     else:
    492         raise ServerOperationError(
    493             get_operations_resp.errorMessage,
    494             {
   (...)
    497             },
    498         )

DatabaseError: (databricks.sql.exc.ServerOperationError)```

Generated ddl
```CREATE TABLE bogdan.__bogdan_test_1 (
	booli BOOLEAN,
	CHECK (booli IN (0, 1))
-^^^
) USING DELTA```

PR #830, opened 2026-06-04
fix(kernel): stop premature sync statement close (H4); bump KERNEL_REV to batch 2

vikrantpuppala:
## What

Connector half of the kernel **batch-2** CUJ-gap fixes ([databricks-sql-kernel#123](https://github.com/databricks/databricks-sql-kernel/pull/123), merged). Pins `KERNEL_REV` to merged kernel main and adds the one connector-side change.

### `KERNEL_REV` bump
`KERNEL_REV` → **`f4ee6fec78aabce8c0ea9c1ff47fc11b8191d013`** — current kernel `main`, which includes the batch-2 kernel work (#123: sync-statement auto-close-on-drain, server-CANCELLED → cancelled class, U2M refresh fail-fast, metadata statement close-on-drop, per-binding OAuth client_id) plus #120 (logging bridge) and #125 (version bump). This is a merged, non-orphan SHA reachable from `main`.

### Connector change — don't close the kernel `Statement` at sync execute-return
`execute_command`'s `finally` used to `stmt.close()` immediately after a successful `stmt.execute()`. For a large CloudFetch result with **paginated chunk links** (`all_fetched=false`), the kernel fetches later links lazily (`get_result_chunks` against the **live** statement) during consumption — so a premature `CloseStatement` broke those fetches. (The common `all_fetched=true` case was unaffected, so this only bit very large paginated-link results.)

The kernel now **auto-closes the server statement when its result stream drains** (#123), with the executed-handle `Drop` as the backstop for partial/abandoned reads. So the connector now flips `close_stmt = False` on a successful execute and closes **only on the error path** (where no executed handle / result set was produced to reap the statement).

The other batch-2 fixes (cancelled-class → `OperationalError`, U2M refresh fail-fast, metadata close-on-drop, per-binding OAuth `client_id`) are entirely kernel-side and need no connector code beyond the `KERNEL_REV` bump.

## Testing
- **Unit**: sync execute does-not-close-on-success / does-close-on-failure. 230 connector unit tests pass; black (pinned 22.3.0) clean.
- **E2E (live, dogfood, `use_kernel=True`)**: large multi-chunk result (5M rows) drains without premature close + cursor reuse; server cancel surfaces as `OperationalError` (not `ProgrammingError`); plus the existing staging fail-loud / diagnostic-info / logging-bridge e2e re-verified against the merged kernel.

Kernel #123 is merged; this is ready to merge.


src/databricks/sql/backend/kernel/client.py
```diff
@@ -473,16 +473,29 @@ def execute_command(
...
+                # Execute succeeded: the kernel now owns the statement
+                # lifecycle. It auto-closes the server statement when the
+                # result stream is fully drained (``ExecutedStatement::
+                # next_batch`` end-of-stream), with the executed handle's
+                # ``Drop`` as the backstop for partial/abandoned reads.
+                # So we must NOT close ``stmt`` here: a premature
+                # ``CloseStatement`` at execute-return broke lazy
+                # CloudFetch chunk-link fetches (``get_result_chunks``
+                # against the live statement) for large paginated-link
+                # results — the H4 gap. Closing here is left ONLY for the
```
msrathore-db: let's not mention H4?

msrathore-db (approved):

src/databricks/sql/backend/kernel/client.py
vikrantpuppala: lol nice catch will do

vikrantpuppala: Done in 06b5bfe1 — dropped the 'H4' shorthand from the comment + the two test docstrings, reworded to describe the behavior directly.

Issue #136, opened 2023-06-03
databricks.sql.exc.RequestError: Error during request to server: [Errno -3] Temporary failure in name resolution


MinuraPunchihewa:
Hi guys,
Any idea what could be causing the error given above? This is the full stack trace,
```
2023-06-03 23:44:47,664 - ERROR - Error connecting to Databricks None, Error during request to server: [Errno -3] Temporary failure in name resolution!
Traceback (most recent call last):
  File "/home/minura/PycharmProjects/mindsdb_og/./mindsdb/integrations/handlers/databricks_handler/databricks_handler.py", line 103, in check_connection
    self.connect()
  File "/home/minura/PycharmProjects/mindsdb_og/./mindsdb/integrations/handlers/databricks_handler/databricks_handler.py", line 67, in connect
    self.connection = sql.connect(
  File "/home/minura/anaconda3/envs/mindsdb_og_autosklearn/lib/python3.8/site-packages/databricks/sql/__init__.py", line 50, in connect
    return Connection(server_hostname, http_path, access_token, **kwargs)
  File "/home/minura/anaconda3/envs/mindsdb_og_autosklearn/lib/python3.8/site-packages/databricks/sql/client.py", line 189, in __init__
    self._session_handle = self.thrift_backend.open_session(
  File "/home/minura/anaconda3/envs/mindsdb_og_autosklearn/lib/python3.8/site-packages/databricks/sql/thrift_backend.py", line 464, in open_session
    response = self.make_request(self._client.OpenSession, open_session_req)
  File "/home/minura/anaconda3/envs/mindsdb_og_autosklearn/lib/python3.8/site-packages/databricks/sql/thrift_backend.py", line 393, in make_request
    self._handle_request_error(error_info, attempt, elapsed)
  File "/home/minura/anaconda3/envs/mindsdb_og_autosklearn/lib/python3.8/site-packages/databricks/sql/thrift_backend.py", line 261, in _handle_request_error
    raise network_request_error
databricks.sql.exc.RequestError: Error during request to server: [Errno -3] Temporary failure in name resolution
```
I am using `databricks-sql-connector` version 2.5.2 and Python 3.8.16.

I am trying to connect to a 10.4.x-cpu-ml-scala2.12 cluster.

susodapop:
[...]

PR #837, opened 2026-06-06
feat(kernel): consume the folded directResults execute()

msrathore-db:
## What

Adapt `KernelDatabricksClient.execute_command` to the kernel's folded directResults `execute()` (databricks/databricks-sql-kernel#136), which now returns a **single** `ExecutedAsyncStatement` — seeded with the inline result when the query finished within the server inline wait (fast path, zero extra round-trips), or a poll/cancel handle when still running — instead of always blocking to terminal.

The sync path now drives **one uniform flow** (no `hasattr(await_result)` arm to feature-detect): register the handle (so `cursor.cancel()` / `close()` reach it), then `await_result()` to drive it to a ready result set — preserving `execute()`'s blocking contract.

## Caller-visible behaviour

`cursor.execute()` is unchanged for callers (still blocks to a ready result set); this just adapts to the kernel's new single-handle return. Mid-run cancel still works via the `_sync_cancellers` canceller registered before execute.

## Testing

Tested e2e against live warehouses (**auto-closing and non-auto-closing**): `SELECT 1`, `range(N)`, and `CREATE`+count all succeed on both. The kernel maps an auto-closed `CLOSED` statement to `Succeeded`, so the connector sees a uniform success regardless of warehouse auto-close behaviour.

### DIFF
--- a/src/databricks/sql/backend/kernel/client.py
+++ b/src/databricks/sql/backend/kernel/client.py
@@ -472,20 +472,26 @@ class KernelDatabricksClient(DatabricksClient):
                 except Exception:
                     # Canceller is best-effort; never block execute on it.
                     pass
-                executed = stmt.execute()
-                # Execute succeeded: the kernel now owns the statement
-                # lifecycle. It auto-closes the server statement when the
-                # result stream is fully drained (``ExecutedStatement::
-                # next_batch`` end-of-stream), with the executed handle's
-                # ``Drop`` as the backstop for partial/abandoned reads.
-                # So we must NOT close ``stmt`` here: a premature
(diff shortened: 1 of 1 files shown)


msrathore-db:
Not needed with the additive kernel approach (databricks/databricks-sql-kernel#140). That PR adds execute_direct() *alongside* an unchanged blocking execute(), so use_kernel keeps using execute() exactly as today — no client.py change. Verified e2e against the additive kernel on the unchanged upstream/main driver: CREATE commits, close() cheap (no close-drive), mid-run cancel works (258ms, via the existing _sync_cancellers canceller). The await_result adaptation here was only required for the in-place fold (kernel #136, closed).

PR #796, opened 2026-05-20
Bump thrift to 0.23.0

leoromanovsky:
## Motivation

`databricks-sql-connector` currently constrains thrift with `~=0.22.0`, which prevents downstream consumers from resolving `thrift 0.23.0`. Issue #783 requests allowing thrift 0.23.0 so consumers can pick up the latest thrift security fixes.

Fixes #783.

cc @Korijn @jprakash-db

## Changes

- Widen the runtime thrift constraint to `>=0.22.0,<0.24.0`.
- Regenerate `poetry.lock` so the lockfile selects `thrift 0.23.0` and includes the updated thrift extras metadata.

## Decisions

- Kept `0.22.0` as the lower bound so existing compatible installs remain valid.
- Capped the dependency at `<0.24.0` to allow the 0.23.x line without automatically accepting a future thrift minor release.

## Validation

- `uvx --from poetry==2.3.1 poetry check --lock` passes with existing Poetry deprecation warnings.
- `poetry install -E pyarrow` installs `thrift 0.23.0`, `pyarrow 22.0.0`, and the current project.
- `poetry run python -c 'from importlib.metadata import version; print(version("thrift"))'` reports `0.23.0`.
- `PYTHONPATH=src poetry run python -c 'import thrift.transport.THttpClient; import thrift.protocol.TBinaryProtocol; from databricks.sql.backend.thrift_backend import ThriftDatabricksClient; print("ok")'` passes.
- `poetry run python -m pytest tests/unit/test_thrift_backend.py -rs` passes: 66 passed in 6.92s.
- `poetry run python -m pytest tests/unit` passes: 742 passed, 4 skipped in 103.76s.
- `poetry build -f wheel` succeeds; inspected wheel metadata contains `Requires-Dist: thrift (>=0.22.0,<0.24.0)`.

Issue #783, opened 2026-04-30
Release new version with thrift >=0.23.0 support (6 critical CVEs fixed)


KiranAtmala:
Apache Thrift 0.23.0 was released on April 27, 2026 and fixes 6 High-severity CVEs (CVE-2026-41636, CVE-2026-41605, CVE-2026-41604, CVE-2026-41603, CVE-2026-41602, CVE-2025-48431). All prior versions are affected.

`databricks-sql-connector 4.2.6` currently requires `thrift >=0.22.0, <0.23.0`, which prevents downstream consumers from picking up the security fix. The thrift 0.23.0 Python library has no breaking API changes -- it adds security hardening (recursion depth limits, payload size limits) and drops EOL Python versions.

**Requests:**

1. **Short-term:** Widen the pin in the current codebase to `thrift >=0.22.0, <0.24.0` to allow 0.23.0
2. **Release:** Publish a new version (4.2.7 or 4.3.0) to PyPI so downstream consumers can resolve the CVEs

Many enterprise users are blocked on security scan SLAs and cannot remediate until this is unblocked. Thrift 0.23.0 is also not yet on PyPI, but once it is, having the pin already widened would unblock everyone immediately.

Related: #695, PR #733

PR #840, opened 2026-06-09
Revert "Bump thrift to 0.23.0 (#796)" — fixes install failure on DBR LTS (ES-1960554)

vikrantpuppala:
## What

Reverts #796 ("Bump thrift to 0.23.0"), restoring the thrift pin to `~=0.22.0` (`>=0.22.0,<0.23.0`) and `poetry.lock` to thrift 0.22.0.

This is an exact `git revert` of commit 95352557.

## Why

`databricks-sql-connector` 4.2.7 widened the thrift pin to `>=0.22.0,<0.24.0`, which allowed thrift **0.23.0** to be resolved. thrift ships **sdist-only** (no wheels), so every install builds `setup.py` from source. thrift 0.23.0's `setup.py` calls `sys.exit(0)` on the build success path, which terminates the PEP 517 in-process build backend before pip writes its expected `output.json`. On **setuptools ≤ 68** (shipped by DBR 14.3 LTS and 15.4 LTS) pip treats this as a hard failure:

```
ERROR: Could not install packages due to an OSError:
[Errno 2] No such file or directory: '.../output.json'
```

This caused install/job failures for customers on those runtimes with an unpinned (transitive) dependency on the connector. Tracked as SEV0 **ES-1960554**; 4.2.7 was yanked from PyPI as the immediate mitigation.

## Follow-ups (not in this PR)

- The upstream root cause (the `sys.exit(0)`) has a fix submitted to Apache Thrift: **THRIFT-6067 / apache/thrift#3584**.
- Once a fixed thrift (0.23.1 / 0.24.0) ships, the pin can be re-widened to restore the CVE remediation that #796 originally intended (#783).
- Pre-release CI to install the built artifact on DBR-LTS-like toolchains (old setuptools) so this class is caught before release.

Co-authored-by: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1849,19 +1849,19 @@ files = [
 
 [[package]]
 name = "thrift"
-version = "0.23.0"
+version = "0.22.0"
 description = "Python bindings for the Apache Thrift RPC system"
 optional = false
 python-versions = "*"
 groups = ["main"]
 files = [
-    {file = "thrift-0.23.0.tar.gz", hash = "sha256:5f43448a92c36ed6a450048355d10e231a1787e4c28965f08fabac0eb978914c"},
+    {file = "thrift-0.22.0.tar.gz", hash = "sha256:42e8276afbd5f54fe1d364858b6877bc5e5a4a5ed69f6a005b94ca4918fe1466"},
 ]
 
 [package.extras]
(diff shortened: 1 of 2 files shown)


msrathore-db (approved):
Thanks

PR #881, opened 2026-07-24
ci: post single "Python Integration Tests" check (unify with go/nodejs)

eric-wang-1990:
## Summary

Python was the last multi-backend driver posting **per-mode** checks (`Python Proxy Tests / thrift` + `/ kernel`). The paired driver-test receiver (`databricks-python-integration-tests.yml`) now fans out both backends internally and reports **one aggregated `Python Integration Tests` check**, matching go/nodejs.

This sender change:
- Collapses the five `MODES=['thrift','kernel']` two-check loops (skip / label auto-pass / dispatch-fail / merge-queue auto-pass / merge-queue-fail) into a single `Python Integration Tests` check.
- Adds `proxy_mode: replay` to both dispatch payloads (parity with the other senders — the PR gate is replay-only).
- Fixes the PR-comment link to the new receiver workflow.
- The dispatch was already a single `python-pr-test` (no per-mode fan-out on the sender side).

## Ordering / follow-up
- Land the **driver-test receiver PR first** (it creates the `Python Integration Tests` check name).
- **Branch protection (manual, after both land):** swap required checks — remove `Python Proxy Tests / thrift` + `/ kernel`, add `Python Integration Tests`. Doing it out of order would deadlock PRs, so it's sequenced last.

## Test Plan
- [x] YAML parses
- [x] All 5 check-creations post `Python Integration Tests`; both dispatches carry `proxy_mode: replay`
- [x] No residual `MODES` loop or per-mode check name (only header/swap-note comments reference the retired names intentionally)

### DIFF
--- a/.github/workflows/trigger-integration-tests.yml
+++ b/.github/workflows/trigger-integration-tests.yml
@@ -6,8 +6,8 @@ name: Trigger Integration Tests
 # Mirrors the canonical pattern in adbc-drivers/databricks. The model:
 #
 #   - On a normal PR event (open / push / reopen / non-IT label) we
-#     post `success` Python Proxy Tests checks immediately so the
-#     required checks don't block the PR. The real tests are gated
+#     post a `success` Python Integration Tests check immediately so the
+#     required check doesn't block the PR. The real tests are gated
 #     in the merge queue.
 #   - When a maintainer adds the `integration-test` label we dispatch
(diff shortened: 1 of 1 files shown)


eric-wang-1990:
Folded into #879 (same file, `trigger-integration-tests.yml`). The single-check unification now rides on #879's branch alongside the coverage-fanout sender; closing this standalone PR.

PR #910, opened 2026-08-14
feat(kernel): forward identity federation client ID

vuanhphung:
Threads identity_federation_client_id from connect() through the kernel auth bridge to the PyO3 Session for PAT, OAuth M2M, and OAuth U2M. This enables SP-wide workload identity federation added in databricks-sql-kernel#248.

Pins KERNEL_REV to the merged kernel commit so kernel E2E builds against the matching API, and updates the public documentation. The published 0.2.0 wheel does not include this API; the dependency floor should be updated after the next kernel wheel release.

Validation: syntax check, whitespace check, and repository commit hooks passed. Unit tests were not run because Poetry is unavailable in the local checkout.


KERNEL_REV
```diff
@@ -1 +1 @@
-7ffb30d533c08651ca707b8dd13894c9e01cb68e
+eff8950428f4e6cc9975c663ec919f334962f7d0
```
vuanhphung, line 1: not sure if it's safe to bump this? is there any SOP?

eric-wang-1990 (approved):

PR #913, opened 2026-08-17
docs: add connection parameter reference (Thrift vs Kernel)

eric-wang-1990:
## What

Adds `docs/connection-parameters.md` — a single reference listing **every public connection / session parameter** `databricks.sql.connect(...)` accepts, and whether each is honored on the **Thrift** backend (default) or the **Kernel** backend (`use_kernel=True`). Linked from the README.

## Format

Modeled on the databricks-sql-nodejs [connection-parameter reference (#457)](https://github.com/databricks/databricks-sql-nodejs/pull/457): a per-section table with columns **Option · Type · Thrift · Kernel · Default Value · Note**, a ✅/❌/⚠️/— legend, a "Sources of truth" list, and a "Summary of gaps." Grouped by concern: connection identity · authentication · HTTP/proxy/retries · TLS/SSL · results & type rendering · session defaults & transactions · telemetry.

## SEA column dropped

Earlier revision had a third **SEA** column. Removed — the pure-Python SEA backend (`use_sea=True`) is being deprecated, so it's intentionally omitted; the reference now compares **Thrift vs Kernel** only.

## How support was determined

Every ✅/❌/⚠️ verified against source, not guessed:

- `Connection.__init__` and `Session._create_backend` — the kernel path receives only a **curated, named subset** of kwargs (no `**kwargs` splat), so unsupported params are silently ignored there; called out explicitly.
- `ThriftDatabricksClient`, `KernelDatabricksClient` + `auth_bridge.py`, and the shared `UnifiedHttpClient` / `build_client_context` / auth layer.

Notable gaps documented:

- **Kernel-ignored:** `oauth_client_secret` (M2M) & custom `oauth_scopes`, `experimental_oauth_persistence`, Azure SP OAuth, `_use_cert_as_auth`, cloud-fetch / LZ4 / `_disable_pandas`, arrow-native decimals/timestamps, `staging_allowed_local_path`, most retry fine-tuning, `_socket_timeout` / `_port` / `_connection_uri`.
- `identity_federation_client_id` marked Kernel-supported (matches #910 on `main`).
- `oauth_client_secret` / custom `oauth_scopes` are effectively Kernel-only — the Thrift auth path never reads them.
- Retry defaults reflect the real `_retry_policy` table (count `30`).

Generated against connector version `4.4.0`. Docs-only — no code or behavior changed.


vuanhphung (approved):
is there instructions to keep this in sync as we develop new features? (or we rely on agents to do this automatically)

eric-wang-1990:
> is there instructions to keep this in sync as we develop new features? (or we rely on agents to do this automatically)

The review-bot should in theory raise if it is not up-to-date, but we need to monitor it and updating/improving the prompt if needed

PR #914, opened 2026-08-17
fix(kernel): forward full OAuth U2M app bundle into kernel (PECOBLR-4040)

eric-wang-1990:
## What

On the `use_kernel=True` path, OAuth U2M now forwards the connector's **full coupled OAuth-app bundle** — `client_id` + `oauth_scopes` + `redirect_port` — into the kernel's `oauth-u2m`, instead of sending a bare `{"auth_type": "oauth-u2m"}` and relying on the kernel default.

## Why

[PECOBLR-4039](https://databricks.atlassian.net/browse/PECOBLR-4039) (Done) changed the **kernel core default** U2M app to `databricks-sql-connector` / `sql offline_access` / port `8030`. The Python connector is an **override** of that default, so a bare U2M connection via the kernel was authenticating as the *wrong identity* (`databricks-sql-connector`) instead of `databricks-sql-python`.

Because `client_id` + scopes + `redirect_port` are coupled per OAuth app (each app registers its own redirect URI), the connector must forward all three together rather than a partial bundle.

## How

In `auth_bridge.py`'s U2M branch, each field falls back to the connector's registered default when the caller doesn't override it, sourced from the existing `PYSQL_OAUTH_*` constants (no hardcoded port):

- `databricks-oauth` → `databricks-sql-python` / `["sql", "offline_access"]` / first of `8020-8024`
- `azure-oauth` → the azure client GUID / `["sql", "offline_access"]` / `8030`

Only the redirect **port** is routable into the kernel (it derives `http://localhost:{port}`); the kernel takes a single port, so the first registered port of the range is forwarded. Explicit caller `oauth_client_id` / `oauth_scopes` / `oauth_redirect_port` are honored verbatim, and the `identity_federation_client_id` forwarding already on `main` is preserved.

This mirrors `get_python_sql_connector_auth_provider` exactly, so the kernel path is at **parity with the Thrift path** — even a custom `client_id` with no scopes/port fills in `sql offline_access` + the default port.

## Acceptance criteria

- [x] Bare U2M via `use_kernel` authenticates as `databricks-sql-python` with `sql offline_access` (parity with Thrift)
- [x] User-supplied custom `client_id` + scopes + redirect port honored end-to-end
- [x] Tests covering both default and override routing

## Testing

TDD (watched tests fail, then pass). `tests/unit/test_kernel_auth_bridge.py`: bare `databricks-oauth`/`azure-oauth` full-bundle, custom full override, custom-`client_id`-only fallback, port int-coercion, custom scopes. **44/44 pass**, black clean, no regressions in kernel/session/auth unit suites.

Related: PECOBLR-4039 (kernel-side default, Done).

---

[PECOBLR-4039]: https://databricks.atlassian.net/browse/PECOBLR-4039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/databricks/sql/backend/kernel/auth_bridge.py
```diff
@@ -214,16 +227,51 @@ def kernel_auth_kwargs(
...
-        if scopes is not None:
-            kwargs["oauth_scopes"] = scopes
+        kwargs = {
+            "auth_type": "oauth-u2m",
+            "client_id": client_id or default_client_id,
+            "redirect_port": (
+                int(redirect_port)
+                if redirect_port is not None
+                else default_port_range[0]
+            ),
```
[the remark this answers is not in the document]
eric-wang-1990: This is OK, we should honor the scope parameter

src/databricks/sql/backend/kernel/auth_bridge.py
```diff
@@ -214,16 +251,54 @@ def kernel_auth_kwargs(
         return kwargs
 
     # 3. OAuth U2M — browser authorization-code flow; the kernel runs it.
-    if auth_type in ("databricks-oauth", "azure-oauth"):
-        kwargs = {"auth_type": "oauth-u2m"}
-        if client_id:
-            kwargs["client_id"] = client_id
+    #    Only databricks-oauth reaches here (azure-oauth was rejected up
```
eric-wang-1990: Comment too verbose, make it concise

src/databricks/sql/backend/kernel/auth_bridge.py
```diff
@@ -214,16 +251,54 @@ def kernel_auth_kwargs(
...
+        _normalize_scopes(opts.get("oauth_scopes"))
+        kwargs = {
+            "auth_type": "oauth-u2m",
+            "client_id": client_id or PYSQL_OAUTH_CLIENT_ID,
+            "redirect_port": (
+                int(redirect_port)
+                if client_id and redirect_port is not None
+                else PYSQL_OAUTH_REDIRECT_PORT_RANGE[0]
+            ),
+            "oauth_scopes": list(PYSQL_OAUTH_SCOPES),
```
eric-wang-1990: Do not hardcode, we should be able to take in customized scopes

src/databricks/sql/backend/kernel/auth_bridge.py
```diff
@@ -214,16 +251,54 @@ def kernel_auth_kwargs(
...
+        # Validate any caller-supplied oauth_scopes (a bad type is still a
+        # caller error worth flagging) but do NOT forward it: the Thrift
+        # path hardcodes PYSQL_OAUTH_SCOPES for U2M, so we do the same for
+        # parity rather than letting the kernel path honor an override the
+        # other backend silently ignores.
+        _normalize_scopes(opts.get("oauth_scopes"))
+        kwargs = {
+            "auth_type": "oauth-u2m",
+            "client_id": client_id or PYSQL_OAUTH_CLIENT_ID,
+            "redirect_port": (
```
eric-wang-1990: redirect_port should just get passed in, we either use the pass in one or default one, no need to judge based on client_id existense.

vuanhphung (approved):

vikrantpuppala (approved):

src/databricks/sql/backend/kernel/auth_bridge.py
```diff
@@ -214,16 +252,33 @@ def kernel_auth_kwargs(
...
-            kwargs["redirect_port"] = int(redirect_port)
+        # Honor a caller-supplied oauth_scopes (normalized to a list of
+        # strings); fall back to the connector default when none is given.
         scopes = _normalize_scopes(opts.get("oauth_scopes"))
-        if scopes is not None:
-            kwargs["oauth_scopes"] = scopes
+        kwargs = {
+            "auth_type": "oauth-u2m",
+            "client_id": client_id or PYSQL_OAUTH_CLIENT_ID,
+            "redirect_ports": (
```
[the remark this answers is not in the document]
eric-wang-1990: Yes it does support

PR #928, opened 2026-08-23
fix: surface the token-exchange error instead of KeyError('access_token')

### LINKED ISSUE #904: Token federation: `_exchange_token` raises KeyError('access_token') on error responses, discarding the real failure reason
## Summary

When token exchange fails, `_exchange_token` raises `KeyError: 'access_token'` instead of surfacing the error the endpoint actually returned. The caller catches it and logs the `KeyError`, so the log line reports the name of a missing dict key rather than why the exchange was rejected:

```
Token exchange failed, using external token: 'access_token'
```

There is no way to tell from that whether the exchange was misconfigured, unauthorized, or unsupported.

## Versions

Reproduced on `databricks-sql-connector==4.3.0`. The relevant code is byte-identical in `v4.4.0` and on `main` today, so this is not fixed in a later release.

## Where

`src/databricks/sql/auth/token_federation.py` on `main`:

```python
191:        token_response = json.loads(response.data.decode())
192:
193:        return Token(
194:            token_response["access_token"], token_response.get("token_type", "Bearer")
195:        )
```

`token_type` is read defensively with `.get()`; `access_token` is not. When the endpoint returns an OAuth error body (`{"error": ..., "error_description": ...}`) rather than a token, line 194 raises `KeyError`.

That propagates to the handler at line 148:

```python
147:            except Exception as e:
148:                logger.warning("Token exchange failed, using external token: %s", e)
```

`str(KeyError("access_token"))` renders as `'access_token'`, which is what reaches the log. The `error` and `error_description` from the response body are never read and are lost.

## Reproduction

1. Connect using an OAuth access token issued by an identity provider whose `iss` host differs from the workspace host — for example an Microsoft Entra ID token for the Azure Databricks resource (`2ff814a6-3304-4ab8-85cb-cd0e6f879c1d`).
2. `TokenFederationProvider` wraps every provider unconditionally (`src/databricks/sql/auth/auth.py:68`, "Always wrap with token federation"), and `_should_exchange_token` returns `True` because the issuer host does not match the workspace host, so an exchange is always attempted.
3. Against a workspace where that exchange is not accepted, every connection logs the message above.

Functionally this is harmless — the fallback to the external token works correctly and queries succeed. The problem is purely diagnostic: the warning fires on every connection and gives no actionable information.

## Suggested fix

Read the response defensively and raise something that names the actual failure, without logging the token itself:

```python
token_response = json.loads(response.data.decode())

if "access_token" not in token_response:
    error = token_response.get("error", "unknown_error")
    description = token_response.get("error_description", "")
    raise RuntimeError(f"Token exchange rejected by {token_url}: {error} {description}".strip())

return Token(token_response["access_token"], token_response.get("token_type", "Bearer"))
```

The existing `except Exception` at line 147 would then log the endpoint's own reason, and the graceful fallback behaviour is unchanged.

A non-JSON or non-2xx response would also currently surface as a confusing `JSONDecodeError`; checking the status code before parsing would cover that case too.

TangoEnSkai:
## Context

`TokenFederationProvider` wraps every credentials provider unconditionally
(`auth.py:68`, "Always wrap with token federation"), and `_should_exchange_token`
returns `True` whenever the token's issuer host differs from the workspace host —
for example a Microsoft Entra ID token against an Azure Databricks workspace. An
exchange is therefore attempted on every such connection.

When the exchange endpoint refuses, it replies with an OAuth error body
(`{"error": ..., "error_description": ...}`) rather than a token.
`_exchange_token` read `token_response["access_token"]` unconditionally, so that
reply raised `KeyError('access_token')`. The handler in `_get_token` logs the
exception, and `str(KeyError("access_token"))` renders as the key name:

```
Token exchange failed, using external token: 'access_token'
```

The endpoint's own `error` and `error_description` were never read, so there was
no way to tell whether the exchange was misconfigured, unauthorized, or simply
unsupported for that workspace. Note `token_type` on the very next line is
already read defensively with `.get()`.

## What

- `_exchange_token` checks for `access_token` before reading it. If it is
  missing, it raises a `ValueError` naming the endpoint, the HTTP status, and the
  returned `error` / `error_description`.
- A non-JSON or truncated body now raises a `ValueError` naming the endpoint and
  status instead of surfacing a bare `JSONDecodeError`.
- `tests/unit/test_token_federation.py`: `test_exchange_token_failure` previously
  asserted `pytest.raises(KeyError)` — it pinned the behaviour this issue is
  about — and is updated to assert the endpoint's reason is surfaced. Three tests
  are added for the no-error-fields body, the non-JSON body, and the unchanged
  fallback to the external token.

The response carries no token on the failure paths, so nothing sensitive is
placed in the message.

## Why

This is diagnostic only — the fallback to the external token works and queries
succeed, which is exactly why it is worth fixing: the warning fires on every
connection and tells the operator nothing actionable. The fix is contained to the
one function and preserves the graceful degradation, since `_get_token` still
catches the exception and falls through to the external token.

## Completion Criteria

- [x] `_exchange_token` surfaces the endpoint's `error` / `error_description`
- [x] Non-JSON responses report the endpoint and status, not a `JSONDecodeError`
- [x] Fallback to the external token is unchanged, covered by a new test
- [x] `pytest tests/unit` — 531 passed, 217 skipped, 185 subtests passed
- [x] Reverting the source change fails exactly the three new/updated tests and
      nothing else
- [x] `black --check src` clean
- [x] CHANGELOG entry under `# Unreleased`
- [x] Commit signed off (DCO)

close #904


TangoEnSkai:
Merged current main into this branch and resolved the CHANGELOG conflict, preserving both the upstream entries and this fix. The PR is mergeable again at be39535f.

Validation after the merge:
- Token federation tests: 34 passed.
- Full unit suite: 586 passed, 237 skipped, 185 subtests passed.
- Black check on src: passed (64 files unchanged).

The new GitHub Actions runs are awaiting maintainer approval (`action_required`), including Code Quality, DCO, and E2E/coverage. Could a maintainer approve those runs and review the fix? The existing external Python Integration Tests check is green.

PR #931, opened 2026-08-25
feat(kernel): forward socket timeout

vuanhphung:
Forward the connector's existing `_socket_timeout` to the kernel's PyO3 `request_timeout_secs` binding introduced in databricks/databricks-sql-kernel#288. Positive values become total HTTP request deadlines; unset or zero selects the kernel's 120-second default.

The kernel pin is updated to the merged binding commit. Focused kernel client/session tests pass, and the changed files pass Black.

---


src/databricks/sql/backend/kernel/client.py
```diff
@@ -330,6 +332,11 @@ def open_session(
             # Translate the connector's ``_retry_*`` kwargs into the
             # kernel's ``retry_*`` kwargs. Empty when at defaults.
             retry_kwargs = _kernel_retry_kwargs(self._retry_options)
+            request_timeout_kwargs: Dict[str, Any] = {}
```
eric-wang-1990: Why this needs to be defined as a dictionary? Do we anticipate multiple attribute to be part of request_timeout_secs? It seem we only need one duration?

vuanhphung: @eric-wang-1990 Updated to simplify

eric-wang-1990 (approved):

PR #935, opened 2026-08-27
Reject timestampAsString override on kernel sessions

jay-xiao446:
## Summary
- reject non-false spark.thriftserver.arrowBasedRowSet.timestampAsString in the Python kernel backend before opening a session
- add a kernel client unit test covering the rejection path
- add an Unreleased changelog entry

## Tests
- .venv/bin/python -m pytest tests/unit/test_kernel_client.py -q


src/databricks/sql/backend/kernel/client.py
```diff
@@ -310,6 +328,7 @@ def open_session(
         session_conf: Optional[Dict[str, str]] = None
         if session_configuration:
             session_conf = {k: str(v) for k, v in session_configuration.items()}
+            _check_session_configuration(session_conf)
```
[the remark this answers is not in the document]
jay-xiao446: the comment is reverted, fix it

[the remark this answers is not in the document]
jay-xiao446: we should inject false , following what Thrift does

src/databricks/sql/backend/kernel/client.py
```diff
@@ -319,6 +328,15 @@ def open_session(
...
             session_conf = {k: str(v) for k, v in session_configuration.items()}
+            # ``_check_session_configuration`` only *rejects* a non-``false``
+            # ``TIMESTAMP_AS_STRING_CONFIG`` override; unlike the Thrift backend
+            # (``thrift_backend.py``), it deliberately does NOT inject
+            # ``TIMESTAMP_AS_STRING_CONFIG = "false"`` into ``session_conf``.
+            # The kernel owns type handling natively (Arrow / complex-types),
+            # and the Thrift-specific conf may be unsupported on the SEA session
+            # path, so pinning it here would be redundant at best and rejected
+            # at worst. The divergence is intentional, not accidental.
+            _check_session_configuration(session_conf)
```
vuanhphung: should we move this check into the try block below? There seems to be an effort to scrub `self._auth_options` regardless of success/failures

vuanhphung (approved):

src/databricks/sql/backend/kernel/client.py
```diff
@@ -106,6 +102,21 @@ def _strip_leading_sql_comments(sql: str) -> str:
     return sql[i:]
 
 
+def _check_session_configuration(session_configuration: Dict[str, str]) -> None:
```
eric-wang-1990: Why do we need this?
Thrift pass this in ExecuteStatement, I am not even sure if session can be opened if this is passed through. Does dbsql support this param at all?
[[CONFIG_NOT_AVAILABLE.WITHOUT_SUGGESTION](https://learn.microsoft.com/azure/databricks/error-messages/config-not-available-error-class#without_suggestion)] Configuration spark.thriftserver.arrowbasedrowset.timestampasstring is not available.  SQLSTATE: 42K0I

eric-wang-1990: The Thrift one come in through ExecuteStatement and it is added by the python connector itself, but here you are guarding against OpenSession.
For kernel/SEA that flag should not matter and we should not need to guard anything about it since nobody gonna pass it on purpose.

eric-wang-1990:
There is no meaningful code changes in this PR I think we can just close it @jay-xiao446

src/databricks/sql/backend/kernel/client.py
jay-xiao446: good catch, thanks

PR #948, opened 2026-09-09
feat: auto-recover Reyden Thrift connections onto the kernel

rahuls-db:
## What type of PR is this?

- [x] Feature

## Description

An unconfigured `connect()` to a Reyden / Real-Time SQL warehouse defaults to the Thrift backend, which the SQL Gateway proxy rejects (SQLSTATE `KP001`). This change detects that rejection at `OpenSession` and transparently re-opens the session on the kernel (SEA) backend, so no customer configuration change is needed.

- **Detection:** `ThriftDatabricksClient._check_response_for_error` raises a distinct `ReydenThriftUnsupportedError` when the OpenSession `TStatus` carries SQLSTATE `KP001` (matched on the SQLSTATE only — no message-string matching).
- **Recovery:** `Connection` catches the marker and re-opens the session once with `use_kernel=True`. If the kernel open also fails, the kernel error is surfaced with the Thrift rejection preserved in the exception chain.
- **Pinning:** a process-wide cache keyed by `(host, warehouse_id)` (~6h TTL, host-scoped for multi-tenant safety, opportunistic eviction of expired entries) remembers warehouses that reject Thrift, so subsequent connects skip the doomed Thrift round-trip and open the kernel directly.
- **Scope / guardrails:** only the default (unspecified) backend path auto-recovers; an explicit `use_kernel`/`use_sea` is always honored. There is no client kill switch — the server-side SAFE flag gates whether the rejection is emitted at all.

Note: the kernel backend is currently an optional extra, so on a base install a rejection surfaces a clear "install `databricks-sql-connector[kernel]`" error. Making the kernel a hard dependency for fully silent recovery is a planned follow-up (packaging change, likely a major version bump).

## How is this tested?

- [x] Unit tests
- [x] E2E Tests
- [x] Manually

Unit: new `tests/unit/test_reyden_warehouse_cache.py` (cache behavior + warehouse-id extraction + expiry sweep); `TestReydenThriftFallback` in `tests/unit/test_session.py` (reactive recovery, cache pre-check, cache marking, explicit-backend guardrail, non-Reyden error not recovered, kernel-retry error chaining); `test_reyden_sqlstate_raises_distinct_marker` in `tests/unit/test_thrift_backend.py`.

E2E (against a real prod Reyden warehouse): an unconfigured `connect()` transparently recovered onto the kernel and `SELECT * from range(10)` returned all rows; a second `connect()` in the same process took the cache pre-check path and opened the kernel without any Thrift round-trip.

Design: "Simplifying Reyden Onboarding on Drivers" (Option A). Mirrors the ADBC driver change adbc-drivers/databricks#670.

---


src/databricks/sql/client.py
```diff
@@ -512,6 +524,85 @@ def read(self) -> Optional[OAuthToken]:
...
+        )
+        if explicit_backend:
+            return build_session(kwargs)
+
+        warehouse_id = extract_warehouse_id(http_path)
+
+        # Pre-check: a warehouse already seen to reject Thrift opens straight on
+        # the kernel, skipping the doomed Thrift OpenSession round-trip.
+        if warehouse_id and is_known_reyden(server_hostname, warehouse_id):
+            logger.info(
```
vuanhphung, line 600: should this be a warning?

src/databricks/sql/client.py
```diff
@@ -512,6 +524,85 @@ def read(self) -> Optional[OAuthToken]:
...
+                warehouse_id,
+                server_hostname,
+            )
+            return build_session({**kwargs, "use_kernel": True})
+
+        try:
+            return build_session(kwargs)
+        except ReydenThriftUnsupportedError as thrift_ex:
+            logger.info(
+                "Thrift is not supported for this Reyden/Real-Time warehouse; "
```
vuanhphung, line 612: ditto

src/databricks/sql/client.py
```diff
@@ -512,6 +524,85 @@ def read(self) -> Optional[OAuthToken]:
...
+            "use_sea", False
+        )
+        if explicit_backend:
+            return build_session(kwargs)
+
+        warehouse_id = extract_warehouse_id(http_path)
+
+        # Pre-check: a warehouse already seen to reject Thrift opens straight on
+        # the kernel, skipping the doomed Thrift OpenSession round-trip.
+        if warehouse_id and is_known_reyden(server_hostname, warehouse_id):
```
vuanhphung, line 599: I think server_hostname may be the same for SPOG workspaces. only http_path contains ` ?o=<workspace-id>`

src/databricks/sql/client.py
```diff
@@ -512,6 +524,85 @@ def read(self) -> Optional[OAuthToken]:
...
+            logger.info(
+                "Warehouse %s on %s is known to require the kernel backend; "
+                "opening on the kernel and skipping Thrift.",
+                warehouse_id,
+                server_hostname,
+            )
+            return build_session({**kwargs, "use_kernel": True})
+
+        try:
+            return build_session(kwargs)
```
vuanhphung, line 609: there's an edge case of passing in auth_type = None (behavioral differences documented in CONNECTION_PARAMETERS.md). 

In Thrift path, it will default to Oauth.

In the case of auto recovery, it will still pass auth_type = None to kernel and then kernel will reject

src/databricks/sql/client.py
rahuls-db: I didn't make it a warning because it is going to be a steady state "feature" and not really an issue that we need to highlight to the customer.

src/databricks/sql/client.py
rahuls-db: same as above

src/databricks/sql/client.py
rahuls-db: warehouse_id is globally unique, there's no collision to worry about on SPOG (or anywhere); host is purely an optimization/defense-in-depth, not a correctness requirement

src/databricks/sql/client.py
rahuls-db: good catch

vuanhphung (approved):
btw, you can also add skip-coverage label for this feature, since it's expected that there are multiple PRs to be landed and we can add one final test at the end

